PR #25, opened 2014-12-23
Max length of deviceToken param is 32 characters

stsai-okta:


eklein-okta:
Looks cool to me. I'm am merging it. Thanks!

PR #26, opened 2015-01-23
Users.md changes second email, profile update post method

eklein-okta:
@larsjohansen-okta @stsai-okta @tthompson-okta @karlmcguinness-okta 

These are the changes to the users api for

1. Remove the 10000 default for api results per DDRR.
2. Added secondary email in many places.
3. New post method for partial user updates.

Please review these changes and let me know if we can push this to master.

### DIFF
--- a/docs/api/rest/users.md
+++ b/docs/api/rest/users.md
@@ -889,13 +889,15 @@ Enumerates users in your organization with pagination.  A subset of users can be
 Parameter | Description                                                                          | Param Type | DataType | Required | Default
 --------- | ------------------------------------------------------------------------------------ | ---------- | -------- | -------- | -------
 q         | Searches `firstName`, `lastName`, and `email` attributes of users for matching value | Query      | String   | FALSE    |
-limit     | Specified the number of results                                                      | Query      | Number   | FALSE    | 
+limit     | Specified the number of results                                                      | Query      | Number   | FALSE    | 200
 filter    | [Filter expression](/docs/getting_started/design_principles.html#filtering) for users   | Query      | String   | FALSE    |
 after     | Specifies the pagination cursor for the next page of users                           | Query      | String   | FALSE    |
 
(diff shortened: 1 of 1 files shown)


docs/api/rest/users.md
```diff
@@ -874,7 +889,7 @@ Enumerates users in your organization with pagination.  A subset of users can be
 Parameter | Description                                                                          | Param Type | DataType | Required | Default
 --------- | ------------------------------------------------------------------------------------ | ---------- | -------- | -------- | -------
 q         | Searches `firstName`, `lastName`, and `email` attributes of users for matching value | Query      | String   | FALSE    |
-limit     | Specified the number of results                                                      | Query      | Number   | FALSE    | 10000
```
tthompson-okta: My concern with this is that it now implies there is no default limit, so a developer would think that they are at risk of getting way too many results if they don't specify the 'limit' manually. I'm in favor of changing it to 200 here, with the logic being that new API customers are going to be reading this more frequently than existing customers who already have their integrations built. @karlmcguinness-okta do you have an opinion?

eklein-okta: @tthompson-okta @karlmcguinness-okta 
At the deep dive review, they specifically asked doc to remove the limit. The deep dive notes are at https://oktawiki.atlassian.net/wiki/display/eng/2015-01-20+DDRR+Meeting+notes. Should we push this to master?

tthompson-okta: Karl says we should change it to 200 here, with a note that this will be rolled out across legacy orgs which currently have a higher default limit.

eklein-okta: @tthompson-okta @karlmcguinness-okta 

I made the change. Are we ready to push this to master?

tthompson-okta: Yep, looks good.

tthompson-okta:
:ship:

PR #35, opened 2015-03-03
Update enabling_cors.md

eklein-okta:
@karlmcguinness-okta @croche-okta 
This is the corrected PR for the CORS GA that just removes Early Access notifications. It would be great to merge this to master March 4, when the preview goes out.


mcguinness:
:ship:

PR #56, opened 2015-05-12
Adds social links for blog posts

jtodd-okta:
Icons come from the existing font awesome library that's in place.

image

I also cleaned up the styling of the title to not show the underline for blog posts.


zdevil:
:shipit:

PR #63, opened 2015-05-15
Fixing broken links to the "Setting up a SAML application in Okta" example page

gene1wood:
This uses relative paths to fix the broken links. If you prefer absolute links it can be changed.


joelfranusic-okta:
Thanks for the PR, we're in the middle of updating our deployment workflow so this will take us a day or so to merge in.

gene1wood:
@joelfranusic-okta no prob. How'd the deployment workflow change go last month?

PR #84, opened 2015-07-30
Removed `co` and `sw` operators from filter operators

benjaminwesson-okta:
Deleted `co` and `sw` operators from the table of filter operators in
the Design Principles guide.

### DIFF
--- a/_source/docs/api/getting_started/design_principles.md
+++ b/_source/docs/api/getting_started/design_principles.md
@@ -159,38 +159,37 @@ The filter parameter **must** contain at least one valid Boolean expression. Eac
 ### Operators
 
 The operators supported in the expression are listed in the following table.
-
-Operator | Description | Behavior
--------- | ----------- | --------
-eq | equal | The attribute and operator values must be identical for a match.
-co | contains | The entire operator value must be a substring of the attribute value for a match.
-sw |starts with | The entire operator value must be a substring of the attribute value, starting at the beginning of the attribute value. This criterion is satisfied if the two strings are identical.
-pr | present (has value) | If the attribute has a non-empty value, or if it contains a non-empty node for complex attributes there is a match.
-gt | greater than | If the attribute value is greater than operator value, there is a match. The actual comparison is dependent on the attribute type. For `String` attribute types, this is a lexicographical comparison and for `Date` types, it is a chronological comparison.
-ge | greater than or equal | If the attribute value is greater than or equal to the operator value, there is a match. The actual comparison is dependent on the attribute type. For `String` attribute types, this is a lexicographical comparison and for `Date` types, it is a chronological comparison.
-lt | less than | If the attribute value is less than operator value, there is a match. The actual comparison is dependent on the attribute type. For `String` attribute types, this is a lexicographical comparison and for `Date` types, it is a chronological comparison.
-le | less than or equal | If the attribute value is less than or equal to the operator value, there is a match. The actual comparison is dependent on the attribute type. For `String` attribute types, this is a lexicographical comparison and for `Date` types, it is a chronological comparison.
-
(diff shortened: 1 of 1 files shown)


karlmcguinness-okta:
What was the reason for removing?

benjaminwesson-okta:
https://oktainc.atlassian.net/browse/OKTA-64880


_source/docs/api/getting_started/design_principles.md
```diff
@@ -159,39 +159,38 @@ The filter parameter **must** contain at least one valid Boolean expression. Eac
...
-lt | less than | If the attribute value is less than operator value, there is a match. The actual comparison is dependent on the attribute type. For `String` attribute types, this is a lexicographical comparison and for `Date` types, it is a chronological comparison.
-le | less than or equal | If the attribute value is less than or equal to the operator value, there is a match. The actual comparison is dependent on the attribute type. For `String` attribute types, this is a lexicographical comparison and for `Date` types, it is a chronological comparison.
-
-Note: Some resources do not support all operators.
+|-----------|-----------------------|---------------------------------------------------------------------------|
+| Operator  | Description           | Behavior                                                                  |
+|:----------|:----------------------|:--------------------------------------------------------------------------|
+| eq        | equal                 | The filter exactly matches the property value                             |
```
mohitchugh-okta: Although the term 'property' is just as apt, everywhere else in the documentation it is referred to as an 'attribute'. We might want to stick to 'attribute' until the time we end up changing all of them to 'property'.

_source/docs/api/getting_started/design_principles.md
```diff
@@ -159,39 +159,38 @@ The filter parameter **must** contain at least one valid Boolean expression. Eac
...
+| pr        | present               | The property value is not empty and/or does not contain an empty node     |
+| gt        | greater than          | The filter is greater than the property value[^1]                         |
+| ge        | greater than or equal | The filter is greater than or equal to the property value[^1]             |
+| lt        | less than             | The filter is less than property value[^1]                                |
+| le        | less than or equal    | The filter is less than or equal to the property value[^1]                |
```
mohitchugh-okta: I like the footnote, much cleaner, though it would have been great if it wasn't all the way at the bottom was instead just below the table.

benjaminwesson-okta: Yeah, unfortunately, the footnote gets rendered at the end of the doc no matter where you place it. 

__"It does not matter where you put a footnote definition in a kramdown document; the content of all referenced footnote definitions will be placed at the end of the kramdown document. Not referenced footnote definitions are ignored."__

mohitchugh-okta: I looked into the reason why the filtering link is not working from the User's page. It is referring to - http://developer.okta.com/docs/getting_started/design_principles.html#filtering instead of 
http://developer.okta.com/docs/api/getting_started/design_principles.html#filtering (it is missing 'api'), in case you wanted to fix that as well.

_source/docs/api/getting_started/design_principles.md
benjaminwesson-okta: Agreed. The terminology is all over the place. @karlmcguinness-okta may have chosen the word "attributes" deliberately because people are more familiar with XSD than JSON Schema. 

As per RFC4627, we should refer to JSON Objects as having "members" or "properties" [see excerpt below]. I opted for "property" because "members" reminds me of C++.

**3.  Core terminology

3.1.  Property, item

   When refering to a JSON Object, as defined by [RFC4627], the terms
   "member" and "property" may be used interchangeably.

   When refering to a JSON Array, as defined by [RFC4627], the terms
   "element" and "item" may be used interchangeably.**

PR #88, opened 2015-08-19
Add tombstone API to Users API

eklein-okta:
@amoghvasekar-okta @haishengwu-okta 

Please review doc for the new Tombstone API.


eklein-okta:
API doc cancelled by PM.

PR #114, opened 2015-12-07
remove embedded resource link of User Model since we don't have yet.

haishengwu-okta:


joelfranusic-okta:
Thanks @haishengwu-okta - this change will be reflected on the site when we next build the live site.

PR #130, opened 2015-12-24
OKTA-77020

larsjohansen-okta:
- Added section for User API Advanced Search in the API docs.
- Modified the User API postman collection.

Reviewers: @joelfranusic-okta @mattloughran-okta @tthompson-okta @karlmcguinness-okta @cwu-okta @eklein-okta


_source/docs/api/resources/users.md
```diff
@@ -1006,6 +1008,30 @@ Users updated after 06/01/2013 but with a status of `LOCKED_OUT` or `RECOVERY`
 
     filter=lastUpdated gt "2013-06-01T00:00:00.000Z" and (status eq "LOCKED_OUT" or status eq "RECOVERY")
 
+###### Advanced Search
+
+Advanced search is similar to filtering, and it gives you all the same possibilities with the `search` query parameter. It also gives you the option to filter on any user profile attribute, including custom defined attributes.
+Note that the results might not yet be up to date, as the most up to date data can be delayed up to a few seconds, so use for convenience.
+
```
cwu-okta: I might also suggest you mention searching is case insensitive. 

Aren't the SCIM operators supported in advanced search more advanced than that of filter? If so we should update the filters section to indicate which new SCIM operators are supported in advanced search.

_source/docs/api/resources/users.md
```diff
@@ -1006,6 +1008,30 @@ Users updated after 06/01/2013 but with a status of `LOCKED_OUT` or `RECOVERY`
 
     filter=lastUpdated gt "2013-06-01T00:00:00.000Z" and (status eq "LOCKED_OUT" or status eq "RECOVERY")
 
+###### Advanced Search
+
+Advanced search is similar to filtering, and it gives you all the same possibilities with the `search` query parameter. It also gives you the option to filter on any user profile attribute, including custom defined attributes.
```
cwu-okta: Consider rewording as this doesn't tell the user that much as he is able to understand which query parameter is used by the description above. Consider:
```Advanced search allows the option to filter on any user profile attribute, including custom defined attributes. All filtering is case insensitive.```

File: _source/docs/api/resources/users.md (hunk 1 above)
Comment (larsjohansen-okta) [inline _source/docs/api/resources/users.md]: I've added a sentence about case insensitivity. I thought that we wanted to support the same operators for normal and advanced search, but I'd be good to get that confirmed. @sswaika-okta @prestonchan-okta do you know?

File: _source/docs/api/resources/users.md (hunk 2 above)
Comment (larsjohansen-okta) [inline _source/docs/api/resources/users.md]: I like that wording better, but I'd change it to be:
```
... against all fields specified in the search parameter.
```
to be clear that we're using a different parameter, and not the ```q``` parameter.

cwu-okta:
@eklein-okta

eklein-okta:
Very minor nit to add comma in line 934 after query in A subset of users can be returned that match a supported filter expression, query or advanced search. But so minor, I will say ship it!

joelfranusic-okta:
I don't see anything that looks obviously wrong. I haven't tested any of the `curl` commands or loaded the Postman collection, but I'm assuming that @larsjohansen-okta has done those things and trust his expertise here.

larsjohansen-okta:
As we're not 100% sure how we should show beta features in the public api doc yet, we'll wait with this PR. Will reopen when time is right.

PR #138, opened 2016-01-04
Add initial /sessions/me documentation

joelfranusic-okta:


_source/docs/api/resources/sessions.md
```diff
@@ -410,3 +410,95 @@ curl -v -X DELETE \
 ~~~ http
 HTTP/1.1 204 No Content
 ~~~
+
+
+#### Get session for current user
```
karlmcguinness-okta: nit: Get Current Session

_source/docs/api/resources/sessions.md
```diff
@@ -410,3 +410,95 @@ curl -v -X DELETE \
...
+{:.api .api-operation}
+
+<span class="api-uri-template api-uri-get"><span class="api-label">GET</span> /sessions/me</span>
+<span class="api-label api-label-cors pull-right"><i class="fa fa-cloud-download"></i> CORS</span>
+
+
+Get session information for the current user. Use this method in a browser based application to determine if the user is logged in.
+
```
karlmcguinness-okta: Need to add Response Parameters that indicates that a 404 will be returned if the user's session does not exist

karlmcguinness-okta:
There is an implementation detail we need to figure out how to addresss.  The /sessions/me endpoint returns and external session id that can't be used in a HTTP header why the legacy endpoints used the internal session id that was the same value as the cookie.  This means the sid=seession_id cooke in the /sessions/me call will not match the session id in the response body (which is fine since this should be considered implementation detail and hidden from developers).

The problem is that the  the legacy "extend session" i.e. PUT /api/v1/sessions/<id> only supports the original sessionId (not externalId)

and we added a new "refresh session": POST api/v1/sessions/<externalId>/lifecycle/refresh which only accepts the externalId

Any suggestion on how to articulate this?

_source/docs/api/resources/sessions.md
```diff
@@ -410,3 +410,95 @@ curl -v -X DELETE \
...
+~~~ json
+{
+    "_links": {
+        "refresh": {
+            "hints": {
+                "allow": [
+                    "POST"
+                ]
+            },
+            "href": "https://example.okta.com/api/v1/sessions/012a34BCDeFGhi56jklMNO_pQ/lifecycle/refresh"
```
karlmcguinness-okta: Need to doc this operation /api/v1/sessions//lifecycle/refresh

_source/docs/api/resources/sessions.md
```diff
@@ -410,3 +410,95 @@ curl -v -X DELETE \
...
+-H "Cookie: sid=${okta_session_id}"
+"https://${org}.okta.com/api/v1/sessions/me"
+~~~
+
+##### Response Example
+{:.api .api-response .api-response-example}
+
+~~~ json
+{
+    "_links": {
```
karlmcguinness-okta: nit: for consistency can we have the JSON HAL stuff and the end of the body like the other examples?

karlmcguinness-okta:
We should also have an updated Postman collection for this operation and a note about using the Postman Interceptor plugin to make AJAX calls from Postman in the browser context with the browser's session cookies (API Client Test doc)

joelfranusic-okta:
Addressed all of @karlmcguinness-okta feedback above, added Postman collection. Ready for review.

cwu-okta:
@tthompson-okta

_source/docs/api/resources/sessions.md
```diff
@@ -410,3 +414,141 @@ curl -v -X DELETE \
...
+
+As noted above, `id` in the example below can only be used with the [Refresh Session](#refresh-session) endpoint.
+
+~~~ json
+{
+    "amr": [
+        "pwd"
+    ],
+    "expiresAt": "2016-01-03T09:13:17.000Z",
+    "id": "012a34BCDeFGhi56jklMNO_pQ",
```
tthompson-okta: Nit: would be nice if the attributes were ordered in the way we typically order them: 
1. id
2. status
3. expiresAt
4. userId
5. login
6. lastPasswordVerification
7. lastFactorVerification
8. amr
9. idp
10. mfaActive

_source/docs/api/resources/sessions.md
```diff
@@ -410,3 +414,141 @@ curl -v -X DELETE \
...
+~~~ sh
+curl -v -X GET \
+-H "Cookie: sid=${okta_session_id}"
+"https://${org}.okta.com/api/v1/sessions/me"
+~~~
+
+##### Response Examples
+{:.api .api-response .api-response-example}
+
+`404 Not Found` status code is returned when a current session can not be found.
```
tthompson-okta: Nit: I'd change `current` to `active`.. its very possible we do find a session, but if its expired we'll still return a 404.

_source/docs/api/resources/sessions.md
```diff
@@ -410,3 +414,141 @@ curl -v -X DELETE \
...
+~~~
+
+##### Response Example
+{:.api .api-response .api-response-example}
+
+~~~ http
+HTTP/1.1 204 No Content
+~~~
+
+### Refresh Session
```
tthompson-okta: We probably need a way to indicate that this supersedes the "Extend Session" operation described earlier in this document. I would put a "deprecated" marker up there, and then a link to this section as the preferred mechanism.

_source/docs/api/resources/sessions.md
```diff
@@ -410,3 +414,141 @@ curl -v -X DELETE \
...
+#### Get Current Session
+{:.api .api-operation}
+
+<span class="api-uri-template api-uri-get"><span class="api-label">GET</span> /sessions/me</span>
+<span class="api-label api-label-cors pull-right"><i class="fa fa-cloud-download"></i> CORS</span>
+
+
+Get session information for the current user. Use this method in a browser based application to determine if the user is logged in.
+
+> The session `id` returned from `/sessions/me` is an `externalId` and can only be used with the [Refresh Session](#refresh-session) endpoint.
```
tthompson-okta: I actually don't think we should call out the "externalId" concept in these docs anywhere. As far as the API developer is concerned, it's just the "id". Also note that this whole API now accepts either externalId or regular Id everywhere (except for the deprecated `Extend Session` call, which is replaced by `Refresh Session`).

tthompson-okta: The one caveat to this is the section at the top about **Create Session with SessionToken**. 
> Creates a new session for a user with a valid session token.  Only use this operation if you need the session `id`, otherwise you should use one of the following flows to obtain a SSO session with a `sessionToken`

We have an unsolved backwards compat problem here with our own selves: the Okta Mobile app actually takes the returned session 'id' from this particular call and manually sets the Okta 'sid' cookie itself with it.. so we cannot yet switch this particular response to give back the externalId. @tbelote-okta @umangshah-okta FYI: we probably should add a special query param for Okta Mobile to continue to get the raw 'sid' value back, and then change default API behavior to start returning the externalId.

So the implication here is that currently if you call Create Session with SessionToken and get the "sessionId" back in the response, you have to use the deprecated "Extend Session" API rather than the new "Refresh Session" call. This is a little goofy; but I think we can't keep that mess out of these docs for now. @karlmcguinness-okta has a plan to reconcile this issue I'm sure.

mystiberry-okta:
@joelfranusic-okta should I be tracking this, or will you merge when it's complete? Thanks!

cwu-okta:
@joelfranusic-okta @mystiberry-okta what's the status on this?

joelfranusic-okta:
I need to hand this off to @mystiberry-okta - Mysti, what is the best way to do that? Shall I open a Jira?

cwu-okta:
OKTA-74731 was already created and tracking this pull request.

mystiberry-okta:
I’ve taken the JIRA, and it looks like we’re waiting for review from Karl, though I need to read Yu’s comments, maybe he supplied the needed info. I don’t have this issue loaded in my head right now.

Mysti
On Jun 9, 2016, at 11:03 AM, cwu-okta <notifications@github.com<mailto:notifications@github.com>> wrote:


OKTA-74731 was already created and tracking this pull request.

mystiberry-okta:
@karlmcguinness-okta and @tthompson-okta  I lack the domain knowledge to issue a ship-it. 
Are we okay with the last commit or are more edits needed?

tthompson-okta:
@mystiberry-okta no, I think a few of the open comments (around mention of "externalId") still need to be addressed.

tthompson-okta:
Closing in favor of #355

PR #148, opened 2016-01-08
Update users.md

gabrielsroka:
lines 1356 & 1357:
- [Update Profile - Strict (PUT)](#update-profile)
- [Update Profile - Partial (POST)](#update-profile-1)

will probably need new anchors since i renamed sections on lines 1386 & 1487 (because they were both called "Update Profile"

### DIFF
--- a/_source/docs/api/resources/users.md
+++ b/_source/docs/api/resources/users.md
@@ -1353,6 +1353,11 @@ curl -v -X GET \
 ### Update User
 {:.api .api-operation}
 
+- [Update Profile - Strict (PUT)](#update-profile)
+- [Update Profile - Partial (POST)](#update-profile-1)
+- [Set Password](#set-password)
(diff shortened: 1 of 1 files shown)


mystiberry-okta:
@gabrielsroka should I merge this? Not sure if it's complete, and if the features should be mentioned in the doc yet. thanks!

karlmcguinness-okta:
I didn't merge as I wasn't sure we wanted to doc this way. I think the feedback here is that we need to do a better job explaining POST vs PUT but not sure the best way to document going forward.  This problem exists for other resourcees

mystiberry-okta:
@karlmcguinness-okta Is this on my plate to drive to resolution, or...? Thanks!

mystiberry-okta:
Closing this and moved issue to https://oktainc.atlassian.net/browse/OKTA-86407

PR #152, opened 2016-01-14
Code correction for -> foreach

raphaellondner-okta:
The for loop in the Enumerations section is not valid in C# for lists that support the IEnumerable interface (and should be a foreach loop)


karlmcguinness-okta:
how are the htm files generated?  just want to make sure the source is updated as well.

karlmcguinness-okta:
@raphaellondner-okta Is this a fix to the source or the generated html?

mystiberry-okta:
@raphaellondner-okta and @karlmcguinness-okta  should I merge this? Doc does the merges and then pushes the changed HTML to the site once a week for Preview (or at other times for off-cycle activity). 

Thanks!

mystiberry-okta:
Closing this as it appears to be a fix on the output files, not the source files. @raphaellondner-okta  I don't know how to edit the "javadoc" or I'd do this for you. Perhaps @eklein-okta or someone can teach us?

PR #154, opened 2016-01-14
Add login field into the session model in seesions.md

yuliu-okta:
Relates-To: OKTA-76767

Primary Reviewer: @joelfranusic-okta 
Reviewer: @federations-okta 
FYI: @mattloughran-okta @eklein-okta  (we'll need your help with the merge once Joel gives it a shipit)

### DIFF
--- a/_source/docs/api/resources/sessions.md
+++ b/_source/docs/api/resources/sessions.md
@@ -32,6 +32,7 @@ Okta provides a very rich [Authentication API](./authn.html) to validate a [user
 {
   "id": "000najcYVnjRS2aZG50MpHL4Q",
   "userId": "00ubgaSARVOQDIOXMORI",
+  "login": "user@example.com",
   "expiresAt": "2015-08-30T18:41:35.818Z",
   "status": "ACTIVE",
   "lastPasswordVerification": "2015-08-30T18:41:35.818Z",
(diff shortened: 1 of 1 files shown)


_source/docs/api/resources/sessions.md
```diff
@@ -87,6 +88,7 @@ Sessions have the following properties:
 | ------------------------ | --------------------------------------------------------------------------------------------- | ----------------------------------------- | -------- | ------ | -------- | --------- | --------- | ---------- |
 | id                       | unique key for the session                                                                    | String                                    | FALSE    | TRUE   | TRUE     |           |           |            |
 | userId                   | unique key for the [user](users.html#get-user-with-id)                                        | String                                    | FALSE    | TRUE   | TRUE     |           |           |            |
+| login                    | unique identifier in the [user profile](users.html#default-profile-properties)                | String                                    | FALSE    | TRUE   | TRUE     |           |           |            |
```
mohitchugh-okta: Might be better to use the same definition from the link you provided, like - "unique identifier for the user (username)", but I'll defer to @joelfranusic-okta

joelfranusic-okta: Agreed @mohitchugh-okta - since these are the same values, they should share the same doc strings and properties.

yuliu-okta: @mohitchugh-okta @joelfranusic-okta done, changed to the same doc string and properties

joelfranusic-okta:
LGTM :shipit:

_source/docs/api/resources/sessions.md
```diff
@@ -87,6 +88,7 @@ Sessions have the following properties:
 | ------------------------ | --------------------------------------------------------------------------------------------- | ----------------------------------------- | -------- | ------ | -------- | --------- | --------- | ---------- |
 | id                       | unique key for the session                                                                    | String                                    | FALSE    | TRUE   | TRUE     |           |           |            |
 | userId                   | unique key for the [user](users.html#get-user-with-id)                                        | String                                    | FALSE    | TRUE   | TRUE     |           |           |            |
+| login                    | unique identifier for the [user](users.html#get-user-with-id)                                 | String                                    | FALSE    | TRUE   | TRUE     |           |           |            |
```
tthompson-okta: Nit: this definition would lead you to believe its the user **ID** rather than the user **login**. Not a big deal, but a more correct definition would be: `"login name (username) for the user"`. 

I did notice that it's defined this way already on the User API doc, but I'd argue that should be reworded too: http://developer.okta.com/docs/api/resources/users.html#profile-object

PR #159, opened 2016-01-22
Okta 77563 mchugh oidc

mohitchugh-okta:
Added documentation to the OIDC endpoints.

Still missing links to public-client app documentation which is not ready yet.

Removed Okta specific error codes for now for the authorize endpoint.

Primary reviewer: @karlmcguinness-okta 
Reviewers: @joelfranusic-okta 

@eklein-okta @mattloughran-okta I think Karl's got some changes planned before it'll come your way, but FYI.

### DIFF
--- a/_source/docs/api/resources/oidc.md
+++ b/_source/docs/api/resources/oidc.md
@@ -1,0 +1,416 @@
+---
+layout: docs_page
+title: OpenId Connect
+---
+
+# Overview
+
+The OpenID Connect API endpoints enable clients to use [OIDC workflows](http://openid.net/specs/openid-connect-core-1_0.html) with Okta.
+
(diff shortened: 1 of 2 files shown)


mohitchugh-okta:
@eklein-okta @mattloughran-okta - Based on a conversation with @karlmcguinness-okta , we are ok with using this draft for the beta. 

Please go ahead with your review/cleanup and publish it without linking it from the side-bar (the way you've done with http://developer.okta.com/docs/api/resources/idps.html) so that @joelfranusic-okta can start referring to this document in the usage documentation he’s working on.

Note that we do expect this document to evolve (we've taken out some error-codes that we aren't sure about and Karl is going to do some rearranging), but that can happen after this document is published.

@cwu-okta FYI.

PR #173, opened 2016-02-03
Include prerequisites in getting started steps

sethp-jive:
On my OS X workstation I found that following the listed steps presented me with the following error:

```
saml2.sigver.SigverError
SigverError: Can't find ['xmlsec1']
```

That problem was remedied by running `brew install libxmlsec1`, but as a complete SAML neophyte it was not obvious that the `xmlsec1` PySAML could not find was a missing system-level dependency.


joelfranusic-okta:
Cool! Thanks!

PR #174, opened 2016-02-03
Changes IdP transaction docs to reflect code changes.

danielzeller-okta:
In OKTA-78503 we renamed the `FINISH` status to `SUCCESS`, and return a `next` link instead of a `finish` link when the status is `SUCCESS`.

In OKTA-78504 we moved the `/tx/{id}/lifecycle/jit` endpoint to `/tx/{id}/lifecycle/provision`, and renamed the HAL links accordingly.

@mattloughran-okta @joelfranusic-okta @slee-okta @eklein-okta 
@federations-okta


mattloughran-okta:
Looks good!

PR #178, opened 2016-02-11
Update authn.md

gabrielsroka:


karlmcguinness-okta:
Are you sure this is correct? My UD profile shows `timezone`?

PR #183, opened 2016-02-16
OKTA-79132 Add cleaned up error codes back in the OIDC doc

mohitchugh-okta:
Cleaned up the code as per OKTA-79132.

Made changes to both oidc and error_codes.md

Resolves: OKTA-79132

Primary reviewer: @joelfranusic-okta 
Reviewers: @tthompson-okta 

FYI - @mystiberry-okta

### DIFF
--- a/_source/docs/api/getting_started/error_codes.md
+++ b/_source/docs/api/getting_started/error_codes.md
@@ -156,10 +156,10 @@ This document provides further information about the errors that the Okta API re
 
 In stituations where Okta needs to pass an error to a downstream application via a `redirect_uri`, the error code and description will be encoded as the query parameters `error` and `error_description`.
 
-> For example, if the **redirect_uri** is **https://example.com**, then the **UNKNOWN_ENDPOINT** error would be passed as follows:
+> For example, if the **redirect_uri** is **https://example.com**, then the **ACCESS_DENIED** error would be passed as follows:
 > 
 > ~~~
-> https://example.com?error=UNKNOWN_ENDPOINT&error_description=Unknown%20social%20endpoint.
+> https://example.com?error=access_denied&error_description=The%20resource%20owner%20or%20authorization%20server%20denied%20the%20request.
 > ~~~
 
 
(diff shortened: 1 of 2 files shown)


_source/docs/api/resources/oidc.md
```diff
@@ -187,13 +187,15 @@ unsupported_response_mode  | The specified response mode is invalid or unsupport
 invalid_scope  | The scopes list contains an invalid or unsupported value.    | 
 server_error  | The server encountered an internal error.    | 
 temporarily_unavailable  | The server is temporarily unavailable, but should be able to process the request at a later time.    |
+invalid_request | The request is missing a necessary parameter or the parameter has an invalid value. |
+invalid_client  | The specified client id is invalid.
+access_denied   | The server denied the request.
```
tthompson-okta: What about invalid_grant? I thought you mentioned we were currently returning that in some places.

tthompson-okta:
Ah, I actually intended to update this page: http://developer.okta.com/docs/api/getting_started/error_codes.html#openid-connect-and-okta-social-authentication

I think the file is called `_source/docs/api/getting_started/error_codes.md`

_source/docs/api/resources/oidc.md
mohitchugh-okta: we only return it for the token endpoint

mohitchugh-okta:
I can update that page as well, but that's a lot of duplication. Isn't it better to have the error-codes next to the API that is calling it rather than provide a list of out of context?

@karlmcguinness-okta FYI as well.

mystiberry-okta:
The customer may receive an error and not know which object/operation it's associated with, yes?


Mysti


mohitchugh-okta:
I'm not sure. Karl and Joel would know more about the usage pattern of the SDK. I'm just worried that these two lists will inevitably go out of sync and end up causing more confusion. 

@joelfranusic-okta Is there a way to pull in the error codes in the two places from the same underlying master list?

mystiberry-okta:
or we can link to one resource from the other location. I've been told there's no reuse in kramdown/markdown, but would love to be wrong!


Mysti


joelfranusic-okta:
I can think of at least two ways to reuse content in Jekyll. However I'm reluctant to fix it right now since a project that I'll be starting soon might make that all moot.

mohitchugh-okta:
Yup, that's why I've just gone ahead and duplicated the codes for now. We'll deal with de-duplication later when we (hopefully) generate these codes directly from code.

_source/docs/api/resources/oidc.md
tthompson-okta: Ah right, I see it below in the token endpoint section.

_source/docs/api/getting_started/error_codes.md
```diff
@@ -156,7 +156,7 @@ This document provides further information about the errors that the Okta API re
 
 In stituations where Okta needs to pass an error to a downstream application via a `redirect_uri`, the error code and description will be encoded as the query parameters `error` and `error_description`.
 
-> For example, if the **redirect_uri** is **https://example.com**, then the **UNKNOWN_ENDPOINT** error would be passed as follows:
+> For example, if the **redirect_uri** is **https://example.com**, then the **ACCESS_DENIED** error would be passed as follows:
 > 
 > ~~~
 > https://example.com?error=UNKNOWN_ENDPOINT&error_description=Unknown%20social%20endpoint.
```
tthompson-okta: You need to update the example here to use ACCESS_DENIED.

mohitchugh-okta: Yup. Fixed.

tthompson-okta:
LGTM. :sheep:

mohitchugh-okta:
@mystiberry-okta Can you please merge this PR in? Thanks.

mystiberry-okta:
Yes--can I rebuild site today or should I wait for Preview or Production?

mystiberry-okta:
@eklein-okta will publish the site on Thursday.

PR #185, opened 2016-02-22
_source/docs/api/resources/users.md

mystiberry-okta:
I've tried to incorporate all the content from Lars's great search_beta.md into users.md. 

1. This is my first month using both git and markdown/kramdown, so please don't be shy about calling out any mistakes. 
2. I'm not set up to test samples yet, so the new cURL sample hasn't been tested. I hope to correct that soon.
3. I tinkered with formatting of some unrelated tables by mistake, don't be surprised if you see a few tables with just a few changed spaces.

Related JIRA: https://oktainc.atlassian.net/browse/OKTA-77828


mystiberry-okta:
@tthompson-okta  I think you wanted to look at this too? Anyone else I should add?

_source/docs/api/resources/users.md
```diff
@@ -1039,13 +1040,129 @@ Enumerates users in your organization with pagination.  A subset of users can be
...
+search    | [Filter expression](/docs/api/getting_started/design_principles.html#filtering) for advanced user search | Query      | String   | FALSE    |
 
-> The `after` cursor should treated as an opaque value and obtained through the next link relation. See [Pagination](/docs/getting_started/design_principles.html#pagination)
+Parameter Details
 
-> Search currently performs a startsWith match but it should be considered an implementation detail and may change without notice in the future
+- Search with `q` currently performs a startsWith match but it should be considered an implementation detail and may change without notice in the future.
+- The `after` cursor should treated as an opaque value and obtained through the next link relation. See [Pagination](/docs/getting_started/design_principles.html#pagination).
+
+###### Advanced Search
```
karlmcguinness-okta: I think we need to do a better job explaining to the developer why we have both regular search and "advanced search"  Why are they different and what use cases can I do with A that I can do with B and  vice-versa.  This feature is going to cause confusion if we don't do a good job explaining scenarios

_source/docs/api/resources/users.md
```diff
@@ -1039,13 +1040,129 @@ Enumerates users in your organization with pagination.  A subset of users can be
...
+> The `advanced search` feature is [Beta](/docs/api/resources/users.html#list-users-with-advanced-search)
+
+Advanced search provides the option to filter on a variety of attributes:
+
+- Any user profile attribute
+- Any custom defined profile attribute
+- The top-level attributes `id`, `status`, `created`, `activated`, `statusChanged` and `lastUpdated` 
+
+Advanced search filters against all fields specified in the search parameter (case insensitive).
+The most up-to-date data is sometimes delayed for up to a few seconds, so a search may miss a recent change.
```
karlmcguinness-okta: This is super important detail and main reason why we don't have the same operation for regular and advanced search.  I think it warrants a callout with more details

_source/docs/api/resources/users.md
```diff
@@ -321,10 +321,10 @@ Creates a new user in your Okta organization with or without credentials.
...
-activate    | Executes [activation lifecycle](#activate-user) operation when creating the user | Query      | Boolean                                   | FALSE    | TRUE
-provider    | Indicates whether to create a user with a specified authentication provider      | Query      | Boolean                                   | FALSE    | FALSE
-profile     | Profile properties for user                                                      | Body       | [Profile Object](#profile-object)         | TRUE     |
-credentials | Credentials for user                                                             | Body       | [Credentials Object](#credentials-object) | FALSE    |
+activate    | Executes [activation lifecycle](#activate-user) operation when creating the user | Query      | Boolean                                   | FALSE    | TRUE   
+provider    | Indicates whether to create a user with a specified authentication provider      | Query      | Boolean                                   | FALSE    | FALSE  
+profile     | Profile properties for user                                                      | Body       | [Profile Object](#profile-object)         | TRUE     |
```
karlmcguinness-okta: There seems to be extra whitespace characters added in the PR, can we remove?

mystiberry-okta: will do!

_source/docs/api/resources/users.md
mystiberry-okta: @cwu-okta mentioned that as well. Who can I work with to develop that content?

_source/docs/api/resources/users.md
mystiberry-okta: Whom should I follow up with to develop this content? Thanks!

cwu-okta: I suggest working with PM (Hassen and Karl) for this.

_source/docs/api/resources/users.md
cwu-okta: Work with Hassen, Karl, and Stephen Lee

cwu-okta:
@larsjohansen-okta FYI

mystiberry-okta:
@cwu-okta @karlmcguinness-okta @larsjohansen-okta (fyi Lars since I'm deleting your original doc).
1. There's an area marking where I'll add the array info.
2. You might want to look at this in preview since there are so many changes.  Everything but typo corrections is under "List Users" section.

If this is okay, @cwu-okta let me know the other folks who should see this.

_source/docs/api/resources/users.md
```diff
@@ -1026,86 +1026,47 @@ curl -v -X GET \
...
+--------- | ----------------------------------------------------------------------------------------- | ---------- | -------- | -------- | -------
+q         | Filter on `firstName`, `lastName`, and `email` properties      | Query      | String   | FALSE    |
+filter    | Specifies the [Filter expression](/docs/api/getting_started/design_principles.html#filtering) | Query      | String   | FALSE    |
+search    | [Filter on](/docs/api/getting_started/design_principles.html#filtering) other properties | Query      | String   | FALSE    |
+limit     | Specifies the number of results returned                                                           | Query      | Number   | FALSE    | 200
+after     | Specifies the pagination cursor for the next page of users                                | Query      | String   | FALSE    |
 
-    filter=lastUpdated gt "2013-06-01T00:00:00.000Z" and lastUpdated lt "2014-01-01T00:00:00.000Z" and status eq "ACTIVE"
+The maximum value for `limit` is `200`, except for a few orgs that haven't yet been updated with the current limit.
```
karlmcguinness-okta: I would kill this line.  The max value is much hire (10,000+).  Whats changed is the default (GET with no params).  Some orgs are grandfathered in and have a large default while the standard going forward is 200 as the default.   Since there are only a few orgs that are grandfathered and would just omit this detail for now.

_source/docs/api/resources/users.md
```diff
@@ -1026,86 +1026,47 @@ curl -v -X GET \
...
+filter    | Specifies the [Filter expression](/docs/api/getting_started/design_principles.html#filtering) | Query      | String   | FALSE    |
+search    | [Filter on](/docs/api/getting_started/design_principles.html#filtering) other properties | Query      | String   | FALSE    |
+limit     | Specifies the number of results returned                                                           | Query      | Number   | FALSE    | 200
+after     | Specifies the pagination cursor for the next page of users                                | Query      | String   | FALSE    |
 
-    filter=lastUpdated gt "2013-06-01T00:00:00.000Z" and lastUpdated lt "2014-01-01T00:00:00.000Z" and status eq "ACTIVE"
+The maximum value for `limit` is `200`, except for a few orgs that haven't yet been updated with the current limit. 
 
-Users updated after 06/01/2013 but with a status of `LOCKED_OUT` or `RECOVERY`
+  * We recommend everyone set the limit to 200 or less, as we expect to bring all orgs to the same default value
```
karlmcguinness-okta: Similar comment as before

_source/docs/api/resources/users.md
```diff
@@ -1210,10 +1171,17 @@ Link: <https://your-domain.okta.com/api/v1/users?after=00ud4tVDDXYVKPXKVLCO&limi
...
-#### List Users with Search
+#### Find Users
 {:.api .api-operation}
 
-Searches for user by `firstName`, `lastName`, or `email` value.  This operation is ideal for implementing a people picker. This operation does not support pagination.
+Use the `q` parameter to find users. The value of `q` is checked against the values in `firstName`, `lastName`, and `email`.  
+
+This operation:
+ 
+ * Is ideal for implementing a people picker.
```
karlmcguinness-okta: I think this may need a little more explanation.  Not sure "people picker" is an industry standard term.  I would combine this point with the last point such as 

Intended for simple lookup of users by name such as a people picker by matching against `firstName`, `lastName`, or `email`.

_source/docs/api/resources/users.md
```diff
@@ -1210,10 +1171,17 @@ Link: <https://your-domain.okta.com/api/v1/users?after=00ud4tVDDXYVKPXKVLCO&limi
...
+#### Find Users
 {:.api .api-operation}
 
-Searches for user by `firstName`, `lastName`, or `email` value.  This operation is ideal for implementing a people picker. This operation does not support pagination.
+Use the `q` parameter to find users. The value of `q` is checked against the values in `firstName`, `lastName`, and `email`.  
+
+This operation:
+ 
+ * Is ideal for implementing a people picker. 
+ * Doesn't support pagination.
```
karlmcguinness-okta: I forget the max results but we should document the limit with something like

Doesn't support pagination and will only return the first X matches

_source/docs/api/resources/users.md
```diff
@@ -1210,10 +1171,17 @@ Link: <https://your-domain.okta.com/api/v1/users?after=00ud4tVDDXYVKPXKVLCO&limi
...
+ * Queries live data.
```
karlmcguinness-okta: We haven't used this term before, so not sure how well it translates.  The important detail is that it will always query against most up-to-date data at the time of the query vs the replicated data that search uses.  You call if "live" is sufficient to express this nuance.

_source/docs/api/resources/users.md
```diff
@@ -1285,12 +1253,63 @@ curl -v -X GET \
...
+#### List Users with a Filter
 {:.api .api-operation}
 
-Enumerates all users that have been updated since a specific timestamp.  Use this operation when implementing a background synchronization job and you want to poll for changes.
+Lists all users that match the filter criteria.
 
-##### Request Example
+This operation:
+
+* Filters against live data. For example, if you create a user or change an attribute and then issue a filter request,
```
karlmcguinness-okta: I like how we attempt to define `live` data there.   This may also help with definition of `live` in the Find User operation

karlmcguinness-okta: Filters against a defined subset of user profile properties

_source/docs/api/resources/users.md
```diff
@@ -1285,12 +1253,63 @@ curl -v -X GET \
...
+`status eq "RECOVERY"`                         | Users that have a `status` of `RECOVERY`
+`status eq "PASSWORD_EXPIRED"`                 | Users that have a `status` of `PASSWORD_EXPIRED`
+`status eq "LOCKED_OUT"`                       | Users that have a `status` of `LOCKED_OUT`
+`status eq "DEPROVISIONED"`                    | Users that have a `status` of `DEPROVISIONED`
+`lastUpdated lt "yyyy-MM-dd'T'HH:mm:ss.SSSZ"`  | Users last updated before a specific timestamp
+`lastUpdated eq "yyyy-MM-dd'T'HH:mm:ss.SSSZ"`  | Users last updated at a specific timestamp
+`lastUpdated gt "yyyy-MM-dd'T'HH:mm:ss.SSSZ"`  | Users last updated after a specific timestamp
+`id eq "00u1ero7vZFVEIYLWPBN"`                 | Users with a specified `id`
+`profile.login eq "login@example.com"`         | Users with a specified `login`
+`profile.department eq "Engineering"`          | Users that have a `department` of `Engineering`
```
karlmcguinness-okta: not supported with filter.  this is a search use case.

_source/docs/api/resources/users.md
```diff
@@ -1285,12 +1253,63 @@ curl -v -X GET \
...
+`profile.occupation eq "Leader"`               | Users that have an `occupation` of `Leader`
```
karlmcguinness-okta: not supported with filter.  this is a search use case.

_source/docs/api/resources/users.md
```diff
@@ -1285,12 +1253,63 @@ curl -v -X GET \
...
+`profile.email eq "email@example.com"`         | Users with a specified `email`*
+`profile.firstName eq "John"`                  | Users with a specified `firstName`*
+`profile.lastName eq "Smith" `                 | Users with a specified `lastName`*
+
+\* If filtering by `email`, `lastName`, or `firstName`, Okta recommends the [User query API](#list-users-with-search). These profile filters are here for your convenience.
```
karlmcguinness-okta: Why?  Is this for Okta performance reasons?  Not sure the advantage for the developer

_source/docs/api/resources/users.md
```diff
@@ -1026,86 +1026,47 @@ curl -v -X GET \
...
-
-Users updated after 06/01/2013 but before 01/01/2014
-
-    filter=lastUpdated gt "2013-06-01T00:00:00.000Z" and lastUpdated lt "2014-01-01T00:00:00.000Z"
+Combine `'limit` and `after` with other relevant parameters.
 
-Users updated after 06/01/2013 but before 01/01/2014 with a status of `ACTIVE`
+Parameter | Description                                                                               | Param Type | DataType | Required | Default
+--------- | ----------------------------------------------------------------------------------------- | ---------- | -------- | -------- | -------
+q         | Filter on `firstName`, `lastName`, and `email` properties      | Query      | String   | FALSE    |
```
karlmcguinness-okta: The descriptions for q, filter, and search seem to mix up the terms.  I suggest we only use the term filter for the filter operation. Example

- q: Finds a user that matches `firstName`, `lastName`, or `email` properties 
- filter: Filters users with a supported expression for a subset of properties
- search: Searches for users with a supported expression for most properties

_source/docs/api/resources/users.md
```diff
@@ -1298,9 +1317,11 @@ curl -v -X GET \
 -H "Accept: application/json" \
 -H "Content-Type: application/json" \
 -H "Authorization: SSWS ${api_token}" \
-"https://${org}.okta.com/api/v1/users?filter=lastUpdated+gt+\"2013-07-01T00:00:00.000Z\""
+"https://${org}.okta.com/api/v1/users?filter=status+eq+\"ACTIVE\"+or+status+eq+\"RECOVERY\""
 ~~~
 
+> Users with a status of `DEPROVISIONED` are not enumerated by default and must be explicitly requested with a status filter.
+
 ##### Response Example
```
karlmcguinness-okta: Pair the request and response.

Request Example: Status
Response Example: Status

_source/docs/api/resources/users.md
```diff
@@ -1430,7 +1446,105 @@ curl -v -X GET \
...
+      }
+    }
+  }
+]
+~~~
+
+#### List Users with Search
+{:.api .api-operation}
+
+> Listing users with search is Beta. Changes may occur that break the current implementation. Don't use this feature in production.
```
karlmcguinness-okta: This is gone with EA right?  We should note this feature is EA and supported

_source/docs/api/resources/users.md
```diff
@@ -1430,7 +1446,105 @@ curl -v -X GET \
...
+
+Searches for user by the properties specified in the search parameter (case insensitive). 
+
+This operation:
+
+* Supports pagination.
+* Requires [URL encoding](http://en.wikipedia.org/wiki/Percent-encoding).
+For example, `search=profile.department eq "Engineering"` is encoded as `search=profile.department%20eq%20%22Engineering%22`.
+* Queries replicated data, so results don't always show the latest changes. 
+* Searches many properties:
```
karlmcguinness-okta: Just curious, is it easier to express what is not supported vs what is?  Do you prefer whitelist or backlist for developer docs for whats allowed?

_source/docs/api/resources/users.md
```diff
@@ -1430,7 +1446,105 @@ curl -v -X GET \
...
+   - Any user profile property
+   - Any custom-defined profile property
```
karlmcguinness-okta: I think this is covered by "Any user profile property"

karlmcguinness-okta:
Nice update,  I think it reads much better without the "Advanced Search" terminology.  I like were this is headed!  Added some feedback for my final round of comments.

cwu-okta:
@hassenkaraa-okta  FYI

_source/docs/api/resources/users.md
```diff
@@ -1285,12 +1253,63 @@ curl -v -X GET \
...
+`lastUpdated lt "yyyy-MM-dd'T'HH:mm:ss.SSSZ"`  | Users last updated before a specific timestamp
+`lastUpdated eq "yyyy-MM-dd'T'HH:mm:ss.SSSZ"`  | Users last updated at a specific timestamp
+`lastUpdated gt "yyyy-MM-dd'T'HH:mm:ss.SSSZ"`  | Users last updated after a specific timestamp
+`id eq "00u1ero7vZFVEIYLWPBN"`                 | Users with a specified `id`
+`profile.login eq "login@example.com"`         | Users with a specified `login`
+`profile.department eq "Engineering"`          | Users that have a `department` of `Engineering`
+`profile.occupation eq "Leader"`               | Users that have an `occupation` of `Leader`
+`profile.email eq "email@example.com"`         | Users with a specified `email`*
+`profile.firstName eq "John"`                  | Users with a specified `firstName`*
+`profile.lastName eq "Smith" `                 | Users with a specified `lastName`*
```
cwu-okta: We should also contain an example of 'sw' (starts with) to indicate we support this operator. We support this for both the filter and search query params (list users with filter and users with search)

Unless @karlmcguinness-okta disagrees to add this addition.

cwu-okta:
I suggest moving the array info into the section describing filter with search. We only support searching arrays for the filter with search syntax. It would be good to include both an example and explanation for how search is done with arrays.

_source/docs/api/resources/users.md
cwu-okta: :+1:

_source/docs/api/resources/users.md
mystiberry-okta: Great question! I wasn't thinking of these as whitelist/blacklist items,  It's usually easier on the reader to write positive, not negative sentences, unless we think the reader is going to expect something to be true and we want to make clear that it is not.


Also, if there are a hundred things that aren't allowed and three that are, it's easier for reader (and more likely to stay correct) to read the short list.


I can rreorder/rewrite this list...was even thinking about a table but the table got ugly...



_source/docs/api/resources/users.md
mystiberry-okta: I couldn't find anyone who uses a different phrase than people picker. I did want to keep the functional definition (what is checked) from the higher-level "why use this" statement. Is this okay, or is there something wrong in the way I phrased the 2nd sentence?

Use the `q` parameter for a simple lookup of users by name, for example when creating a people picker.
The value of `q` is matched against `firstName`, `lastName`, or `email`.

_source/docs/api/resources/users.md
mystiberry-okta: asking Lars for limit. I know I should be able to dive into the code and find it myself, but I'm not there yet.

_source/docs/api/resources/users.md
mystiberry-okta: Your description is less ambiguous. Going with it.

_source/docs/api/resources/users.md
karlmcguinness-okta: fine with me

_source/docs/api/resources/users.md
cwu-okta: If you are a non-grandfathered org it is 200. If you are a grandfathered org it is 10k. Yes, crazy, I know. We should use similar language we've used in the past about the page size and how it can change so don't count on it.

mystiberry-okta: @cwu-okta & @karlmcguinness-okta , @larsjohansen-okta  seems like either I'm confused or the feedback's inconsistent. Is this correct:

limit default value: 200 (what about grandfathered orgs?)
limit max value: 200 except for those grandfathered orgs, who have 10k.

>Hi Mysti,

The limit is 200 if LIMIT_USER_API_RESULTS feature flag is enabled, else it is 10000.

Lars K. Johansen

karlmcguinness-okta: @larsjohansen-okta  define limit.   It just appears to be the default and not a limit. I can request more.

mystiberry-okta: Cindy emailed me this:

If you do not specify any limit AND you specify a query, then it is defaulted to a limit of 10.  If you do specify a limit and a query, then we cap the limit to 200 or 10k, depending if you are grandfathered or not.

However, if you do not specify a query then default is 200 or 10k, depending if you are grandfathered or not.

_source/docs/api/resources/users.md
mystiberry-okta: Do customers have to contact Support to get it enabled (like UI features)?

cwu-okta: Yes. For any feature which is not GA the customer must contact support and/or PS/Stephen Lee.

mystiberry-okta:
Final draft changes in! @karlmcguinness-okta @cwu-okta @larsjohansen-okta  This is the same as what was just sent round in RTF (plus a few typo fixes).

_source/docs/api/resources/users.md
```diff
@@ -1026,86 +1026,44 @@ curl -v -X GET \
...
+
 Parameter | Description                                                                               | Param Type | DataType | Required | Default
 --------- | ----------------------------------------------------------------------------------------- | ---------- | -------- | -------- | -------
-q         | Searches `firstName`, `lastName`, and `email` properties of users for matching value      | Query      | String   | FALSE    |
-limit     | Specified the number of results                                                           | Query      | Number   | FALSE    | 200
-filter    | [Filter expression](/docs/api/getting_started/design_principles.html#filtering) for users | Query      | String   | FALSE    |
+q         | Finds a user that matches `firstName`, `lastName`, and `email` properties      | Query      | String   | FALSE    |
+filter    | [Filters](/docs/api/getting_started/design_principles.html#filtering) users with a supported expression for a subset of properties | Query      | String   | FALSE    |
+search    | Searches for users witha  supported expression for most properties  | Query      | String   | FALSE    |
```
cwu-okta: Nit: should be "with a"

mystiberry-okta: thanks! fixed.


_source/docs/api/resources/users.md
```diff
@@ -1430,13 +1443,199 @@ curl -v -X GET \
...
+    }
+  }
+]
+~~~
+
+#### List Users with Search
+{:.api .api-operation}
+
+> Listing users with search is an Early Access feature which you can request from Okta Support.
+It's been tested to the same level as Generally Available (GA) features, so you  can use it in a production environment.
```
cwu-okta: I don't think we typically go to this level of detail describing what EA vs GA is. That type of description belongs on service cloud. Here i'd just say it is an EA feature.

_source/docs/api/resources/users.md
```diff
@@ -1430,13 +1443,199 @@ curl -v -X GET \
...
+
+> Listing users with search is an Early Access feature which you can request from Okta Support.
+It's been tested to the same level as Generally Available (GA) features, so you  can use it in a production environment. 
+
+Searches for user by the properties specified in the search parameter (case insensitive). 
+
+This operation:
+
+* Supports pagination.
+* Requires [URL encoding](http://en.wikipedia.org/wiki/Percent-encoding).
```
cwu-okta: This is no different than any other request against the API so I suggest leaving it out OR moving this to a top level section.

_source/docs/api/resources/users.md
```diff
@@ -1430,13 +1443,199 @@ curl -v -X GET \
...
+For example, `search=profile.department eq "Engineering"` is encoded as `search=profile.department%20eq%20%22Engineering%22`.
+* Queries data from a replicated store, so changes aren't always immediately available in search results. 
+Don't use search results directly for record updates, as the data might be stale and therefore overwrite newer data (data loss).
```
cwu-okta: Nit: I think it is ```(data loss.)```

_source/docs/api/resources/users.md
```diff
@@ -1430,13 +1443,199 @@ curl -v -X GET \
...
+   `lastUpdated gt "yyyy-MM-dd'T'HH:mm:ss.SSSZ"`  | Users last updated after a specific timestamp
+   `id eq "00u1ero7vZFVEIYLWPBN"`                 | Users with a specified `id`
+   `profile.department eq "Engineering"`          | Users that have a `department` of `Engineering`
+   `profile.occupation eq "Leader"`               | Users that have an `occupation` of `Leader`
+   `profile.lastName sw "Sm" `                    | Users whose `lastName` starts with "Sm"
+
+##### Searching Arrays
+
+You can search properties that are arrays. If any element matches the search term, the entire array (object) is returned. 
+For examples, see [Request Example for Array](#request-example-for-array) and [Response Example for Array](#response-example-for-array).
```
cwu-okta: Nit: Why not collapse the "see request example and response example" into "see request example" ? Once you see the example a user might naturally follow it through to view the response. Minor comment though and I don't feel strongly if you leave this as is.

_source/docs/api/resources/users.md
```diff
@@ -1430,13 +1443,199 @@ curl -v -X GET \
...
 
+##### Request Example for Array
+{:.api .api-request .api-request-example}
+
+~~~sh
+curl -v -X GET \
+-H "Accept: application/json" \
+-H "Content-Type: application/json" \
+-H "Authorization: SSWS ${api_token}" \
+"https://${org}.okta.com/api/v1/users?search=profile.arrayAttr eq "arrayAttrVal1" "
```
cwu-okta: Consider setting the stage for this request. Consider suggesting you have an array called arrayAttr which contains values ["arrayAttrVal1", "arrayAttrVal2"...]. This lets the reader have the context you have a User object with an array attribute called ```arrayAttr``` which is a string array

PR #188, opened 2016-03-03
Adding callout about escaping custom attributes for OKTA-80022

keithchin-okta:
@karlmcguinness-okta 
Does this look ok? Or anyone else I should add to this PR?  Thanks!


keithchin-okta:
@mystiberry-okta Fixed wording to match

PR #194, opened 2016-03-15
Remove security image

joelfranusic-okta:


joelfranusic-okta:
@raphaellondner-okta can you review?

PR #198, opened 2016-03-20
Added expired password/AD info

mystiberry-okta:
@georgekwon-okta @huzhou-okta Hoping this is correct now! Let me know: 1. Should this note also be in the admin doc and 2. Is it correct/complete?

Thanks!


mystiberry-okta:
Draft approved.

PR #199, opened 2016-03-22
minor updates to provisioning section

dhorvath:
bold / spacing


dhorvath:
@mystiberry-okta made a few other updates here. broken links added links, etc.

PR #215, opened 2016-04-06
Change ToC for dev site

mystiberry-okta:
@karlmcguinness-okta @tthompson-okta  and @larsjohansen-okta , this will modify the Table of Contents as discussed. Here's what it will look like--is that right? Lars, FYI, social_authentication.md will move to the API resources as shown in this pic. @joelfranusic-okta and @eklein-okta might also be interested.
toc


mystiberry-okta:
Oh, my redirect didn't work, but the new location should work fine. I'll work on the redirect.

mystiberry-okta:
@karanharibhakti-okta @tthompson-okta @larsjohansen-okta this is what's being published this afternoon (Lars, I moved social_authentication.md and this last commit is the redirect, which now works in my local build).

_source/docs/guides/social_authentication.md
```diff
@@ -1,133 +1,8 @@
----
```
tthompson-okta: So does all this content already exist under the API "resources" section? It seems like this PR is removing it instead of moving it.

mystiberry-okta: Yes, I already merged the file into /resources (bad practice, sorry about that!).


This has the old version being replaced by a redirect, and in the same file in the new location, /resources, I addd the one line to complete the redirect.


Mysti


mystiberry-okta:
I need a ship it for this change? @cwu-okta @tthompson-okta or @karlmcguinness-okta

karlmcguinness-okta:
looks good, my only comment is for the Authentication->Authentication doc should we add API to the title or just call it API?

cwu-okta:
I'd put a vote in for former: authentication API



On Apr 6, 2016, at 5:37 PM, Karl <notifications@github.com<mailto:notifications@github.com>> wrote:


looks good, my only comment is for the Authentication->Authentication doc should we add API to the title or just call it API?

-

mystiberry-okta:
To me it's weird to label something in the API table of contents API, but if your spider senses tell you someone might land here, not pay attention to the context, and confuse that heading with admin-type authentication, I'll add it. typically it can confuse readers to label only one item in a set.

mystiberry-okta:
Ack, ignore previous post, Authentication API is perfect for the first article under label "Authentication" (no article). will get it in with sign-in widget EA removal this morning.

mystiberry-okta:
will change the title in another checkin (it's a different file).

cwu-okta:
I don’t feel strongly about it. Just thought having the word “API” when it doesn’t appear in any of the other TOC might throw people off. Perhaps ‘Reference’? Either way leaving it as you suggested is ok to me too.

PR #220, opened 2016-04-13
More info about okta_post_message for OKTA-83543

mystiberry-okta:
@tthompson-okta I added more content. I'm a little worried since it veers into "how to implement," but if you think it's helpful here, I'll just make sure it stays aligned with the how-to info elsewhere.

Also, check the language carefully--I did some edits for consistency and want to make sure I didn't change the meaning.

### DIFF
--- a/_source/docs/api/resources/oidc.md
+++ b/_source/docs/api/resources/oidc.md
@@ -180,17 +180,7 @@ nonce          | Specifies a nonce that is reflected back in the ID Token. Can b
       
       This response mode validates the `redirect_uri` from the client against Okta's client registration.
       If successful, it returns an HTML page which contains JavaScript to extract the ID token and send postMessage() back to the parent window. 
-      This page needs to know the `targetOrigin` in order for postMessage() to work; it will grab this from the redirect_uri passed in by the client (i.e. myApp.example.com). 
-
-      The pattern to configure the Okta Sign-in Widget:
-      
-      1. Register a client ID. Populate `redirect_uri` with the location of your SPA: https://myApp.example.com.
-      
-      2. Bootstrap the widget with the (Okta) `baseUrl,` `client_id,` and for social authentication, a list of IDP IDs.
-      
-      3. For social authentication, the widget should specify `response_mode=okta_post_message` as shown earlier.
-      
-      4. The `okta-post-message` page handles extracting the `id_token` and postMessage() back to the parent page where the Sign-In Widget is running.
+      This HTML page extracts the `id_token` and `targetOrigin` for postMessage() to consume. The `redirect_uri` passed in by the client (i.e. myApp.example.com) contains `targetOrigin`.
       
  * `state`: Always pass `state` with each authorize request to correlate the request and response. This correlation prevents and attacker from sending a different response (CSRF).
  For more information, see [this blog post about the importance of state in OAuth 2.0](http://www.twobotechnologies.com/blog/2014/02/importance-of-state-in-oauth2.html).
--- a/_source/docs/api/resources/oidc.md
+++ b/_source/docs/api/resources/oidc.md
@@ -177,6 +177,21 @@ nonce          | Specifies a nonce that is reflected back in the ID Token. Can b
       Okta Post Message is an adaptation of the [Web Message Response Mode](https://tools.ietf.org/html/draft-sakimura-oauth-wmrm-00#section-4.1). 
       This value provides a secure way for a single-page application to perform a sign-in flow 
       in a popup window or an iFrame and receive the ID token and/or access token back in the parent page without leaving the context of that page.
+      
+      This response mode validates the `redirect_uri` from the client against Okta's client registration.
+      If successful, it returns an HTML page which contains JavaScript to extract the ID token and send postMessage() back to the parent window. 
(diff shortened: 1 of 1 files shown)


tthompson-okta:
Needs some @karlmcguinness-okta love. I don't think we should go into login widget specifics but I do think we need the spec (i.e. what the contents of the postMessage() call will be) documented somehow.

mystiberry-okta:
@karlmcguinness-okta I took a stab at rewriting based on @tthompson-okta . Not sure if it's complete or correct for spec'ing the contents of postMessage().

mystiberry-okta:
@karlmcguinness-okta or @tthompson-okta what did you think of the update? More content still needed?

_source/docs/api/resources/oidc.md
```diff
@@ -177,6 +177,11 @@ nonce          | Specifies a nonce that is reflected back in the ID Token. Can b
       Okta Post Message is an adaptation of the [Web Message Response Mode](https://tools.ietf.org/html/draft-sakimura-oauth-wmrm-00#section-4.1). 
       This value provides a secure way for a single-page application to perform a sign-in flow 
       in a popup window or an iFrame and receive the ID token and/or access token back in the parent page without leaving the context of that page.
+      
+      This response mode validates the `redirect_uri` from the client against Okta's client registration.
```
tthompson-okta: I think we should replace this with:

https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage (add this link above, to the text `HTML Web Messaging`).

The data model for the [postMessage](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage) call is as follows:

* message:

(a JSON object with the following possible attributes):

Parameter | Description | Data Type
-------------- | --------------- | --------------
id_token | The ID Token JWT contains the details of the authentication event and the claims corresponding to the requested scopes. This is returned if the response_type includes `id_token`. | String | Only returned if the response types included 'id_token' |
access_token | The access_token that be used to access the userinfo endpoint. This is returned if the response_type included token. Unlike the ID Token JWT, the access_token structure is Okta internal only and is subject to change. | String
state | If the request contained a state parameter, then the same unmodified value is returned back in the response. | String
error | The error-code string providing information if anything went wrong. | String
error_description | Further description of the error. | String

* targetOrigin:

     `Specifies what the origin of 'parentWindow' must be for the event to be dispatched. The okta-post-message response mode always uses the origin from the 'redirect_uri' specified by the client. This is crucial to prevent the sensitive token data from being exposed to a malicious site.`

Here is a [sample app](https://github.com/mcguinness/okta-oidc-sample) which uses the Okta Post Message response mode.

mystiberry-okta: @tthompson-okta just double-checking, replace all three sentences, or just line 181?

This response mode validates the `redirect_uri` from the client against Okta's client registration.
      If successful, it returns an HTML page which contains JavaScript to extract the ID token and send postMessage() back to the parent window. 
      This HTML page extracts the `id_token` and `targetOrigin` for postMessage() to consume. The `redirect_uri` passed in by the client (i.e. myApp.example.com) contains `targetOrigin`.

mystiberry-okta:
@tthompson-okta I hit a kramdown limitation (wouldn't let me nest that table without hosing the content after it), so I tried a little restructure. Is it okay?

_source/docs/api/resources/oidc.md
```diff
@@ -173,13 +173,39 @@ nonce          | Specifies a nonce that is reflected back in the ID Token. Can b
...
-    * <em>okta_post_message</em> -- Uses HTML5 Web Messaging (for example, window.postMessage()) instead of the redirect for the authorization response from the authorization endpoint
+    * <em>okta_post_message</em> -- Uses [HTML5 Web Messaging](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage) (for example, window.postMessage()) instead of the redirect for the authorization response from the authorization endpoint.
       Okta Post Message is an adaptation of the [Web Message Response Mode](https://tools.ietf.org/html/draft-sakimura-oauth-wmrm-00#section-4.1). 
       This value provides a secure way for a single-page application to perform a sign-in flow 
       in a popup window or an iFrame and receive the ID token and/or access token back in the parent page without leaving the context of that page.
+      
+      If successful, the request returns an HTML page which contains JavaScript to extract the ID token and send postMessage() back to the parent window. 
+      This HTML page extracts the `id_token` and `targetOrigin` for postMessage() to consume. The `redirect_uri` passed in by the client (i.e. myApp.example.com) contains `targetOrigin`.
```
tthompson-okta: Sorry I wasn't clear; I think you should remove this whole paragraph (both sentences). The previous paragraph makes it clear that this is meant for popup windows and/or iFrames, and the data model below spells out the contract for postMessage(), so I don't think this paragraph is valuable anymore.

mystiberry-okta: Will do!

mystiberry-okta:
@tthompson-okta Sorry to bother you--I resolved all the conflicts, but am still not in a state where I can merge (even though I thought I did a manual merge after resolving conflicts).  Not sure how to complete this, assuming you approve the last changes.

mystiberry-okta:
Had to merge into master after conflict resolution.

PR #224, opened 2016-04-15
Require the state parameter on the authorize endpoint

danielzeller-okta:
Notes that the `state` parameter is no longer optional on the authorize endpoint.

@mystiberry-okta 
@tthompson-okta 
@federations-okta

### DIFF
--- a/_source/docs/api/resources/oidc.md
+++ b/_source/docs/api/resources/oidc.md
@@ -160,7 +160,7 @@ display           | Specifies how to display the authentication and consent UI. 
 max_age           | Specifies the allowable elapsed time, in seconds, since the last time the end user was actively authenticated by Okta. | Query      | String    | FALSE    | |
 response_mode     | Specifies how the authorization response should be returned. [Valid values: <em>fragment</em>, <em>form_post</em>, <em>query</em> or <em>okta_post_message</em>](#parameter-details). If <em>id_token</em> is specified as the response type, then <em>query</em> can't be used as the response mode. Default: Defaults to and is required to be <em>fragment</em> in implicit and hybrid flow. Defaults to <em>query</em> in authorization code flow. | Query        | String   | FALSE      | See Description.
 scope          | Can be a combination of <em>openid</em>, <em>profile</em>, <em>email</em>, <em>address</em> and <em>phone</em>. The combination determines the claims that are returned in the id_token. The openid scope has to be specified to get back an id_token. | Query        | String   | TRUE     | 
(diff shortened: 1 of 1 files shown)


mystiberry-okta:
ack. @danielzeller-okta was this a doc bug, or is it a change in behavior (and this commit just has the doc)? Thanks!

mystiberry-okta:
ack. Will merge next week. wondering if we need to warn the platform support person since we're going from optional to required in EA feature?

danielzeller-okta:
@cwu-okta ^

karlmcguinness-okta:
We missed this in the giant doc review last week.  The plan was to mark it required last week as we knew we were going to fix this week.

mystiberry-okta:
ack.

tthompson-okta:
We should add a callout with a brief description of why Okta chooses to make this required. Preferably with links to the detailed info from the specs:
* https://tools.ietf.org/html/rfc6749#section-10.12
* https://tools.ietf.org/html/rfc6819#section-3.6

Something like:
> Okta requires use of the OAuth 2.0 "state" parameter on all requests to the Authorization Endpoint in order to prevent cross-site request forgery (CSRF). The OAuth 2.0 specification [requires](https://tools.ietf.org/html/rfc6749#section-10.12) that clients protect their redirect URIs against CSRF by sending a value in the authorize request which binds the request to the user-agent's authenticated state. Using the "state" parameter is also a countermeasure to several other known attacks as outlined in the [OAuth 2.0 Threat Model and Security Considerations](https://tools.ietf.org/html/rfc6819).

danielzeller-okta:
We already have something like that here [here](http://developer.okta.com/docs/api/resources/oidc.html#parameter-details) which I will augment

karlmcguinness-okta:
@mystiberry-okta @danielzeller-okta Trevor's comment is a really good answer for state.  It does a great job justifying why we make this param required as Okta's implementation of CSRF protection.  I prefer we use links to the OAuth 2.0 Threat Model and Security Considerations vs the blog link we currently have as its a more normative answer.  The blog link is useful but probably doesn't make sense here in the OIDC reference doc.  

@cwu-okta @tthompson-okta We should consider writing an Okta blog post about OAuth security that summarizes a larger set of threats and mitigations that developers need to think about with OAuth and how Okta addresses them.  A more friendly version of http://www.oauthsecurity.com/

danielzeller-okta:
@karlmcguinness-okta @mystiberry-okta I've updated this branch with Trevors comment.

tthompson-okta:
⛵

mystiberry-okta:
Looks great, will merge for the next site update.

PR #227, opened 2016-04-28
Add Condition Expressions to Okta EL doc

eklein-okta:
@georgekwon-okta @mystiberry-okta 

@georgekwon-okta This is for OKTA-87666. I have attached a rendered version to this PR, but will also email it, just in case. 
conditional_expressions_el

Mysti, just FYI...


georgekwon-okta:
Comments:
-looks like we copied most of the content over from the Group Rules Beta doc, but this needs to be something that more general and not tied specifically do how these expressions are used in Group Rules.  For example we don't require an "IF" in front of these expressions.  We have a separate ternary operator for building IF/ELSE statements that return other values.  Group Rules is unique in that it interprets TRUE as a trigger to add users to a specific group.
-Need to specify how we handle logical groupings with parentheses
-I think the examples should show what they evaluate to, either TRUE or FALSE
-Typo in 4th bullet point "expressionscan" > "expressions can"

Here's how SpringEL documents Operators:
https://docs.spring.io/spring/docs/current/spring-framework-reference/html/expressions.html

Example Java doc:
http://www.tutorialspoint.com/java/java_basic_operators.htm

eklein-okta:
@georgekwon-okta @praveenprakash-okta  Here is the latest edit. If you can think of better samples, please let me know! Here is the rendered version. Please let me know if it is not helpful to attach it. Thanks!  -E

conditional_expressions_el_2

praveenprakash-okta:
String.len(middleInitial) > 0 ? fullName=String.join(params...) : fullName=String.join(params...) would not work, instead specify.

**fullName = String.len(middleInitial) > 0 ? String.join(params...) : String.join(params...)** where fullName can be the target field in mappings.

WRT the courtesyTitle example, as we discussed tdy morning, ternary operator wouldn't work with only true part. It should follow (condition)? expr1: expr2 format.

comment for fullName applies to additionalEmail ex as well.

I would suggest try these expressions in our mapping page or rules page and see if they work before putting it in the doc.

nit : 
Under **Several rules for specifying the condition**

1. How about 'Expressions must evaluate to boolean' ?, 'Expressions cannot contain an assignment operator'.

2. Attributes specified in the conditional expression should contain only available User or AppUser attributes.


In the last table last column, are they groupNames ?

eklein-okta:
Latest comments in!

eklein-okta:
@georgekwon-okta @praveenprakash-okta Good to go?

georgekwon-okta:
In this line:
[Condition] ? [Action if TRUE] : [Action if FALSE]

I would say "Value if TRUE" and "Value if FALSE" ?

Other than that, ship it!

eklein-okta:
Closing this PR without merging as 37 extra files got added to it. I will check in the Okta EL file to another PR and merge from there.

PR #228, opened 2016-05-03
Update social_authentication.md

Robingaal:
The requiered state parameter is missing from the documentation (it became requiered 2 weeks ago and was in the release notes, but for some reason documentation has not been updated). If not included Social auth will not work an Okta will just display it's normal login page. I've included it in the documentation.

### DIFF
--- a/_source/docs/api/resources/social_authentication.md
+++ b/_source/docs/api/resources/social_authentication.md
@@ -104,8 +104,14 @@ Advantages of using Okta Social Authentication include:
         After adding the `scope` and `response_type` parameters to
         your URL, it should look something like this:
         `https://example.okta.com/oauth2/v1/authorize?idp=0oa0bcde12fghiJkl3m4&client_id=AbcDE0fGHI1jk2LM34no&scope=openid%20email%20profile&response_type=id_token`
+
+    3.  Add the requiered `state` GET parameter to the Social Auth "Login
+        URL":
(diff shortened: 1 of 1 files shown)


Robingaal:
Guys what is the problem of committing this small update to the master branche? The community is trying to help you improving your documentation making it easier for others to adopt your platform so I would at expect something is being done with content or remarks delivered by customers / partners.

mystiberry-okta:
We're very grateful for your suggestion, and love input from the community.

FYI, we updated the documents (we had a few more changes to make than your suggestions, which are excellent!). Okay if I close this PR? The updates are live.

Robingaal:
Agree, thanks for updating the page. Much more clear now also in relation with the openid documentation. 

Still 1 think though; most of my customers want the social logon just for the app dashboard (return_url = {org}.okta.com) and now I still have to create a stub application so I can craft the url to start the login sequence, while in essence this application is not needed.

Normal Inbound Saml connection can be triggered by setting them as default Idp or using a specific url (not the very long crafted ones as you need for social). Are there any plans of making the social idp's behaving the same as "normal" saml2 idp's?

mcguinness:
hi @Robingaal, we intentionally disabled the Inbound SAML default IdP feature for Social Authentication as we didn't have clear use cases and most folks using social don't use the end-user experience in Okta. 

What scenario are you trying to achieve with Social?

mystiberry-okta:
Closed as the customer's excellent suggestions have been implemented via another PR.

Robingaal:
@mystiberry-okta Do I need a flag for this to have that function available? I still can't select is as "Default Identity Provider" at Security -> Identity Providers so I'm still using a stub application with a long crafted url with the return url specified as {org}.okta.com.

@mcguinness Usecase is a customer who offers the Okta page to it's own customers which only login via facebook (b2c case).

PR #229, opened 2016-05-03
Update oidc.md

Robingaal:
Refresh tokens are not yet supported by Okta. Removed it from the documentation since this is confusing for the developpers.

### DIFF
--- a/_source/docs/api/resources/oidc.md
+++ b/_source/docs/api/resources/oidc.md
@@ -380,7 +380,7 @@ Based on the grant type, the returned JSON contains a different set of tokens.
 
 Input grant type   | Output token types                    |
 -------------------|---------------------------------------|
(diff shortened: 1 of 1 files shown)


karlmcguinness-okta:
We support refresh tokens if you create a mobile or web OIDC app.  You need to check the `Allow Offline Access` checkbox to enable refresh tokens and pass the `offline_access` scope to your authorize request.  We are working on some UI updates to make this more clear.

Robingaal:
Thanks, adding the offline_access scope did the trick to my request. It's not being discussed anywhere in the Okta documentation (just found it in the oidc spec) so this should be added to it imo.

karlmcguinness-okta:
good feedback, we will add more clarity on this topic to our docs

PR #230, opened 2016-05-04
draft of changes for operators in search

mystiberry-okta:
@raphaellondner-okta  Here are the changes you requested about search operators. I'm verifying with Fed team as I'm not sure the list of non-supported items is correct.

### DIFF
--- a/_source/docs/api/resources/users.md
+++ b/_source/docs/api/resources/users.md
@@ -1055,6 +1055,8 @@ after     | Specifies the pagination cursor for the next page of users          
   * An HTTP 500 status code usually indicates that you have exceeded the request timeout.  Retry your request with a smaller limit and paginate the results. For more information, see [Pagination](/docs/getting_started/design_principles.html#pagination)).
   * Use `limit` and `after` with all four query types.
   * Treat the `after` cursor as an opaque value and obtain it through the next link relation. See [Pagination](/docs/getting_started/design_principles.html#pagination).
+  * We support most of the operators listed in the [SCIM Protocol Specification](https://tools.ietf.org/html/rfc7644#section-3.4.2.2). 
+    We don't support `co`, `ne`, `pr`, or `le`. See each parameter's section for example searches using supported operators.
 
 ##### Response Parameters
 {:.api .api-response .api-response-params}
@@ -1483,11 +1485,6 @@ Use an Id lookup for records that you update to ensure your results contain the 
    `profile.occupation eq "Leader"`               | Users that have an `occupation` of `Leader`
    `profile.lastName sw "Sm" `                    | Users whose `lastName` starts with "Sm"
 
-##### Searching Arrays
-
-You can search properties that are arrays. If any element matches the search term, the entire array (object) is returned. 
-For examples, see [Request Example for Array](#request-example-for-array) and [Response Example for Array](#response-example-for-array).
-
 ##### Search Examples
 
 List users with an occupation of `Leader`.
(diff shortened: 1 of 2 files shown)


_source/docs/api/resources/users.md
```diff
@@ -1055,6 +1055,8 @@ after     | Specifies the pagination cursor for the next page of users
   * An HTTP 500 status code usually indicates that you have exceeded the request timeout.  Retry your request with a smaller limit and paginate the results. For more information, see [Pagination](/docs/getting_started/design_principles.html#pagination)).
   * Use `limit` and `after` with all four query types.
   * Treat the `after` cursor as an opaque value and obtain it through the next link relation. See [Pagination](/docs/getting_started/design_principles.html#pagination).
+  * We support most of the operators listed in the [SCIM Protocol Specification](https://tools.ietf.org/html/rfc7644#section-3.4.2.2). 
+    We don't support `co`, `ne`, `pr`, or `le`. See each parameter's section for example searches using supported operators.
```
larsjohansen-okta: We do support 'pr' and 'le', so we shouldn't list that as unsupported.

cwu-okta: Do we really want to explicitly call out what we don't support? I'd suggest only calling out what we do support.

cwu-okta: Also, i thought you would make the update here: http://developer.okta.com/docs/api/getting_started/design_principles.html#filtering

This is the general list of filters supported. We don't support the above operators in any of the  resources.

mystiberry-okta: will make Lars' corrections, move it to design_principles, and link to that section from users.md, since I'm probably not the only person to miss this.

mystiberry-okta: Actually, @cwu-okta , in that doc we tell people to check the individual resource to verify which operators are actually supported.  Then for /users, we just point them back to the Operators topic.  So I need to say something.

cwu-okta: I was only trying to point out that we should remove the unsupported operators from http://developer.okta.com/docs/api/getting_started/design_principles.html#filtering since there isn't a single resource (today) which supports those filter operators.

mystiberry-okta: If they're unsupported in all resources, then yes! But there's this awful circular thing going on that I have to fix, where we say "here's the super set, check each resource for the subset" but then never document the subset. So not sure if I should correct the statement (all resources do support the same operator) or fix the gap (list the subset in each resource, not just /users.md).

mystiberry-okta: @larsjohansen-okta can you clarify that co and ne aren't supported for any resource, or just not supported by filter/search params on users?  Thanks!

cwu-okta: If you look at the emails and jiras I linked for you, we do not support co and ne anywhere. 'co' would be impossible since it causes a full DB table scan. 'ne' we just never did.

cwu-okta: I would leave that each resource indicates what it supports.

mystiberry-okta: Except that they don't (well, now /users will), but I will leave that note as is, and mark in users.md that co, ne not supported. Sorry for the confusion. Not sure how I missed this back in march.

mystiberry-okta: Just saw your note clarifying that we don't support those two anywhere. Will fix as you suggest. @larsjohansen-okta ignore my question.

mystiberry-okta:
Fixed.

mystiberry-okta:
https://oktainc.atlassian.net/browse/OKTA-85985

mystiberry-okta:
@raphaellondner-okta is this ready to ship?

raphaellondner-okta:
Yes, it's ready to ship as far as I'm concerned. Interestingly, I found out not are attribute operators not case-sensitive, but attribute values are not case-sensitive either (i.e. "john" and "John" work) WHEN USING THE SEARCH KEYWORD. However, attribute value are indeed case sensitive when using the "filter" keyword. Not sure if we want to mention this or let users find it out on their own.

mystiberry-okta:
I'll open a new JIRA for that. I want to find out if that's working as expected before doc'ing it (should be doc'd!).

PR #231, opened 2016-05-04
Added key rollover information

eklein-okta:
@karlmcguinness-okta @cwu-okta @tthompson-okta @mystiberry-okta 

Key rollover information added per discussion with Cindy. Please let me know if you would like any changes. -Evan

### DIFF
--- a/_source/docs/api/resources/oidc.md
+++ b/_source/docs/api/resources/oidc.md
@@ -451,7 +451,16 @@ Each public key is identified by a `kid` attribute, which corresponds with the `
 
 The ID token and the access token are signed by an RSA private key. Okta publishes the corresponding public key and adds a public-key identifier `kid` in the ID token header. To minimize the effects of key rotation, your application should check the `kid`, and if it has changed, check the `jwks_uri` value in the [well-known configuration](#openid-connect-metadata) for a new public key and `kid`. 
 
-All applications must roll over keys for adequate security. Be sure to include key rollover in your implementation.
+All applications must roll over keys for adequate security. Please note the following:
+
+* Okta automatically rotates keys used to sign token, for security purposes.
+* The current key rotation schedule is at least twice a year. This schedule can change without notice.
+* In case of an emergency, Okta can rotate keys as needed.
+* Okta always publishes keys to the JWKS.
+* If your app follows the best practice to always resolve the `kid`, key rotations will not cause problems.
+* If you download the key and store it locally, **you are responsible for updates**.
+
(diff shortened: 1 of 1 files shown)


mystiberry-okta:
Looks great to me!

_source/docs/api/resources/oidc.md
```diff
@@ -449,6 +449,12 @@ Step 3 involves downloading the public JWKS from Okta (specified by the `jwks_ur
 
 Each public key is identified by a `kid` attribute, which corresponds with the `kid` claim in the [ID token header](#claims-in-the-header-section).
 
+The ID token and the access token are signed by an RSA private key. Okta publishes the corresponding public key and adds a public-key identifier `kid` in the ID token header. To minimize the effects of key rotation, your application should check the `kid`, and if it has changed, check the `jwks_uri` value in the [well-known configuration](#openid-connect-metadata) for a new public key and `kid`.
```
mcguinness: I was looking more for a section like https://msdn.microsoft.com/en-us/library/azure/dn641920.aspx#BKMK_Overview

We need to make it clear the following points

- For security purposes, Okta will automatically rotate keys used to sign token.
- Current schedule is atleast 2x year but this is subject to change without notice
- We may rotate keys sooner case of security emergency
- We will always publish keys to the JWKS
- If you follow best practice outlined (resolving kid) you will be fine
- If you download the key and store locally, **YOU MUST BE RESPONSIBLE FOR UPDATING**

I would have a callout as well to emphasize the last point such as

> Keys used to sign tokens will automatically rotate and should always be resolved dynamically against the published JWKS.  Your app may break if you hardcode public keys in your applications!

mcguinness: It may make sense for Signing Keys to have its own section as well

_source/docs/api/resources/oidc.md
```diff
@@ -484,7 +490,7 @@ Each public key is identified by a `kid` attribute, which corresponds with the `
 }
 ~~~
 
-For efficiency your application can retrieve these public keys, cache them, and validate the ID token signatures locally. 
+> Okta strongly recommends retrieving and caching public keys and validating the ID token signatures locally.
```
mcguinness: Something seems off here.  I think we are trying to say something like

Okta strongly recommends that keys are retrieved dynamically via the JWKS published in the [discovery document](#opendid-discovery-document).  It is safe to cache or persist downloaded keys for performance but be careful if you pin your application to a signing key as Okta automatically rotates signing keys

 retrieving and caching public keys and validating the ID token signatures locally.

eklein-okta:
The same changes got checked in with PR 252 due to EBKAC issue, so I am closing this PR.

PR #237, opened 2016-05-11
add refresh token instructions and fix typos doc88783

mystiberry-okta:
@tthompson-okta or @karlmcguinness-okta I fixed two typos and added clarifying info about refresh token use. Is this the best place for the new info?

### DIFF
--- a/_source/docs/api/resources/oidc.md
+++ b/_source/docs/api/resources/oidc.md
@@ -213,7 +213,7 @@ Irrespective of the response type, the contents of the response is always one of
 Parameter         | Description                                                                                        | DataType  | 
 ----------------- | -------------------------------------------------------------------------------------------------- | ----------| 
 id_token          | The ID Token JWT contains the details of the authentication event and the claims corresponding to the requested scopes. This is returned if the <em>response_type</em> includes <em>id_token</em> .| String    | 
-access_token      | The access_token that be used to access the userinfo endpoint. This is returned if the <em>response_type</em>  included token. Unlike the ID Token JWT, the access_token structure is Okta internal only and is subject to change.| String  |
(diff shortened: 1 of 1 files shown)


_source/docs/api/resources/oidc.md
```diff
@@ -374,6 +374,13 @@ For authentication with Basic auth, an HTTP header with the following format mus
 Authorization: Basic ${Base64(<client_id>:<client_secret>)} 
 ~~~
 
+#####Refresh Tokens
+
+You can use refresh tokens if you create a mobile or web OIDC app. To use refresh tokens:
+
+* Check the <b>Allow Offline Access</b> checkbox in Okta to enable refresh tokens.
```
tthompson-okta: I'd reword to: `Check the <b>Allow Offline Access</b> checkbox on the client application page in the Okta Administrative UI (this option is only visible for WEB and NATIVE application types).`

_source/docs/api/resources/oidc.md
```diff
@@ -374,6 +374,13 @@ For authentication with Basic auth, an HTTP header with the following format mus
 Authorization: Basic ${Base64(<client_id>:<client_secret>)} 
 ~~~
 
+#####Refresh Tokens
```
tthompson-okta: This should actually go up next to the "Authorization Endpoint" rather than here by the "Token Endpoint". Also, we need to update the description for the `scope` parameter to list "offline_access" as a valid scope.

mystiberry-okta:
I'm just waiting on these until we get the doc restructure merged. @royalchan-okta  if you don't think it'll cause a conflict, I can go ahead and fix these now. What would you advise?

mystiberry-okta:
Because the file was split, conflict resolution is making my head hurt. Closing this PR and opening a new one with Trevor's comments implemented. Since the file split up, probably worth a second look anyway.

PR #240, opened 2016-05-13
API Doc for SysLog2 System Log

mystiberry-okta:
Someone should take this and finish it. It should not be merged in current state!

@guyzehavi-okta @jeffpeeler-okta

### DIFF
--- a/_source/docs/api/resources/system_log.md
+++ b/_source/docs/api/resources/system_log.md
@@ -158,6 +158,7 @@ Describes the target of the action
 | displayName | Display name of the target                                    | String          | TRUE     |         |           |           |            |
 | detail     | A map that goes from a String key to a value                   | Map[String->Object | TRUE  |         |           |           |            |
 |------------+----------------------------------------------------------------+-----------------+----------+---------+-----------+-----------+------------|
+
 ~~~ json
 {
     "id": "00u3gjksoiRGRAZHLSYV",
@@ -261,6 +262,7 @@ Describes authentication data regarding an event
 ### Issuer Object
 
 Describes an issuer in the authentication context
+
 |------------+----------------------------------------------------------------+-----------------+----------+---------+-----------+-----------+------------|
 | Property   | Description                                                    | DataType        | Nullable | Default | MinLength | MaxLength | Validation |
 | ---------- | -------------------------------------------------------------- | --------------- | -------- | ------- | --------- | --------- | ---------- |
@@ -292,7 +294,7 @@ q         | Finds a user that matches firstName, lastName, and email properties 
 until     | Specifies the first date after which results aren't returned                        | Query      | DateTime | FALSE    |
 after     | An opaque identifier used for pagination                                            | Query      | String   | FALSE    |
 
-The `after` cursor should treated as an opaque value and obtained through the next link relation. See [Pagination](/docs/api/getting_started/design_principles.html#pagination)
+The `after` cursor should treated as an opaque value and obtained through the next link relation. See [Pagination](/docs/api/getting_started/design_principles.html#pagination).
 
 ###### Filter
 
@@ -304,9 +306,9 @@ Filter                                       | Description
 `target.id eq ":id"`                         | Events published with a specific target id
 `actor.id eq ":id"`                          | Events published with a specific target id
 
-See [Filtering](/docs/getting_started/design_principles.html#filtering) for more information on expressions
-Filter allows a filter expression per http://developer.okta.com/docs/api/getting_started/design_principles.html#filtering . The 'q' parameter searches
+The 'q' parameter searches
 on all string fields for the given string and returns the results in the order of relevance.
+See [Filtering](/docs/getting_started/design_principles.html#filtering) for more information about expressions.
 
 **Filter Examples**
 Events published for a target user
@@ -346,7 +348,7 @@ http://rain-admin.okta1.com:1802/sage/api/v1/logs?since=2016-05-25T00:00:00+00:0
 ##### Response Example
 {:.api .api-response .api-response-example}
 
-~~~http
+~~~json
 [
   {
     "version": "0",
--- a/_source/docs/api/resources/system_log.md
+++ b/_source/docs/api/resources/system_log.md
@@ -1,0 +1,619 @@
+---
+layout: docs_page
+title: System Log
+---
+
+## Overview
+
+The Okta System Log API provides read access to your organization's system log. This API provides more functionality than the Events API.
+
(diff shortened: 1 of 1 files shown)


jeffpeeler-okta:
We've definitely deleted old categories & renamed ObjectTypes.
@guyzehavi-okta can you answer other TODO questions?  For example:
`TODO: What does the description of "from" mean?`

guyzehavi-okta:
ack @jeffpeeler-okta

eklein-okta:
@guyzehavi-okta Can you go through the TODO questions in this one file based on our discussion? Don't worry about the old/new mapping questions. Thanks!

guyzehavi-okta:
@eklein-okta Here are my answers for the TODOs. Tell me if you want additional clarification:
TODO: Verify if these are the properties of Log. Log is an object, right? Or is it just a new endpoint for Event object?
Ans: The results returned are a list of Log objects. The endpoint is /api/v1/logs .

 TODO: This is from the Design Doc Syslog2 API spec. Should be validated. Do we need to put any caveats on the content linked to in "filter"?
Ans: I would just make sure that you include the link the filter expressions page. See like '	Filter expression for events' in http://developer.okta.com/docs/api/resources/events for an example.

TODO: What does the description of "from" mean?
Ans: "from" was the old term for "after". "after" is an opaque identifier used for pagination. For example, lets say we give you 20 results per page and your query returned 100 results. Then you need some way to specify the "next" page. The from parameter is a string (a token) that allows us to do that. This is exactly the same as "after" in the old events api, see http://developer.okta.com/docs/api/resources/events, "Specifies the pagination cursor for the next page of events".

TODO: Verify examples in table and example section are all correct. Links to action still make sense? 
Ans: There are two minor issues with the examples given. The date range is actually supplied by the args to the api. Specifically, 'since' (start of date range) and 'until' (end of date range).

TODO: this section doesn't explain q, only the section in Users does. We need to expand the getting started section I think to include this.
Ans: The idea is that there are two ways one can perform a search. 'Filter' reduces the result set (for example all actors with id xyz would be actor.id eq 'xyz'). Search (parameter 'q') is more analogous to a search on Google, it returns the most relevant objects first.

TODO: Add examples for the other parameters. Validate that these are all still good. Events published after 06/01/2013
Ans: This has a slight error. since and until are required parameters. The reason is that otherwise we could return too much data (especially for large orgs). The parameters since and until should be used to filter time.

mystiberry-okta:
@guyzehavi-okta Can you make the changes and submit a new PR? The process is for developers to create the API doc and writer to edit. Here's the internal doc (heading "Standard Workflow"): https://oktawiki.atlassian.net/wiki/display/eng/API+Documentation+Lifecycle

guyzehavi-okta:
@mystiberry-okta can you please clarify what you are trying to do in the sections 'Attributes' and 'Action Object' ? Right above that we defined the response object. What is the purpose of those?

mystiberry-okta:
Typically we include both an example and a table defining the object.

guyzehavi-okta:
@mystiberry-okta Understood. I'll update the tables.

_source/docs/api/resources/system_log.md
```diff
@@ -158,6 +158,7 @@ Describes the target of the action
 | displayName | Display name of the target                                    | String          | TRUE     |         |           |           |            |
 | detail     | A map that goes from a String key to a value                   | Map[String->Object | TRUE  |         |           |           |            |
 |------------+----------------------------------------------------------------+-----------------+----------+---------+-----------+-----------+------------|
+
 ~~~ json
```
mystiberry-okta: The characters are showing as red because this isn't valid code. Even if you pull the "json" it still doesn't like pseudo code. Can you replace this with a real example?And if you do, are the valid values captured somewhere else?

_source/docs/api/resources/system_log.md
```diff
@@ -292,7 +294,7 @@ q         | Finds a user that matches firstName, lastName, and email properties
 until     | Specifies the first date after which results aren't returned                        | Query      | DateTime | FALSE    |
 after     | An opaque identifier used for pagination                                            | Query      | String   | FALSE    |
 
-The `after` cursor should treated as an opaque value and obtained through the next link relation. See [Pagination](/docs/api/getting_started/design_principles.html#pagination)
+The `after` cursor should treated as an opaque value and obtained through the next link relation. See [Pagination](/docs/api/getting_started/design_principles.html#pagination).
 
 ###### Filter
```
mystiberry-okta: The previous Filter section and this has unresolved comments. I'm not sure how to resolve. It seems like the description for 'q' actually describes 'filter' (q is a very limited query). But I'm not an expert here.

mystiberry-okta:
@guyzehavi-okta  I fixed tiny nits with the two tables that weren't formatting (turns out kramdown wants a blank line before and after every table), and asked two questions in my own commit for you as I couldn't fix them myself. Nearly there! One more question:

Is this EA or GA (if EA, I can add a blurb. If GA, I'll add it to the TOC on the left).

mystiberry-okta:
@guyzehavi-okta I added some clarifying content to filter/q--my examples may have syntax or naming errors. I tried to pick interesting things that illustrate value of q, but feel free to suggest better examples or correct.

mystiberry-okta:
@guyzehavi-okta also think about where you want this topic to appear in the table of contents. After Events?

guyzehavi-okta:
@mystiberry-okta Hi. I'll look at the latest comments, per your notes above. I'll change the examples/look over them if needed. This is definitely an EA feature, we are nto ready for GA yet. Also, quick question. You said 'also think about where you want this topic to appear in the table of contents' . What topic are you referring to?

_source/docs/api/resources/system_log.md
```diff
@@ -331,6 +332,26 @@ Events published for a given ip address
 
     filter=client.ipAddress eq "184.73.186.14"
 
+##### Query with q
+
+> Important: The query parameter `q` behaves differently than described in  [Filtering](/docs/getting_started/design_principles.html).
+
+The query parameter `q` searches string fields, and supports the same expressions as the `filter` parameter.
```
guyzehavi-okta: Note: I'll update this but this is wrong. q is just a general search on all fields. It does not support any filtering sytax

guyzehavi-okta:
@mystiberry-okta when you have a chance, can you give this another glance? Just so we can finish this up!

mystiberry-okta:
Looks great @guyzehavi-okta . Just let me know when it should be merged & published (ASAP? for a certain release?)

guyzehavi-okta:
@mystiberry-okta will do. I'll check with PM tomorrow and post the details here (or in an email)

PR #243, opened 2016-05-17
Update session_cookie.md

Robingaal:
Old deprecated api calls are being used here. You can't create a session anymore with username / pwd with the session API. You need to create a sessiontoken first using the Authn API which you later exchange for a valid session and cookietoken if needed.


mystiberry-okta:
Great catch! This error occurs multiple times in this file, and I have a bug to remove the old bad instructions as well as add the new instructions. Thank you for the great suggestion!

Robingaal:
@mystiberry-okta This is still not updated in the documentation. On that specific page. When is this going to be fixed?

PR #254, opened 2016-05-26
Create user call expanded to allow group ids to be passed in.

brianzuzga-okta:
The user will be created and immediately added to that group.  This is necessary for User Admins who can only create users in their targeted groups.

Relates-to: OKTA-89867

### DIFF
--- a/_source/docs/api/resources/users.md
+++ b/_source/docs/api/resources/users.md
@@ -324,6 +324,7 @@ Creates a new user in your Okta organization with or without credentials.
 - [Create User with Password](#create-user-with-password)
 - [Create User with Password & Recovery Question](#create-user-with-password--recovery-question)
 - [Create User with Authentication Provider](#create-user-with-authentication-provider)
+- [Create User in Group](#create-user-in-group)
 
 ##### Request Parameters
 {:.api .api-request .api-request-params}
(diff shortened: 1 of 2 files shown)


brianzuzga-okta:
@mystiberry-okta  @eklein-okta  @karlmcguinness-okta  @jpf @slee-okta API doc PR. The process lists says to list you all as reviewers.

@tthompson-okta @mohitchugh-okta You two were at the API design review.

@georgekwon-okta The process said that you need to give final signoff on this.

brianzuzga-okta:
Postman collection addition coming shortly.

_source/docs/api/resources/users.md
```diff
@@ -693,6 +695,68 @@ curl -v -X POST \
 }
 ~~~
 
+#### Create User in Group
+{:.api .api-operation}
+
+Creates a user that is immediately added the specified groups upon creation.  This supplemental function can be used in conjunction with the other create operations.  This function is necessary for a User Admin that is scoped to only create users in specified groups.
```
mystiberry-okta: Just need  a "to" in line 701 "...that is immediately added **to** the specified groups...
Could be my ignorance, but not sure what 2nd sentence means. Are we saying "if your admin can't create a user in a desired group, use this API function to create users?"

brianzuzga-okta: The intent of the second sentence is to say that you can add groupIds on a simple create operation, as illustrated, or any of the other create operations, like Create User with Password and Recovery Question.

As for the last point, this change really just brings the UI and API into alignment around this for the first time.  Back in December, a User Admin with group scope could create users in their allowed groups only while using our admin UI.  Stuff broke.  While fixing it, we are allowing the same operation to work in the API for the first time.

PR #260, opened 2016-06-06
Update error_codes.md

Robingaal:
Added missing error code E0000080 which you will get when trying to change your password but it's not complex enough according to the password policy complexity rules.


Robingaal:
e0000080 doesnt seemed to be used anymore by the change pwd api.

PR #261, opened 2016-06-06
Update authn.md

Robingaal:
Faulty information given in this file. Error code E0000080 does not exist! Instead of that, the change password api returns a e0000014 error when updating the password failes if its not strong enough according to the security policy rules configured.

### DIFF
--- a/_source/docs/api/resources/authn.md
+++ b/_source/docs/api/resources/authn.md
@@ -1240,9 +1240,9 @@ HTTP/1.1 403 Forbidden
 Content-Type: application/json
 
 {
-  "errorCode": "E0000080",
-  "errorSummary": "The password does meet the complexity requirements of the current password policy.",
-  "errorLink": "E0000080",
+  "errorCode": "E0000014",
+  "errorSummary": "Update of credentials failed",
+  "errorLink": "E0000014",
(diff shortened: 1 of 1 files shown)


Robingaal:
@mystiberry-okta Can you look at this? The devvers from India which are building our solution are now programming the wrong stuff since the documentation is not up to date.

mystiberry-okta:
I'm sorry you encountered this error! Am just verifying now with the team.

Robingaal:
Thanks, please update it asap since devvers are now complaining to me the documentation I've delivered to them is not correct.

mystiberry-okta:
Robin, we couldn't replicate your issue. We get the errors as described in the doc.

Robingaal:
I'm having the issue in 2 oktapreview environments so it's definitely not limited to 1 environment. It seems the occur only in the user-api when doing a **change-password** over there: http://developer.okta.com/docs/api/resources/users.html#lifecycle-operations, but that part of the documentation is not discussing any error codes that are returned on failures. Which makes developer assume it will return the same error codes (including myself..) as the **change password** from the authentication APi.

mystiberry-okta:
http://developer.okta.com/docs/api/resources/authn.html has already been updated with these corrections.

PR #266, opened 2016-06-09
Note that URLs must be encoded

jeffpeeler-okta:
@mystiberry-okta @johnfung-okta  copied a bulletpoint about URL encoding from user API doc.  Based on Michael Rucker's feedback.

### DIFF
--- a/_source/docs/api/resources/system_log.md
+++ b/_source/docs/api/resources/system_log.md
@@ -284,9 +284,7 @@ Describes an issuer in the authentication context
 
 Fetch a list of events from your Okta organization system log.
 
-This operation:
-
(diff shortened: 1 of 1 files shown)


mystiberry-okta:
@jeffpeeler-okta and @johnfung-okta if my minor format change is acceptable, I can merge this asap.

johnfung-okta:
@mystiberry-okta The bulletpoint description is good. But I think it would be even better to put it inside the "Request Example".

screen shot 2016-06-09 at 2 32 01 pm

mystiberry-okta:
@guyzehavi-okta  or @jeffpeeler-okta someone left a note in the doc at this spot (it's not in John's screenshot but it is in the file now). Can someone fix that, and move the description as John suggests as well?

mystiberry-okta:
found a draft comment as well, so pushing this out now. @guyzehavi-okta the draft comment was my mistake, just cc'ing you since we added a piece of technical content.

PR #280, opened 2016-06-16
Clarify steps for using refresh token in oauth2 api

frankmontanaro-okta:


frankmontanaro-okta:
I hope I did this correctly! Didn't see any other commits relate it to a JIRA, but just let me know if there is some process for that as well, or if I missed any of the files (I assume I just need the _source from the doc). @mystiberry-okta

mystiberry-okta:
@frankmontanaro-okta perfectamundo! Will merge immediately, it'll be on the website the next time we publish (Monday) unless you think it needs to go out now.

frankmontanaro-okta:
Monday sounds good to me!

PR #282, opened 2016-06-17
replace text that was lost from PR 239

mystiberry-okta:
@larsjohansen-okta here's the recreation of PR 239 (sounds like a country song!). Please review.


larsjohansen-okta:
🚢

PR #286, opened 2016-06-22
add groups claim limit for 88276

mystiberry-okta:
The prod doc also mentions the limit: 
https://support.okta.com/help/articles/Knowledge_Article/Using-OpenID-Connect
(bottom of the topic, step 5).

Because there were three "notes" in a row, I had to turn them into a bullet list. The first two bullet's content shouldn't have changed/is unrelated to groups claim.

@royalchan-okta or @ashishgupta-okta is this okay to ship? Preview RN go out tonight...


ashishgupta-okta:
Can we change the prod doc to also say "Expect that this limit may change in the future." instead of the current text "This limit is likely to change as more information about size limitations in ID tokens is discovered." ?

mystiberry-okta:
Great catch, will change the prod doc!

PR #272, opened 2016-06-12
overview for OIDC flows first draft

mystiberry-okta:
@federations-okta this is a very drafty first draft of overview content for OIDC, meant to explain OIDC flows and, I think, map to our definition of different types of apps.

It's not ready for a line-by-line review (and the images are rough, not the final images), but can you look at it, and at-mention me if you'd like to participate in a one-hour review meeting so I can turn this into a good final draft. Will schedule for Wed. or Tuesday if RN work goes well.

comments like "this is way off target" or "close but you've got errors" or "don't start with the spec-defined flows, start with our app types" are most helpful at this stage.

PR #296, opened 2016-06-28
Access Token for OIDC

yuliu-okta:
Reviewer: @mystiberry-okta @federations-okta

### DIFF
--- a/_source/docs/api/resources/oidc.md
+++ b/_source/docs/api/resources/oidc.md
@@ -238,7 +238,7 @@ Step 3 involves downloading the public JWKS from Okta (specified by the `jwks_ur
 
 Each public key is identified by a `kid` attribute, which corresponds with the `kid` claim in the [ID Token header](#claims-in-the-header-section).
 
-The ID Token are signed by an RSA private key. Okta publishes the corresponding public key and adds a public-key identifier `kid` in the ID Token header. To minimize the effects of key rotation, your application should check the `kid`, and if it has changed, check the `jwks_uri` value in the [well-known configuration](#openid-connect-discovery-document) for a new public key and `kid`.
+The ID Token is signed by an RSA private key. Okta publishes the corresponding public key and adds a public-key identifier `kid` in the ID Token header. To minimize the effects of key rotation, your application should check the `kid`, and if it has changed, check the `jwks_uri` value in the [well-known configuration](#openid-connect-discovery-document) for a new public key and `kid`.
 
 All apps must roll over keys for adequate security. Please note the following:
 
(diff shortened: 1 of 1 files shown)


mystiberry-okta:
I did a small copy edit. It looks ready to ship to me!

_source/docs/api/resources/oidc.md
```diff
@@ -148,6 +148,8 @@ For more information about configuring an app for OpenID Connect, including grou
 
 Both the Access Token and the ID Token are acquired via [OAuth 2.0](oauth2.html) endpoints.
 
+>The Access Token for OIDC is for Okta use only. The application doesn't need to validate it since the public signing key might not be included in JWKS. The format of the Access Token and the key used to sign it are subject to change without prior notice.
```
tthompson-okta: Could use some re-wording here. The token is not "for Okta use", it's for the application to use when making an API call against Okta. We should instead say that the OpenID Connect access token is applicable only for the Okta `/oauth2/v1/userinfo` endpoint and thus should be treated as opaque by the application.

> The application doesn't need to validate it since the public signing key might not be included in the JWKS..

That's not the reason why the application doesn't need to validate. It should say something like `The access token should not be used against your own resource server.` I don't think we should specifically call out anything about JWK validation; instead we should just tell the developer not to even try to validate the token and if they do, they'll realize it doesn't work (since we don't publish the signing key in the JWK, etc)

_source/docs/api/resources/oidc.md
```diff
@@ -236,7 +238,7 @@ Step 3 involves downloading the public JWKS from Okta (specified by the `jwks_ur
 
 Each public key is identified by a `kid` attribute, which corresponds with the `kid` claim in the [ID Token header](#claims-in-the-header-section).
 
-The ID Token and the access token are signed by an RSA private key. Okta publishes the corresponding public key and adds a public-key identifier `kid` in the ID Token header. To minimize the effects of key rotation, your application should check the `kid`, and if it has changed, check the `jwks_uri` value in the [well-known configuration](#openid-connect-discovery-document) for a new public key and `kid`. 
+The ID Token is signed by an RSA private key. Okta publishes the corresponding public key and adds a public-key identifier `kid` in the ID Token header. To minimize the effects of key rotation, your application should check the `kid`, and if it has changed, check the `jwks_uri` value in the [well-known configuration](#openid-connect-discovery-document) for a new public key and `kid`.
```
tthompson-okta: @mystiberry-okta is hopefully going to fix this paragraph per https://github.com/okta/okta.github.io/pull/272#discussion_r68814473

Your change in this PR (removing the `access token` from the language) is fine, but I really dislike the stuff about `minimizing the effects of key rotation`.

_source/docs/api/resources/oidc.md
yuliu-okta: fixed

tthompson-okta:
🚀

PR #298, opened 2016-06-28
Switched typekit to async mode

keithcasey-okta:
Typekit was doing blocking on load, this makes it async so other things can load. Initial tests show a 20-25% speed up in load times.

Ref:  http://blog.typekit.com/2015/08/04/new-embed-code-for-asynchronous-font-loading/


mystiberry-okta:
@karlmcguinness-okta can you give us your blessing? Keith is new but demo'd the change for JF and me.

karlmcguinness-okta:
Its fine. I think what will happen is the font may pop in when it eventually loads.

keithcasey-okta:
Yes, there may be the "un-themed flash" of text for a few things on page.

karlmcguinness-okta:
🚢

PR #303, opened 2016-07-05
error code fix for password complexity

mystiberry-okta:
@edjohns-okta can you approve these changes for OKTA-94177 (was OKTA-92800).
Do you think I should add to the text for the error code itself? Currently it's this:

"E0000014	Update of credentials failed.	403"

Thanks!

### DIFF
--- a/_source/docs/api/resources/authn.md
+++ b/_source/docs/api/resources/authn.md
@@ -1240,9 +1240,9 @@ HTTP/1.1 403 Forbidden
 Content-Type: application/json
 
 {
-  "errorCode": "E0000080",
+  "errorCode": "E0000014",
   "errorSummary": "The password does meet the complexity requirements of the current password policy.",
-  "errorLink": "E0000080",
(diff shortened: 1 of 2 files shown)


_source/docs/api/resources/authn.md
```diff
@@ -1240,9 +1240,9 @@ HTTP/1.1 403 Forbidden
 Content-Type: application/json
```
edjohns-okta: I found the html usage by looking at the web site.  Where are these .md files used?

edjohns-okta:
I don't think you need to modify the text for the error code, since it looks like we rely on the errorSummary for the details about the specific failure.
As for the rest: LGTM (I had one minor question)
🚀

_source/docs/api/resources/authn.md
mystiberry-okta: The .md (for Markdown) are the source files. We generate the HTML from them and publish that to the website. each .html file matches its same-named .md file.

PR #308, opened 2016-07-07
Added new liquid tags for api lifecycle, operations, and cors

karlmcguinness-okta:


mystiberry-okta:
@karlmcguinness-okta @cwu-okta I love these--but we should only use one scheme to mark Beta in the docs. Are we going with Royal's plan or will we use Beta tag? I'm happy to have it, just don't want us marking Beta in different ways (which, I now realize, means scrubbing the existing doc no matter which we choose).

cwu-okta:
As we do not yet have a way to define beta on the developer site, and our definition of beta can also apply to attributes the short term solution is to use the beta query string param. I think it makes sense to explore using this new mechanism to see if it suits our needs for beta labels.

mystiberry-okta:
@cwu-okta just double-checking, okay to merge?

cwu-okta:
Ok from me if all things check out to address the beta query param.

_source/docs/api/resources/users.md
```diff
@@ -1527,24 +1527,24 @@ curl -v -X GET \
 #### List Users with Search
 {:.api .api-operation}
 
-> Listing users with search is an Early Access feature which you can request from Okta Support.
-It's been tested to the same level as Generally Available (GA) features, so you  can use it in a production environment. 
+> {% api_lifecycle ea margin-bottom %}<br>Listing users with search is an Early Access feature which you can request from Okta Support.
+It's been tested to the same level as Generally Available (GA) features, so you can use it in a production environment.
```
tthompson-okta: @mystiberry-okta we need to move the descriptions of BETA vs EA vs GA out of the Users topic and into a top-level topic.

tthompson-okta:
I think we need to move @royalchan-okta's `beta=true` query param into a top-level UI toggle on the site, and then tie it together with this PR to use the liquid tags.

tthompson-okta:
🚀

_source/docs/api/resources/users.md
mystiberry-okta: Okay--why?

tthompson-okta: Otherwise we'll end up duplicating this paragraph (which needs more detail anyways) in every single page eventually. I think there's already a Jira for adding top level description of platform feature lifecycles (EA/GA), right?

mystiberry-okta: Ah! Leave the "this feature is EA" in place, move the long explanation of what it means to a top level topic. Happy to, but @cwu-okta says that's gated on us getting a definition of Beta for dev. I'm happy to consolidate it all.

mystiberry-okta: will do the moving on a separate JIRA, as it's likely to affect more than just this file...

PR #329, opened 2016-07-14
Describe form of allowable custom scopes.

danielzeller-okta:
Link to spec defining acceptable scopes.

Primary reviewer: @yuliu-okta 
@federations-okta 

Related to: OKTA-95050

### DIFF
--- a/_source/docs/api/resources/oauth2.md
+++ b/_source/docs/api/resources/oauth2.md
@@ -129,7 +129,7 @@ The admin can configure custom scopes and claims via the Authorization Server ta
 {:.beta}
 If the request that generates the access token contains any custom scopes, those scopes will be part of the <b>scp</b> claim together with the scopes provided from the [OIDC specification](http://openid.net/specs/openid-connect-core-1_0.html). The form of these custom scopes must conform to the [OAuth2 specification](https://tools.ietf.org/html/rfc6749#section-3.3).
 
->*Note:* Scope names can't contain the characters < (less than) or > (greater than).
+>*Note:* Scope names can contain the characters < (less than) or > (greater than), but not both characters.
 {:.beta}
 
 ##### Custom claims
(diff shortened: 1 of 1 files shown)


danielzeller-okta:
Code pr: https://github.com/okta/okta-core/pull/15333

mystiberry-okta:
@danielzeller-okta One question: I read the spec referenced, found their example hard to understand. Could be it's my ignorance in this area, but wondering if we need to give customers a little more help here?  Also, should we mention they can't use < or >, or we intend to fix this soon.

danielzeller-okta:
@mystiberry-okta the spec is basically just defining a range of acceptable characters in hexadecimal. I don't think we need to provide further guidance in this because if they find their scopes not being valid they are using some _really_ strange characters.

As for point 2 yes thank you for reminding me. It's not that `<` **or** `>` is disallowed, it's `<` **and** `>` that isn't allowed (in any order). I'm not sure how we should best surface that detail. I'd greatly appreciate you owning that change :)

yuliu-okta:
Do we really need to mention a parameter must follow the spec? We do not do that for all others. It seems to me that if we do not want to mention the exception of <>, then the doc change is not needed.

danielzeller-okta:
@yuliu-okta Karl would like us to call out the <> exception

mystiberry-okta:
@danielzeller-okta did I capture the < > restriction correctly?

danielzeller-okta:
@mystiberry-okta almost. Either `<` or `>` are fine, but BOTH are not allowed in any order.

mystiberry-okta:
Okay, @danielzeller-okta , I captured either but not both in the comment. Please review. Thanks!

danielzeller-okta:
LGTM @mystiberry-okta 🍧

mystiberry-okta:
Just waiting to merge until .29 is pushed next week, to go with OKTA-95050. Let me know if it can go sooner.

PR #330, opened 2016-07-15
Update User Links doc to reflect individual vs collection returns

alandry-okta:
@karlmcguinness-okta @jaikumarbathija-okta @tthompson-okta @mystiberry-okta  

Support was requesting documentation for the change to the HAL links on User collections before we start turning it on. Our current phrasing is a little vague about when exactly you get each link, this is clarifying that consumers should only expect a self link for collections of Users

Main change is in the Links Object doc, and then the various example responses are also updated:

screen shot 2016-07-15 at 10 57 29 am

### DIFF
--- a/_source/docs/api/resources/groups.md
+++ b/_source/docs/api/resources/groups.md
@@ -1094,17 +1094,8 @@ Link: <https://your-domain.okta.com/api/v1/groups/00g1fanEFIQHMQQJMHZP/users?aft
       }
     },
     "_links": {
-      "resetPassword": {
-        "href": "https://your-domain.okta.com/api/v1/users/00u1f96ECLNVOKVMUSEA/lifecycle/reset_password"
-      },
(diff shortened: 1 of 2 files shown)


jaikumarbathija-okta:
lgtm.

mystiberry-okta:
@alandry-okta looks great--should I merge this now or is it waiting on a particular JIRA or FF or release?

alandry-okta:
I want to get @karlmcguinness-okta to agree on the wording too and then we can merge. 

Support is looking for this to go in now, before we do an FF rollout since we're changing our responses and they wanted us to clarify the return values before that started.

mcguinness:
seems fine to me 🚢

mystiberry-okta:
Will merge and publish with the .29 refresh tomorrow afternoon/evening. Thanks!

PR #332, opened 2016-07-20
doc-87540-oidc-tweaks

mystiberry-okta:
Fixed typos and removed email_verified

### DIFF
--- a/_source/docs/api/resources/oidc.md
+++ b/_source/docs/api/resources/oidc.md
@@ -128,7 +128,6 @@ The ID Token (*id_token*) consists of three period-separated, base64URL-encoded 
   "locale":"en-US",
   "updated_at":1311280970,
   "email":"john.doe@example.com",
-  "email_verified":true,
   "address" : { "street_address": "123 Hollywood Blvd.", 
   		"locality": "Los Angeles", 
   		"region": "CA", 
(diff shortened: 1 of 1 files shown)


mystiberry-okta:
@tthompson-okta please review these changes. Great catches!

mystiberry-okta:
Poke @tthompson-okta

_source/docs/api/resources/oidc.md
```diff
@@ -128,7 +128,6 @@ The ID Token (*id_token*) consists of three period-separated, base64URL-encoded
   "locale":"en-US",
   "updated_at":1311280970,
   "email":"john.doe@example.com",
-  "email_verified":true,
```
tthompson-okta: Ah, apologies. The Jira is quite old and out of date. We *do* support "email_verified", but it has the limitation that it refers to the user's "primary email". In Okta there is the primary email and the secondary email, but this attribute specifically refers to the primary email as it's implemented right now. I think we should just note that somewhere and then leave it in all the examples, since it does come back if you request the `email` scope.

mystiberry-okta:
@tthompson-okta  Please review--this is due today. Thanks!

tthompson-okta:
🚀

PR #334, opened 2016-07-21
RN 2016-29 updates

mystiberry-okta:
Fixing first heading ("expires_in" is very very small)
Fixing file names so they all follow the same convention. Will delete errant files shortly.


mystiberry-okta:
https://oktainc.atlassian.net/browse/OKTA-95866

PR #338, opened 2016-07-25
adjust some text for response_mode for authorize call

frankmontanaro-okta:


frankmontanaro-okta:
@royalchan-okta let me know if this text makes any more sense in the current state of things
cc: @mystiberry-okta

royalchan-okta:
Response Mode query can't be used with `reponse_type=token` either. Specifically, it can only be used with `code`. Otherwise everything else looks good.

frankmontanaro-okta:
k cool updated 'If <em>id_token</em> **or <em>token</em>** is specified as the response type, then <em>query</em> can't be used as the response mode.'

mystiberry-okta:
We already have some caveats/extra info in the "Parameter Details" section below the table (the content itself looks good!). Can I move the caveats down there? I should probably put a link in the current description.

royalchan-okta:
I'm ok either location, whatever you think is best!

PR #341, opened 2016-07-26
/users/me/appLinks and /users/me/groups update

raphaellondner-okta:
Explicit mention of /users/me/appLinks and /users/me/groups endpoints

### DIFF
--- a/_source/docs/api/resources/users.md
+++ b/_source/docs/api/resources/users.md
@@ -1993,6 +1993,8 @@ curl -v -X POST \
 
 Fetches appLinks for all direct or indirect (via group membership) assigned applications.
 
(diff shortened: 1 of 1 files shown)


mystiberry-okta:
@raphaellondner-okta should someone give this a technical review, or okay to merge?

karlmcguinness-okta:
Looks right technically but not from style perspective.  We should use CORS badge to indicate CORS and call out separate examples for each flow (with `id` and with `me`).  Can we harmonize the style of the update?

mystiberry-okta:
@raphaellondner-okta and @karlmcguinness-okta I can do that, no problem.

PR #352, opened 2016-08-02
Add beta OIDC Playground tab

royalchan-okta:
OKTA-96681

Doesn't behave super nicely since most links remove the beta=true query param but you can see the tab if you are in beta mode:

image
image

@mystiberry-okta 
@federations-okta

### DIFF
--- a/_source/oidc_playground/index.md
+++ b/_source/oidc_playground/index.md
@@ -1,0 +1,7 @@
+---
+layout: docs_page
+---
+
+<html>
+<div>Hello World!</div>
+</html>
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 GEM
   remote: https://rubygems.org/
   specs:
-    activesupport (4.2.6)
+    activesupport (4.2.7)
       i18n (~> 0.7)
       json (~> 1.7, >= 1.7.7)
       minitest (~> 5.1)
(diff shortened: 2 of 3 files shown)


_source/oidc_playground/index.md
```diff
@@ -0,0 +1,7 @@
+---
+layout: docs_page
+---
```
abinashbehera-okta: nit: We could add a title to this page, something on the lines of - title: OIDC Playground

_source/oidc_playground/index.md
```diff
@@ -0,0 +1,7 @@
...
+
+<html>
```
abinashbehera-okta: I guess, we can leave the html tags out of this file as these will be generated later. Also, we can add something like - '# Hello World!' or '# OIDC Playground' as a placeholder for now.

_source/oidc_playground/index.md
royalchan-okta: The title comes with the puzzle piece and the edit on github cruft, it will probably look cleaner with out it

_source/oidc_playground/index.md
royalchan-okta: yeah this should just be embedded html

royalchan-okta:
Might want to eventually add some JS to start the side nav minimized as well.

_source/oidc_playground/index.md
mystiberry-okta: Is it a disruptive experience for the user to, say, use the beta query param to see the content in oauth2.html and suddenly have that tab appear?  Or does it stay hidden unless the URL is developer.okta.com?beta="true"  If we're not going to use the left nav, why is this exposed in a tab? Or does the left nav become useful at some point?

royalchan-okta: it will show up regardless of where in the docs the user is. I don't think it's too disruptive, and certainly more disruptive that navigating anywhere non-beta will hide the tab

mystiberry-okta:
@abinashbehera-okta and @royalchan-okta , okay to merge this for Wed. Preview?

mystiberry-okta:
Closing--a new PR will be posted.

PR #357, opened 2016-08-05
Adds information on the groups claim in the access token.

danielzeller-okta:
Primary reviewers: @larsjohansen-okta @mystiberry-okta 

@federations-okta


larsjohansen-okta:
🚢

mystiberry-okta:
Should I push out right away or wait for .32? Thanks!

_source/docs/api/resources/oidc.md
```diff
@@ -209,7 +209,7 @@ Instead, the ID token contains the `name` and `preferred_username` claims if the
 
 * The full set of claims for the requested scopes is available via the [/oauth2/v1/userinfo](#get-user-information) endpoint. Call this endpoint using the Access Token.
 
-* To protect against arbitrarily large numbers of groups matching the group filter, the group claim has a limit of 100. 
+* To protect against arbitrarily large numbers of groups matching the group filter, the groups claim has a limit of 100.
```
raphaellondner-okta: It is my understanding that this limitation also applies to the groups claim in the Access Token, so maybe we should add it in  _source/docs/api/resources/oauth2.md as well?

danielzeller-okta: This is true. I'll add it to the callout as well.

danielzeller-okta:
@mystiberry-okta wait for .32 (cut from today's codefreeze, right?)

PR #362, opened 2016-08-10
remove a bad link from oidc.md

mystiberry-okta:
We missed this link to the content we just removed.


mystiberry-okta:
Resolves: OKTA-98225

PR #368, opened 2016-08-17
okta-97071 doc update

mystiberry-okta:
Clarify that while ID Tokens are valid at introspection endpoint, Accces Tokens and Refresh Tokens are the intended uses.

@raphaellondner-okta  or @tthompson-okta , this section is missing request and response examples. Not sure I'm set up to create them, but if you supply the examples, I'll happily add them. Long term, will get set up properly so I can do these simple tasks.


mystiberry-okta:
@raphaellondner-okta or @tthompson-okta  can you please review? Okay to publish without the request/response examples (I can create a follow-up JIRA for those).

tthompson-okta:
Looks fine, shipit.

PR #378, opened 2016-08-24
remove x5c and x5t from oidc doc

mystiberry-okta:
@raphaellondner-okta , I wasn't sure if I needed to add any text about not knowing which key is used, let me know if you think it still needs to be added.

This should also be reviewed by someone on @federations-okta (Royal? Yu?) to ensure I haven't introduced any errors.

### DIFF
--- a/_source/docs/api/resources/oidc.md
+++ b/_source/docs/api/resources/oidc.md
@@ -352,6 +352,8 @@ It is safe to cache or persist downloaded keys for performance.
 However, if the client application is pinned to a signing key, the verification might fail since Okta rotates the key automatically. 
 Pinned client applications must periodically check the Okta signing keys.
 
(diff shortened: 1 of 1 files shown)


raphaellondner-okta:
I'm fine with that PR. I think it would make sense to mention that "any of the 2 or 3 keys listed in that endpoint might be used to sign tokens at any given point in time" (avoids alternate but more accurate language that would state "there is no way to know which of the 2 or 3 keys listed in that endpoint is used to sign tokens other than by inspecting the kid of a current, valid token". If we don't mention this one way or another, I'm pretty sure this will end up as questions in the support queue (which will get passed over to us for lack of documentation...)

mystiberry-okta:
Added the suggested text.

PR #385, opened 2016-08-30
Oktane16 update

joelfranusic-okta:


joelfranusic-okta:
... will merge at 6am PT

PR #392, opened 2016-09-02
OKTA-99877

mystiberry-okta:
@yuliu-okta , can you review this? I was wondering if we need to include any of the following from the spec, as well (we try to avoid just copying the spec, but this info seemed important (access token vs. ID Token):

_________________________
	The Claims requested by the profile, email, address, and phone scope values are returned from the UserInfo Endpoint, as described inSection 5.3.2, when a response_type value is used that results in an Access Token being issued. However, when no Access Token is issued (which is the case for the response_type value id_token), the resulting Claims are returned in the ID Token.
	
	In some cases, the End-User will be given the option to have the OpenID Provider decline to provide some or all information requested by RPs. To minimize the amount of information that the End-User is being asked to disclose, an RP can elect to only request a subset of the information available from the UserInfo Endpoint.
	
	The following is a non-normative example of an unencoded scope request:
	
	  scope=openid profile email phone 
	
_____________________________	
Also, do we need one sample or example? To show where they specify the scopes in a request?

### DIFF
--- a/_source/_docs/api/resources/oidc.md
+++ b/_source/_docs/api/resources/oidc.md
@@ -42,10 +42,36 @@ The properties you need depend on which client profile and use case you are targ
 
 ## Claims
 
-ID tokens issued by Okta contain claims, which are statements about a subject (user).
+Tokens issued by Okta contain claims, which are statements about a subject (user).
 For example, the claim can be about a name, identity, key, group, or privilege.
-The claims in a security token are dependent upon the type of token, the type of credential used to authenticate the user, 
-and the application configuration.
+The claims in a security token are dependent upon the type of token, the type of credential used to authenticate the user, and the application configuration.
+
+The claims requested by the `profile`, `email`, `address`, and `phone` scope values are returned from the UserInfo Endpoint, as described in [the OpenID spec Section 5.3.2](http://openid.net/specs/openid-connect-core-1_0.html#UserInfoResponse), when a `response_type` value is used that results in an Access Token being issued. However, when no Access Token is issued (which is the case for the `response_type` value `id_token`), the resulting Claims are returned in the ID Token.
+
+## Scopes
+
+OpenID Connect uses scope values to specify what access privileges are being requested for Access Tokens. 
+The scopes associated with Access Tokens determine which claims are available when they are used 
+to access [the OIDC `userinfo` endpoint](http://developer.okta.com/docs/api/resources/oidc.html#get-user-information). The following scopes are supported:
+
+|--------------+--------------------------------------------------------------------------------|--------------|
+| Property     | Description                                                                    | Required     |
+|--------------+---------+----------------------------------------------------------------------|--------------|
+| openid       | Identifies the request as an OpenID Connect request.                           | Yes          |
+| profile      | Requests access to the end-user's default profile claims.                      | No           |
+| email        | Requests access to the `email` and `email_verified` claims.                    | No           |
+| phone        | Requests access to the `phone_number` and `phone_number_verified` claims.      | No           |
+| address      | Requests access to the `address` claim.                                        | No           |
+| groups       | Requests access to the `groups` claim.                                         | No           |
+| offline_access | Requests a refresh token, used to obtain more access tokens without re-prompting the user for authentication. | No           |
+
+### Scope Values
+
+* `openid` is required for any OpenID request connect flow. If no openid scope value is present, the request may
+  be a valid OAuth 2.0 request, but it's not an OpenID Connect request.
+* `profile` requests access to these default profile claims: `name`, `family_name`, `given_name`, `middle_name`, `nickname`, `preferred_username`, `profile`,  
+`picture`, `website`, `gender`, `birthdate`, `zoneinfo`,`locale`, and `updated_at`.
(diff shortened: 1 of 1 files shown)


_source/_docs/api/resources/oidc.md
```diff
@@ -40,6 +40,31 @@ you must have in your application to ensure it’s always updated with the lates
 The flow of requests and responses for the authentication process is based on OAuth 2.0 and OpenID Connect protocols.
 The properties you need depend on which client profile and use case you are targeting.
 
+## Scopes
+
+OpenID Connect uses scope values to specify what access privileges are being requested for Access Tokens. 
+The scopes associated with Access Tokens determine which resources are available when they are used
```
yuliu-okta: This sentence is generally talking about OAuth2. I'd rather saying "The scopes associated with Access Tokens determine which claims are available when they are used to access OIDC userinfo endpoint".

_source/_docs/api/resources/oidc.md
```diff
@@ -40,6 +40,31 @@ you must have in your application to ensure it’s always updated with the lates
...
+
+OpenID Connect uses scope values to specify what access privileges are being requested for Access Tokens. 
+The scopes associated with Access Tokens determine which resources are available when they are used 
+to access OAuth 2.0 protected endpoints. The following default scopes are supported:
+
+|--------------+--------------------------------------------------------------------------------|--------------|
+| Property     | Description                                                                    | Required     |
+|--------------+---------+----------------------------------------------------------------------|--------------|
+| openid       | Identifies the request as an OpenID Connect request.                           | Yes          |
+| profile      | Requests access to the end-user's default profile claims.                      | No           |
```
yuliu-okta: It is a little wired in the Claim section, it says "ID tokens issued by Okta contain claims, which are statements about a subject (user)", it looks like only ID tokens have claims.

_source/_docs/api/resources/oidc.md
```diff
@@ -40,6 +40,31 @@ you must have in your application to ensure it’s always updated with the lates
 The flow of requests and responses for the authentication process is based on OAuth 2.0 and OpenID Connect protocols.
 The properties you need depend on which client profile and use case you are targeting.
 
+## Scopes
+
+OpenID Connect uses scope values to specify what access privileges are being requested for Access Tokens. 
+The scopes associated with Access Tokens determine which resources are available when they are used 
+to access OAuth 2.0 protected endpoints. The following default scopes are supported:
+
+|--------------+--------------------------------------------------------------------------------|--------------|
```
yuliu-okta: This table and the section below seems having some duplication with http://developer.okta.com/docs/api/resources/oidc.html#scope-dependent-claims-not-always-returned.

_source/_docs/api/resources/oidc.md
```diff
@@ -40,6 +40,31 @@ you must have in your application to ensure it’s always updated with the lates
 The flow of requests and responses for the authentication process is based on OAuth 2.0 and OpenID Connect protocols.
 The properties you need depend on which client profile and use case you are targeting.
 
+## Scopes
```
yuliu-okta: Since this section is talking about which claim will return for which scopes, it might be better to put after claim section.

yuliu-okta:
Yes, I think it is necessary to include "The Claims requested by the profile, email, address, and phone scope values are returned from the UserInfo Endpoint, as described inSection 5.3.2, when a response_type value is used that results in an Access Token being issued. However, when no Access Token is issued (which is the case for the response_type value id_token), the resulting Claims are returned in the ID Token."
But again it is kind of duplication with "The client can also optionally request an Access Token along with the ID Token. In this case, in order to keep the size of the ID Token small, the ID Token body does not contain all the scope dependent claims."

_source/_docs/api/resources/oidc.md
mystiberry-okta: Great! We mean this endpoint, right? http://developer.okta.com/docs/api/resources/oidc.html#get-user-information

_source/_docs/api/resources/oidc.md
```diff
@@ -40,6 +40,31 @@ you must have in your application to ensure it’s always updated with the lates
...
+
+OpenID Connect uses scope values to specify what access privileges are being requested for Access Tokens. 
+The scopes associated with Access Tokens determine which resources are available when they are used 
+to access OAuth 2.0 protected endpoints. The following default scopes are supported:
```
mystiberry-okta: Fixed.

_source/_docs/api/resources/oidc.md
mystiberry-okta: @yuliu-okta changed "ID tokens" to "Tokens".

_source/_docs/api/resources/oidc.md
yuliu-okta: yes

mystiberry-okta: done.

_source/_docs/api/resources/oidc.md
mystiberry-okta: So, they are similar in that the scope properties tend to reference the same-named claims, but it's my understanding that they are different properties, because they are scopes, not claims.

_source/_docs/api/resources/oidc.md
mystiberry-okta: Done.

PR #405, opened 2016-09-16
changes for Key sharing

ashishgupta-okta:
changes for Key sharing

Primary Reviewer: @mystiberry-okta  @yuliu-okta 
Reviewers : @federations-okta

### DIFF
--- a/_source/_docs/api/resources/apps.md
+++ b/_source/_docs/api/resources/apps.md
@@ -290,7 +290,7 @@ Specifies credentials and scheme for the application's `signOnMode`.
       "type": "BUILT_IN"
     },
     "signing": {
-      "kid": "akm5hvbn1vojA9Fsa0h7"
+      "kid": "SIMcCQNY3uwXoW3y0vf6VxiBb5n9pf8L2fK8d-FIbm4"
     },
     "userName": "test",
     "password": {}
@@ -343,7 +343,7 @@ Determines the [key](#application-key-credential-model) used for signing asserti
 |------------+---------------------------------------------------------------------------------------------+----------------------------------+----------+-------------------+-----------+-----------+------------|
 | Property   | Description                                                                                 | DataType                         | Nullable | Default           | MinLength | MaxLength | Validation |
 | ---------- | ------------------------------------------------------------------------------------------- | -------------------------------- | -------- | ----------------- | --------- | ----------| ---------- |
-| kid        | Reference for key credential in the [app's key store](#application-key-store-operations)    | String                           | FALSE    |                   |           |           |            |
+| kid        | Reference for [key credential for the app](#application-key-store-operations)    | String                           | FALSE    |                   |           |           |            |
 |------------+---------------------------------------------------------------------------------------------+----------------------------------+----------+-------------------+-----------+-----------+------------|
 
 > You must to enable the key rollover feature to view `kid`. Key rollover is an Early Access feature; contact Customer Support to enable it.
@@ -351,7 +351,7 @@ Determines the [key](#application-key-credential-model) used for signing asserti
 ~~~json
 {
   "signing": {
-    "kid": "akm5hvbn1vojA9Fsa0h7"
+    "kid": "SIMcCQNY3uwXoW3y0vf6VxiBb5n9pf8L2fK8d-FIbm4"
   }
 }
 ~~~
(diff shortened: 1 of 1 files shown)


ashishgupta-okta:
image

ashishgupta-okta:
image

ashishgupta-okta:
image

_source/_docs/api/resources/apps.md
```diff
@@ -3029,7 +3029,7 @@ curl -v -X PUT \
       "type": "BUILT_IN"
     },
     "signing": {
-      "kid": "akmip6HgvTMB1TcKU0g3"
+      "kid": "SIMcCQNY3uwXoW3y0vf6VxiBb5n9pf8L2fK8d-FIbm4"
     }
   }
 }' "https://${org}.okta.com/api/v1/apps/0oainmLkOL329Jcju0g3"
```
mystiberry-okta: Why isn't the value here changed to the longer value?

_source/_docs/api/resources/apps.md
```diff
@@ -4103,6 +4103,81 @@ Content-Type: application/json
...
+
+##### Request Parameters
+{:.api .api-request .api-request-params}
+
+Parameter     | Description                                                                     | Param Type | DataType                                      | Required | Default
+------------- | ------------------------------------------------------------------------------- | ---------- | --------------------------------------------- | -------- | -------
+sourceAid           | unique key of source [Application](#application-model)                                 | URL        | String                                        | TRUE     |
+kid           | unique key of [Application Key Credential](#application-key-credential-model)   | URL        | String                                        | TRUE     |                                      |      |
+targetAppInstanceId |  unique key of target [Application](#application-model)   | Query      | String                                        | TRUE     |
+
```
mystiberry-okta: Where does the developer find the values for sourceAid, targetAppInstanceId? that is, should these be added to the response parameters/examples in http://developer.okta.com/docs/api/resources/apps.html#example or http://developer.okta.com/docs/api/resources/apps.html#application-properties ?

mystiberry-okta: Also please start the Description text with a capital letter: "Unique key of ...."

_source/_docs/api/resources/apps.md
ashishgupta-okta: this is the appInstanceId and not the keyId

_source/_docs/api/resources/apps.md
ashishgupta-okta: `sourceAid` is the 'unique key of the source application' and the `targetAppInstanceId` is the 'unique key of the target application'.

Changed `unique...` to `Unique...`

_source/_docs/api/resources/apps.md
```diff
@@ -4103,6 +4103,81 @@ Content-Type: application/json
 }
 ~~~
 
+### Clone Application Key Credential
+{:.api .api-operation}
+
+<span class="api-uri-template api-uri-post"><span class="api-label">POST</span> /apps/*:sourceAid*/credentials/keys/*:kid*/clone?targetAppInstanceId=*:targetAppInstanceId*</span>
```
cwu-okta: we should probably follow convention here and rename to 
- rename sourceAid => aid
- targetAppInstanceId =>targetAid

Remember we do not distinguish between app instance and app in developer.okta.com. We take great pains to not hide this abstraction. aid stands for app id, I believe.

_source/_docs/api/resources/apps.md
```diff
@@ -4103,6 +4103,81 @@ Content-Type: application/json
...
+
+Clones a X.509 certificate for an application key credential from a source application to target application.
```
cwu-okta: Would it be more clear to say ```...from a source application to another (target) application```?
@mystiberry-okta @karlmcguinness-okta  thoughts?

_source/_docs/api/resources/apps.md
```diff
@@ -4103,6 +4103,81 @@ Content-Type: application/json
...
+
+> To update the target application with the cloned key credential, see [Update Key Credential](#update-key-credential-for-application)
+
+##### Request Parameters
+{:.api .api-request .api-request-params}
+
+Parameter     | Description                                                                     | Param Type | DataType                                      | Required | Default
+------------- | ------------------------------------------------------------------------------- | ---------- | --------------------------------------------- | -------- | -------
+sourceAid           | Unique key of source [Application](#application-model)                                 | URL        | String                                        | TRUE     |
```
cwu-okta: Same comment as above about renaming var names.

cwu-okta: Should this be "unique key of app" ? Under [application properties](http://developer.okta.com/docs/api/resources/apps.html#application-properties) that is the description used for id.

_source/_docs/api/resources/apps.md
ashishgupta-okta: gotcha. Following the convention, should we instead use `sourceAid` and `targetAid`, instead of `aid` and `targetAid` ? Thoughts @yuliu-okta ?

cwu-okta: I prefer to leave it ```aid``` instead of ```sourceAid```. Source is kind of implicit here. In other parts of this page we always refer to it as ```aid```, which means it is the app [instance] of interest. I don't see a need to distinguish it any differently for this new endpoint.

_source/_docs/api/resources/apps.md
```diff
@@ -4103,6 +4103,81 @@ Content-Type: application/json
...
+  ],
+  "kid": "SIMcCQNY3uwXoW3y0vf6VxiBb5n9pf8L2fK8d-FIbm4",
+  "kty": "RSA",
+  "use": "sig",
+  "x5t#S256": "5GOpy9CQVtfvBmu2T8BHvpKE4OGtC3BuS046t7p9pps"
+}
+~~~
+
+If key is already cloned for target application, you will receive an error response.
```
ashishgupta-okta: This is for the case when there already exists a mapping for the target app(Instance), or, to elaborate, if key already exists in target app Instance key store. Target app is set to use the cloned key by the Apps API (http://developer.okta.com/docs/api/resources/apps.html#update-key-credential-for-application). So seems wrong to say "If target app is already set to the key, you will receive a 400 error reponse"

cwu-okta: So this use case is for if the target app already has that key in its list of key credentials, regardless of whether the target app is set to use that specific key or not, correct?

If the above understanding is correct, it might be better to say 
```If target app already is associated to the key credential, you will receive a 400 error response```

File: _source/_docs/api/resources/apps.md (hunk 1 above)
Comment (mystiberry-okta) [inline _source/_docs/api/resources/apps.md]: Ack.

File: _source/_docs/api/resources/apps.md (hunk 6 above)
Comment (ashishgupta-okta) [inline _source/_docs/api/resources/apps.md]: Yes, thats correct.

File: _source/_docs/api/resources/apps.md (hunk 4 above)
Comment (mystiberry-okta) [inline _source/_docs/api/resources/apps.md]: Not sure what "another" means here. It should be "a target application."  Source app just means "the app that has the cert" and target just means "the app you want to give the cert to", yes?

Comment (ashishgupta-okta) [inline _source/_docs/api/resources/apps.md]: Yes, thats what it looks like to a consumer using the API.

**Comment by ashishgupta-okta** [comment]
Addressed all changes @cwu-okta .

File: _source/_docs/api/resources/apps.md (hunk 5 above)
Comment (ashishgupta-okta) [inline _source/_docs/api/resources/apps.md]: Changed this to Unique key of [Application] to be consistent with the whole doc.

File: _source/_docs/api/resources/apps.md [hunk 7]
```diff
@@ -4103,6 +4103,81 @@ Content-Type: application/json
...
+  ],
+  "kid": "SIMcCQNY3uwXoW3y0vf6VxiBb5n9pf8L2fK8d-FIbm4",
+  "kty": "RSA",
+  "use": "sig",
+  "x5t#S256": "5GOpy9CQVtfvBmu2T8BHvpKE4OGtC3BuS046t7p9pps"
+}
+~~~
+
+If key is already present in the list of key credentials for target application, you will receive a 400 error response.
```
cwu-okta: Nit: should read ```...for the target application...```

_source/_docs/api/resources/apps.md
```diff
@@ -4103,6 +4103,81 @@ Content-Type: application/json
...
+Content-Type: application/json
+
+{
+  "errorCode": "E0000001",
+  "errorSummary": "Api validation failed: cloneKey",
+  "errorLink": "E0000001",
+  "errorId": "oaeQACJOHl1TKSGj8jA3hEpAg",
+  "errorCauses": [
+    {
+      "errorSummary": "Key is already cloned for target app."
```
cwu-okta: Should this error message be different? I realize that is in the code PR but wondering if it should be a more generic message. Perhaps a better message would be ```Key already exists in the target app's list of key credentials```

ashishgupta-okta: fixed

cwu-okta: Actually isn't this no longer relevant? I thought the last discussion for this is even if you are cloning a key to the target app (and target app already contains it in the list of keys) it would return in a success to the API (and no opt in the server)

ashishgupta-okta: umm, as I remember we decided to throw the error (key already exists) if the key already exists in the list. Its a POST call so we should fail. What we discussed yesterday was around default keys not being allowed for cloning. Earlier we were not allowing customer to pass default keys but now you can.

cwu-okta: ok makes sense

_source/_docs/api/resources/apps.md
ashishgupta-okta: @mystiberry-okta  I can change that but wont that be inconsistent with rest of the api doc?

PR #411, opened 2016-09-23
Add links to .NET and Python SDKs for the Okta API

joelfranusic-okta:
@raphaellondner-okta could you review the updates before we go live?

Here are what the changes would look like if they were to go live:

_net___okta_developer

python___okta_developer


mystiberry-okta:
@joelfranusic-okta  We don't usually call out the docs in the platform RN, but if you want to announce these new SDK docs, I'm happy to. Up to you, just let me know.

joelfranusic-okta:
Well ... these were available on the old website, but not migrated to the new website. We should announce the next one, which should be new.

joelfranusic-okta:
@mystiberry-okta any other concerns?

PR #412, opened 2016-09-26
changes for list Apps API : Support for kid filter

ashishgupta-okta:
i18n changes for list Apps API : Support for kid filter
updated postman collection

Primary Reviewer : @mystiberry-okta @royalchan-okta 
Reviewer : @federations-okta


_source/_docs/api/resources/apps.md
```diff
@@ -1862,6 +1863,7 @@ Filter                 | Description
 `status eq "INACTIVE"` | Apps that have a `status` of `INACTIVE`
 `user.id eq ":uid"`    | Apps assigned to a specific user such as `00ucw2RPGIUNTDQOYPOF`
 `group.id eq ":gid"`   | Apps assigned to a specific group such as `00gckgEHZXOUDGDJLYLG`
+`credentials.signing.kid eq ":kid"`   | Apps using a particular key such as `SIMcCQNY3uwXoW3y0vf6VxiBb5n9pf8L2fK8d-FIbm4`
```
cwu-okta: What happens if i pass in a value other than 'eq'? Ie gt or lt.  Should we explicitly say we only support the eq operator for kid comparisons?

@mystiberry-okta thoughts?

ashishgupta-okta: We don't support any of those expressions (le, gt, lt etc.) for other filters as well. We only support 'eq'. Also, we only support single expressions here, no ORs and ANDs, this is mentioned in the doc, may be we could add that we don't support other comparators there? 
On the other hand, it doesn't make much sense to support those on the first place since all the filters use `id's` which are strings.

mystiberry-okta: The docs are unclear about this. If it's easy, I'd recommend saying just eq and no comparison operators, but there is a larger issue which I haven't thought about for a while. This topic says "go look at the resource-specific documentation", but most of the resource-specific documentation doesn't specify limits. It'd be really great for the customer if we did here; it's on me to sort this out for other resources. http://developer.okta.com/docs/api/getting_started/design_principles.html#filtering

_source/_docs/api/postman/apps.json
```diff
@@ -710,22 +725,22 @@
...
 			"description": "",
-			"collectionId": "dd82f5f5-ea36-aade-5fd3-0dac0245db69",
+			"collectionId": "ff632500-bef8-294c-fcc5-7543fdcd79f1",
 			"responses": [],
-			"rawModeData": "{\n  \"profile\": {\n    \"roles\": [\n        \"CEO\"\n    ],\n    \"accountType\": \"FREE\",\n    \"role\": \"CEO\",\n    \"region\": \"Americas\",\n    \"groups\": [\n        \"engineering@clouditude.net\"\n    ]\n  } \n}"
+			"rawModeData": ""
 		}
 	]
-}
+}
```
mystiberry-okta: Um, is this the same file as in PR 418? Are you going to remove it? The other two files in this PR look great to me. But I don't want to merge this tomorrow if it's going to cause trouble with PR 418. Thanks!

ashishgupta-okta: This will not be merged this week in the apiDocs. Related okta-core changes will go out this week, apiDoc should be merged next week.

royalchan-okta:
Does this need a shoutout that the KEY_ROLLOVER is required somewhere?

_source/_docs/api/resources/apps.md
ashishgupta-okta: Updated,

ashishgupta-okta:
@royalchan-okta Added a note.

royalchan-okta:
🚢

mystiberry-okta:
Will publish for .40 refresh.

PR #415, opened 2016-09-26
OSW documentation update to v1.6.0

raphaellondner-okta:
Removed unnecessary parameters, split OAuth and social providers documentation and clarified some OIDC/OAuth parameters

### DIFF
--- a/_source/_code/javascript/okta_sign-in_widget.md
+++ b/_source/_code/javascript/okta_sign-in_widget.md
@@ -5,10 +5,11 @@ excerpt: Easily add sign-in capabilities with Okta's pre-packaged UI component.
 redirect_from:
     - "/docs/guides/okta_sign-in_widget.html"
 ---
+# Getting Started with the Okta Sign-In Widget
 
 ![Screenshot of basic Okta Sign-In Widget](/assets/img/okta-signin.png)
 
-# Introduction
+## Introduction
 
 The Okta Sign-In Widget is a JavaScript widget from Okta that gives
 you a fully featured and customizable login experience which
@@ -75,21 +76,21 @@ and easily set up a fully featured login experience:
 ~~~ html
 <head>
 	<!-- Core widget js and css -->
-	<script src="https://ok1static.oktacdn.com/assets/js/sdk/okta-signin-widget/1.4.0/js/okta-sign-in.min.js" type="text/javascript"></script>
-	<link href="https://ok1static.oktacdn.com/assets/js/sdk/okta-signin-widget/1.4.0/css/okta-sign-in.min.css" type="text/css" rel="stylesheet">
+	<script src="https://ok1static.oktacdn.com/assets/js/sdk/okta-signin-widget/1.6.0/js/okta-sign-in.min.js" type="text/javascript"></script>
+	<link href="https://ok1static.oktacdn.com/assets/js/sdk/okta-signin-widget/1.6.0/css/okta-sign-in.min.css" type="text/css" rel="stylesheet">
 	<!-- Customizable css theme options. Link your own customized copy of this file or override styles in-line -->
-	<link href="https://ok1static.oktacdn.com/assets/js/sdk/okta-signin-widget/1.4.0/css/okta-theme.css" type="text/css" rel="stylesheet">
+	<link href="https://ok1static.oktacdn.com/assets/js/sdk/okta-signin-widget/1.6.0/css/okta-theme.css" type="text/css" rel="stylesheet">
 </head>
 <body>
 	<div id="okta-login-container"></div>
 	<script type="text/javascript">
-		var baseUrl = 'https://example.okta.com';
-		var oktaSignIn = new OktaSignIn({baseUrl: baseUrl});
+		var orgUrl = 'https://example.okta.com';
+		var oktaSignIn = new OktaSignIn({baseUrl: orgUrl});
 		
 		oktaSignIn.renderEl(
 		  { el: '#okta-login-container' },
 		  function (res) {
-		    if (res.status === 'SUCCESS') { res.session.setCookieAndRedirect('https://example.okta.com'); }
+		    if (res.status === 'SUCCESS') { res.session.setCookieAndRedirect(orgUrl); }
 		  }
 		);
 	</script>
@@ -101,10 +102,9 @@ Here is what is happening in the HTML above:
 First, in the `<head>`
 tag, we include the  `okta-sign-in.min.js` and
 `okta-sign-in.min.css` files. These files have all of the
-logic for the Okta Sign-In Widget. 
+logic and default stylesheets used by the Okta Sign-In Widget. 
 
-We also include the `okta-theme.css` file, which contains all
-of the styling information for the Okta Sign-In Widget.
+We also include the `okta-theme.css` file, which essentially adds the blue theme you get in the login page of your Okta organization if you have enabled the New Sign-In Page setting in `Admin -> Settings -> Appearance` in the `Sign-In Configuration` section. We can choose not to include this okta-theme.css stylesheet if you so wish.
 
 In the `<body>`, we add a `<div>` tag with an "`id`" of
 `okta-login-container`  &#x2013; you can use any "`id`" value in this tag,
@@ -113,13 +113,13 @@ we are just using `okta-login-container` here for the sake of clarity.
 Next, we add some JavaScript code to insert the Okta Sign-In Widget
 into the `okta-login-container` tag. In this short example, this
 JavaScript is included at the end of the file. However, when you use
-this code on your own website, you will need to run these functions
+this code on your own web site, you will need to run these functions
 in the parts of your code that are run when the DOM is ready.
 
-Here is what that JavaScript code is doing: First, the line below
+Here is what that JavaScript code is doing: first, the line below
 initializes the Okta Sign-In Widget object, note that 
 the `baseUrl` value **MUST** be the domain for *your* Okta
-organization.
+organization. If you don't already have an Okta organization, please read the "Creation an Okta organization" section below. 
 
 For example, if your Okta organization is "`acme.okta.com`" you
 would replace the string "`example.okta.com`" below with
@@ -127,12 +127,12 @@ would replace the string "`example.okta.com`" below with
 
 ~~~ javascript
 var baseUrl = 'https://example.okta.com';
-var oktaSignIn = new OktaSignIn({baseUrl: baseUrl});
+var oktaSignIn = new OktaSignIn({baseUrl: orgUrl});
 ~~~
 
 Finally, the lines below actually render the Okta Sign-In
 Widget. Note that the value for `el` can be whatever `id` that you 
-specify, also note that we only define a "SUCCESS" callback. In a
+specify, also note that we only define a "SUCCESS" callback in which we create an Okta session and redirect the browser to the Okta organization, which will log the user straight into her Okta dashboard. In a
 production environment, you will want to handle statuses beyond
 "SUCCESS", you will likely also want to define an "ERROR"
 callback as well. 
@@ -141,7 +141,7 @@ callback as well.
 oktaSignIn.renderEl(
   { el: '#okta-login-container' },
   function (res) {
-    if (res.status === 'SUCCESS') { res.session.setCookieAndRedirect('https://example.com/'); }
+    if (res.status === 'SUCCESS') { res.session.setCookieAndRedirect(orgUrl); }
   }
 );
 ~~~
@@ -173,6 +173,16 @@ organization, or do not have an Okta organization, you will need to
 create an [Okta Developer Edition](https://www.okta.com/developer/signup/) account for yourself before
 continuing on the steps below.
 
+## Configuring CORS support on your Okta organization
+
+This step is necessary for Okta to accept authentication requests from your application through the Sign-In Widget.
+
+You can enable CORS by following the steps in our guide for
+[guide for Enabling CORS](http://developer.okta.com/docs/api/getting_started/enabling_cors.html). Configure CORS using the same URL for the
+web server you are using to host the HTML for the Okta Sign-In Widget (see below for instructions). For example, if you plan to host your Sign-In Widget in a page accessible under `http://localhost:8000`, you will need to add `http://localhost:8000` as a trusted CORS origin in Okta.
+
+To do so, navigate to the `Admin -> Security -> API` page of your Okta dashboard and in the `CORS` tab, add the url of the web site that will host the Okta Sign-In Widget. For instance, if you want to test it on a development machine on a web site hosted at `http://localhost:8081`, just add `http://localhost:8081` in the CORS multi-line text box, click `Enable CORS for the following base URLs` and press `Save`. [will need to be updated when we GA the new Trusted Origins tab].
+
 ## Creating an HTML file with the widget code
 
 The first step of getting the Okta Sign-In Widget set up for your Okta
@@ -190,15 +200,16 @@ Copy this to a file named `login-to-okta.html`:
       <meta charset="utf-8">
       <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1">
       <meta name="viewport" content="width=device-width, initial-scale=1">
-      <title>Example Okta SignIn Widget</title>
+      <title>Example Okta Sign-In Widget</title>
     
       <!--[if lt IE 9]>
           <script src="//cdnjs.cloudflare.com/ajax/libs/html5shiv/3.7.2/html5shiv.min.js"></script>
       <![endif]-->
     
-      <script src="https://ok1static.oktacdn.com/assets/js/sdk/okta-signin-widget/1.4.0/js/okta-sign-in.min.js" type="text/javascript"></script>
-      <link href="https://ok1static.oktacdn.com/assets/js/sdk/okta-signin-widget/1.4.0/css/okta-sign-in.min.css" type="text/css" rel="stylesheet">
-      <link href="https://ok1static.oktacdn.com/assets/js/sdk/okta-signin-widget/1.4.0/css/okta-theme.css" type="text/css" rel="stylesheet">
+      <script src="https://ok1static.oktacdn.com/assets/js/sdk/okta-signin-widget/1.6.0/js/okta-sign-in.min.js" type="text/javascript"></script>
+      <link href="https://ok1static.oktacdn.com/assets/js/sdk/okta-signin-widget/1.6.0/css/okta-sign-in.min.css" type="text/css" rel="stylesheet">
+      <!-- Optional, customizable css theme options. Link your own customized copy of this file or override styles in-line -->
+      <link href="https://ok1static.oktacdn.com/assets/js/sdk/okta-signin-widget/1.6.0/css/okta-theme.css" type="text/css" rel="stylesheet">
     </head>
     <body>
       <!-- Render the login widget here -->
@@ -206,21 +217,24 @@ Copy this to a file named `login-to-okta.html`:
     
       <!-- Script to init the widget -->
       <script>
-        var baseUrl = 'https://example.okta.com';
-        var oktaSignIn = new OktaSignIn({baseUrl: baseUrl});
+        var orgUrl = 'https://example.okta.com';
+        var redirectUrl = 'http://localhost:8000/signed-in.html';
+        var oktaSignIn = new OktaSignIn({baseUrl: orgUrl});
     
         oktaSignIn.renderEl(
           { el: '#okta-login-container' },
           function (res) {
             if (res.status === 'SUCCESS') {
               console.log('User %s successfully authenticated %o', res.user.profile.login, res.user);
-              res.session.setCookieAndRedirect('https://example.com/');
+              res.session.setCookieAndRedirect(redirectUrl);
             }
           }
         );
       </script>
     </body>
     </html>
+
+Note that in this code snippet, we are redirecting the user to a custom page we host locally at `http://localhost:8081/signed-in.html` instead of redirecting the user to her Okta dashboard (as was done in the simple example section above)
 
 ## Modify the HTML for your Okta organization
 
@@ -239,23 +253,14 @@ would be "`acme.okta.com`". Likewise, if you are using an Okta
 Developer Edition organization, your domain will be something like
 "`dev-12345.oktapreview.com`".
 
-There are four lines that you will need to modify in
-`login-to-okta.html`, modify each line, replacing
-`example.okta.com` with your fully qualified domain name:
+There is just one line you will need to modify in
+`login-to-okta.html` to get the sample working in your own environment. Simply  update the two lines below to match your environment with your fully qualified domain name in the following line:
 
-1.  The `<script>` tag for `okta-sign-in.js`:
-    
-        <script src="https://ok1static.oktacdn.com/assets/js/sdk/okta-signin-widget/1.4.0/js/okta-sign-in.min.js" type="text/javascript"></script>
-2.  The `<link>` tag for `okta-sign-in.css`:
-    
-        <link href="https://ok1static.oktacdn.com/assets/js/sdk/okta-signin-widget/1.4.0/css/okta-sign-in.min.css" type="text/css" rel="stylesheet">
-3.  The `<link>` tag for `okta-theme.css`:
-    
-        <link href="https://ok1static.oktacdn.com/assets/js/sdk/okta-signin-widget/1.4.0/css/okta-theme.css" type="text/css" rel="stylesheet">
-4.  The string passed as the `baseUrl` to the `OktaSignIn` constructor:
+		var orgUrl = 'https://example.okta.com';
+       var redirectUrl = 'http://localhost:8000/signed-in.html';
 
-        var baseUrl = 'https://example.okta.com';
-        var oktaSignIn = new OktaSignIn({baseUrl: baseUrl});
+Replace `example.okta.com` in the `orgUrl` variable with the fully qualified domain name of your organization.   
+Replace also `http://localhost:8000/signed-in.html` in the `redirectUrl` variable with the url of the page in your own web application where you would like to redirect the user to after a successful login.
 
 ## Copy the HTML to a web server
 
@@ -264,7 +269,7 @@ get this file hosted on a web server. The web server you use will
 depend on your development environment and what you are comfortable
 with using. 
 
-If you aren't sure what web server to use, or don't have one set up
+If you aren't sure which web server to use, or don't have one set up
 already, an easy web server to set up on Mac OS X or GNU/Linux
 is the `SimpleHTTPServer` that is included with `python`.
 
@@ -280,19 +285,6 @@ running a simple HTTP server in Node.js, which you can do as follows:
     $ npm install -g http-server
     
     $ http-server
-
-## Configuring CORS support on your Okta organization
-
-The last step that you'll need to do before you can test out
-the Okta Sign-In Widget is to enable CORS for the web server you just set
-up.
-
-You can enable CORS by following the steps in our guide for
-[guide for Enabling CORS](http://developer.okta.com/docs/api/getting_started/enabling_cors.html). Configure CORS using the same URL for the
-web server you are using to host the HTML for the Okta Sign-In Widget. For
-example, if you used the `python -m SimpleHTTPServer` example above, the URL
-you will use will likely be "`http://localhost:8000`" (assuming that the server is
-started on TCP port 8000 of your local machine).
 
 ## Testing the Okta Sign-In Widget
 
@@ -314,7 +306,7 @@ Here are two things for you to try:
     For example: Try using "invalid@example.com" as the user name and
     "invalid" as the password. You should see an error.
 2.  Try using a valid user name and password. If everything works,
-    you will be redirected to `https://example.com`.
+    you will be redirected to `http://localhost:8000/signed-in.html`.
 3.  [Configure Multifactor Authentication](https://support.okta.com/help/articles/Knowledge_Article/27315047-Configuring-Multifactor-Authentication)
     for your Okta org and try logging in using the Okta Sign-In Widget.
     
@@ -327,10 +319,10 @@ In the example above, you set up a very basic version of
 the Okta Sign-In Widget. Now that you've seen it in action, it's time to
 start configuring the widget for your specific needs.
 
-the Okta Sign-In Widget is fully customizable via CSS and JavaScript. 
-Stylistic customizations, changing the way that the widget looks, is
-done via CSS. Customization of widget features, the way that the
-widget works, and text labels for the widget are done via JavaScript. 
+The Okta Sign-In Widget is fully customizable via CSS and JavaScript. 
+Stylistic customizations (how the widget looks) are
+done via CSS. Customizations of widget features (how the
+widget works), as well as text labels for the widget, are done via JavaScript. 
 
 The sections below go into detail on how to customize
 the Okta Sign-In Widget using CSS and JavaScript.
@@ -346,9 +338,9 @@ Okta Sign-In Widget is created when the `renderEl()` JavaScript
 method is called. When the `renderEl()` method is called, the
 Okta Sign-In Widget will be created as a `<div>` tag with an
 `id` of `okta-sign-in` which will be inserted inside of the tag that
-you specified in the `el` option to `renderEl` method.
+you specified in the `el` option of the `renderEl` method.
 
-Here is what the opening tag for the Okta Sign-In Widget will look like:
+Here is what the opening tag for the Okta Sign-In Widget looks like:
 
     <div id="okta-sign-in" class="auth-container main-container no-beacon">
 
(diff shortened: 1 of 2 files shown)


_source/_code/javascript/okta_sign-in_widget.md
```diff
@@ -101,10 +102,9 @@ Here is what is happening in the HTML above:
 First, in the `<head>`
 tag, we include the  `okta-sign-in.min.js` and
 `okta-sign-in.min.css` files. These files have all of the
-logic for the Okta Sign-In Widget. 
+logic and default stylesheets used by the Okta Sign-In Widget. 
 
-We also include the `okta-theme.css` file, which contains all
-of the styling information for the Okta Sign-In Widget.
+We also include the `okta-theme.css` file, which essentially adds the blue theme you get in the login page of your Okta organization if you have enabled the New Sign-In Page setting in `Admin -> Settings -> Appearance` in the `Sign-In Configuration` section. We can choose not to include this okta-theme.css stylesheet if you so wish.
```
rchild-okta: "which essentially adds" -> "which adds Okta's own custom theme"

"We can choose" -> "You can choose"

And in general, now might be a good time to mention that they can use the okta-theme.css file as a template to create their own custom styles.

_source/_code/javascript/okta_sign-in_widget.md
```diff
@@ -113,26 +113,26 @@ we are just using `okta-login-container` here for the sake of clarity.
...
 "`acme.okta.com`":
 
 ~~~ javascript
 var baseUrl = 'https://example.okta.com';
-var oktaSignIn = new OktaSignIn({baseUrl: baseUrl});
+var oktaSignIn = new OktaSignIn({baseUrl: orgUrl});
 ~~~
 
 Finally, the lines below actually render the Okta Sign-In
 Widget. Note that the value for `el` can be whatever `id` that you
```
rchild-okta: Might be better to be more explicit:

Note that the value for `el` can be any selector of your choice - "#okta-login-container" is a selector for an element in the html that has an id attribute of "okta-login-container".

_source/_code/javascript/okta_sign-in_widget.md
```diff
@@ -113,26 +113,26 @@ we are just using `okta-login-container` here for the sake of clarity.
...
 ~~~ javascript
 var baseUrl = 'https://example.okta.com';
-var oktaSignIn = new OktaSignIn({baseUrl: baseUrl});
+var oktaSignIn = new OktaSignIn({baseUrl: orgUrl});
 ~~~
 
 Finally, the lines below actually render the Okta Sign-In
 Widget. Note that the value for `el` can be whatever `id` that you 
-specify, also note that we only define a "SUCCESS" callback. In a
+specify, also note that we only define a "SUCCESS" callback in which we create an Okta session and redirect the browser to the Okta organization, which will log the user straight into her Okta dashboard. In a
```
rchild-okta: Probably should break the sentence into two here (replace , with . )

Also am wondering about our convention around using "him", "her", or "their" <-- would probably prefer "their". @mystiberry-okta ?

_source/_code/javascript/okta_sign-in_widget.md
```diff
@@ -173,6 +173,16 @@ organization, or do not have an Okta organization, you will need to
 create an [Okta Developer Edition](https://www.okta.com/developer/signup/) account for yourself before
 continuing on the steps below.
 
+## Configuring CORS support on your Okta organization
```
rchild-okta: on -> for

_source/_code/javascript/okta_sign-in_widget.md
```diff
@@ -173,6 +173,16 @@ organization, or do not have an Okta organization, you will need to
...
+
+This step is necessary for Okta to accept authentication requests from your application through the Sign-In Widget.
```
rchild-okta: Might be nice to be more specific:

Because your application is hosted on a different origin than Okta's API endpoints, you will need to enable CORS support in the Okta admin application.

_source/_code/javascript/okta_sign-in_widget.md
```diff
@@ -173,6 +173,16 @@ organization, or do not have an Okta organization, you will need to
...
+
+You can enable CORS by following the steps in our guide for
+[guide for Enabling CORS](http://developer.okta.com/docs/api/getting_started/enabling_cors.html). Configure CORS using the same URL for the
```
rchild-okta: same URL -> same base url

_source/_code/javascript/okta_sign-in_widget.md
```diff
@@ -173,6 +173,16 @@ organization, or do not have an Okta organization, you will need to
...
+web server you are using to host the HTML for the Okta Sign-In Widget (see below for instructions). For example, if you plan to host your Sign-In Widget in a page accessible under `http://localhost:8000`, you will need to add `http://localhost:8000` as a trusted CORS origin in Okta.
+
+To do so, navigate to the `Admin -> Security -> API` page of your Okta dashboard and in the `CORS` tab, add the url of the web site that will host the Okta Sign-In Widget. For instance, if you want to test it on a development machine on a web site hosted at `http://localhost:8081`, just add `http://localhost:8081` in the CORS multi-line text box, click `Enable CORS for the following base URLs` and press `Save`. [will need to be updated when we GA the new Trusted Origins tab].
```
rchild-okta: Question about general convention in our docs - for multi step flows, is it better to break this out into an ordered list of steps?

_source/_code/javascript/okta_sign-in_widget.md
```diff
@@ -239,23 +253,14 @@ would be "`acme.okta.com`". Likewise, if you are using an Okta
...
-    
-        <link href="https://ok1static.oktacdn.com/assets/js/sdk/okta-signin-widget/1.4.0/css/okta-theme.css" type="text/css" rel="stylesheet">
-4.  The string passed as the `baseUrl` to the `OktaSignIn` constructor:
+		var orgUrl = 'https://example.okta.com';
+       var redirectUrl = 'http://localhost:8000/signed-in.html';
 
-        var baseUrl = 'https://example.okta.com';
-        var oktaSignIn = new OktaSignIn({baseUrl: baseUrl});
+Replace `example.okta.com` in the `orgUrl` variable with the fully qualified domain name of your organization.   
+Replace also `http://localhost:8000/signed-in.html` in the `redirectUrl` variable with the url of the page in your own web application where you would like to redirect the user to after a successful login.
```
rchild-okta: redirect the user to -> redirect the user

_source/_code/javascript/okta_sign-in_widget_ref.md
```diff
@@ -73,12 +70,13 @@ oktaSignIn.session.exists(function (exists) {
...
-`stateToken`    | Bootstrap the widget in a specific flow (e.g., Enroll MFA or MFA challenge). | String    | No           | None
-`logo`          | Logo to use in the widget (e.g., `https://acme.com/assets/logo/acme-logo.png`). | String    | No           | `Okta logo`
-`username`      | Bootstrap the widget with a username i.e., pre-fill the username in the widget. | String    | No           | None
-`transformUsername` | Transform the username before sending the request. The function is passed the username and the operation as parameters. [Example](#transformusername-example) | Function   | No           | None    
+`baseUrl`       | The base URL for your Okta organization. (e.g., `https://acme.okta.com`, `https://acme.oktapreview.com` or `https://acme.okta-emea.com`).  |String    | Yes          | None
+`recoveryToken` | Bootstraps the widget in the recovery flow (e.g., Unlock Account or Forgot Password). | String    | No           | None
+`stateToken`    | Bootstraps the widget in a specific flow (e.g., Enroll MFA or MFA challenge). | String    | No           | None
```
rchild-okta: I was thinking of removing this parameter in my docs - is the stateToken something the developer will get from other okta endpoints? I'm guessing its mostly internal.

_source/_code/javascript/okta_sign-in_widget_ref.md
```diff
@@ -138,6 +135,10 @@ Enable or disable widget functionality with the following options for `features`
     
 -   `router` `(default: false)`
     Update the browser location bar with a route on navigation.
+
+-   `windowsVerify` `(default: false)`
+    Need to get input from engineering.
```
rchild-okta: When enrolling Okta Verify, this option specifies whether we show the windows phone icon/instructions:
https://oktawiki.atlassian.net/wiki/display/PM/MFA+Enrollment+Re-design?preview=/75399397/75399437/MFA_oktaVerifyOld_step1.png

_source/_code/javascript/okta_sign-in_widget_ref.md
```diff
@@ -202,24 +200,90 @@ The following labels are among the most frequently used:
...
 
-Options for the [OpenID Connect](/docs/api/resources/oidc.html#request-parameters) authentication flow.
-This flow is required for social authentication, and requires OAuth client registration with Okta.
-For instructions, see [Social Authentication](/docs/api/resources/social_authentication.html).
+For OpenID Connect and OAuth setup instructions, please take a look at these [setup instructions](https://github.com/oktadeveloper/okta-oauth-spa-authjs-osw/blob/master/Okta-OIDC_SPA_JS-OSW_DevSetupGuide.pdf).
 
-#### Example OIDC flow
+A working [OpenID Connect sample with the Okta Sign-In Widget](https://github.com/oktadeveloper/okta-oauth-spa-authjs-osw) can also be found on GitHub.
+
+The Okta Sign-In Widget also provides support for social authentication (Facebook, LinkeIn and Google), which is built on top OpenID Connect and OAuth 2.0. For social authentication setup instructions, please refer to [Social Authentication](/docs/api/resources/social_authentication.html).
```
rchild-okta: "on top of"

_source/_code/javascript/okta_sign-in_widget_ref.md
```diff
@@ -202,24 +200,90 @@ The following labels are among the most frequently used:
...
+
+
+Important note: to get the Sign-In Widget to work with OpenID Connect, OAuth and social authentication, you must allow iframe embedding in the `Admin Console -> Settings -> Customization` page (`IFrame Embedding` section).
```
rchild-okta: Oh, this is interesting - have you tried it without this setting?

@varrunramani-okta, @tthompson-okta, should the authorization page be excluded from the X-Frame-Options check since it's a pretty common scenario to embed it in an iframe (our sessionToken -> idToken flows)?

_source/_code/javascript/okta_sign-in_widget_ref.md
```diff
@@ -202,24 +200,90 @@ The following labels are among the most frequently used:
...
+ `oAuthTimeout` | Timeout for OIDC authentication flow requests. | Number | No | `120000`
+ `authParams`   | Authentication parameters for OIDC. [List](#authParams-parameters) |See [List](#authParams-parameters) |See [List](#authParams-parameters) | See [List](#authParams-parameters)
+ 
+##### authParams parameters
+
+You can use any of the following parameters for `authParams`.
+
+ authParams Parameter       | Description                              |  Type    |  Required    | Default   | Valid Values
+--------------------------  | ---------------------------------------  | ------   | -----------  | -------   | ----------------
+`display`                   | Specify how to display the authentication UI. | String  | No           | `none` (for Okta) and `popup` (External IdP such as social) | `none`, `popup`, `page`
```
rchild-okta: Might be clearer to go into more detail here (I had to look into the code last week to see what was going on):
- 'popup' <-- when you click the social button, opens a popup
- 'page' <-- when you click the social button, will do a full page redirect (also this won't be out until the next version, so maybe we can leave it for now)

_source/_code/javascript/okta_sign-in_widget.md
raphaellondner-okta: Agree. Good catch on 'we' vs. 'you' too.

mystiberry-okta: +1

_source/_code/javascript/okta_sign-in_widget.md
raphaellondner-okta: You redirect a user TO a page, don't you? This is really a grammar question here and I admit I'm not an expert... Most likely, the common usage is not to include "to" in that context...

_source/_code/javascript/okta_sign-in_widget.md
mystiberry-okta: So, "their" is still third-person plural. @tthompson-okta pushed back against singular pronoun. These days tech writers take this approach:

-- avoid if possible. In this case, "the Okta dashboard" seems clear to me.
-- if you have to use "they" or "their," match the number case with the rest of the sentence, in this case, "which logs users straight into their Okta dashboard." If you've been talking about a single user in the example, this may be confusing. 

Also, please don't use future tense unless absolutely necessary. Should be "which logs the user directly into the Okta dashboard."   "straight" is colloquial. I try to avoid adverbs but here it really clarifies the action.

_source/_code/javascript/okta_sign-in_widget_ref.md
raphaellondner-okta: Agree. I guess that if we include it here, we must have some sort of sample demonstrating how to use it (along with 'recoveryToken') because it's not quite clear how stateToken can and should be used...

_source/_code/javascript/okta_sign-in_widget_ref.md
raphaellondner-okta: Also correct spelling of 'LinkedIn' (instead of 'LinkeIn')

_source/_code/javascript/okta_sign-in_widget.md
mystiberry-okta: should it be "an application"? It's not always the developer's app? (if it is, then leave as is). 

What is an "origin," that sounds weird. Nothing is hosted on an endpoint anyway, it's accessible from an endpoint?

_source/_code/javascript/okta_sign-in_widget_ref.md
raphaellondner-okta: Yes, I've tried it and it didn't work for me if you don't enable it in OIDC scenarios. I understand that this is necessary because we really on iframes to initiate the calls to the /authorize endpoint, which is not CORS-enabled (hence the use of iframes)

_source/_code/javascript/okta_sign-in_widget.md
mystiberry-okta: Always best to make ordered steps explicit. Markdown has limited support for fancy indenting or breaks within a list, so if that's the problem you were avoiding, ping me and we'll try to figure something out.

_source/_code/javascript/okta_sign-in_widget_ref.md
raphaellondner-okta: Agree. BTW, 'page' has been there for a while, but was never properly documented. I'm fine about excluding it from the docs for now.

mystiberry-okta: If you choose to more thoroughly document the values, the convention is to describe them below the table and link to that section. Too much content in the description table makes it hard to read.

_source/_code/javascript/okta_sign-in_widget.md
```diff
@@ -75,21 +76,21 @@ and easily set up a fully featured login experience:
...
 	<!-- Customizable css theme options. Link your own customized copy of this file or override styles in-line -->
-	<link href="https://ok1static.oktacdn.com/assets/js/sdk/okta-signin-widget/1.4.0/css/okta-theme.css" type="text/css" rel="stylesheet">
+	<link href="https://ok1static.oktacdn.com/assets/js/sdk/okta-signin-widget/1.6.0/css/okta-theme.css" type="text/css" rel="stylesheet">
 </head>
 <body>
 	<div id="okta-login-container"></div>
 	<script type="text/javascript">
-		var baseUrl = 'https://example.okta.com';
-		var oktaSignIn = new OktaSignIn({baseUrl: baseUrl});
+		var orgUrl = 'https://example.okta.com';
```
mystiberry-okta: Agree with orgBaseUrl, since baseUrl is a commonly understood concept. oktaDomain might be more confusing than not, since from the customer's POV it's *their* domain...?

_source/_code/javascript/okta_sign-in_widget.md
```diff
@@ -113,26 +113,26 @@ we are just using `okta-login-container` here for the sake of clarity.
 Next, we add some JavaScript code to insert the Okta Sign-In Widget
 into the `okta-login-container` tag. In this short example, this
 JavaScript is included at the end of the file. However, when you use
-this code on your own website, you will need to run these functions
+this code on your own web site, you will need to run these functions
```
mystiberry-okta: website is one word now. The alternative is Web site, but I think CMS or someone just ruled against Web. We should go with the more modern convention. And yes, we need a style guide so people don't have to thrash over the details.

_source/_code/javascript/okta_sign-in_widget.md
```diff
@@ -190,38 +200,42 @@ Copy this to a file named `login-to-okta.html`:
...
-              res.session.setCookieAndRedirect('https://example.com/');
+              res.session.setCookieAndRedirect(redirectUrl);
             }
           }
         );
       </script>
     </body>
     </html>
 
+Note that in this code snippet, we are redirecting the user to a custom page we host locally at `http://localhost:8081/signed-in.html` instead of redirecting the user to her Okta dashboard (as was done in the simple example section above)
```
mystiberry-okta: Best to keep the verbs simple: "we redirect the user." and "the Okta dashboard" will solve that pronoun-gender problem. Or "redirecting users to their Okta dashboard."

_source/_code/javascript/okta_sign-in_widget.md
mystiberry-okta: "where" takes the place of "to" in the syntax:

You would like to redirect the user to > You would like to redirect the user where > Where you would like to redirect the user.

_source/_code/javascript/okta_sign-in_widget.md
```diff
@@ -327,10 +319,10 @@ In the example above, you set up a very basic version of
...
 start configuring the widget for your specific needs.
 
-the Okta Sign-In Widget is fully customizable via CSS and JavaScript. 
-Stylistic customizations, changing the way that the widget looks, is
-done via CSS. Customization of widget features, the way that the
-widget works, and text labels for the widget are done via JavaScript. 
+The Okta Sign-In Widget is fully customizable via CSS and JavaScript. 
+Stylistic customizations (how the widget looks) are
+done via CSS. Customizations of widget features (how the
+widget works), as well as text labels for the widget, are done via JavaScript.
```
mystiberry-okta: Not required, but you could simplify this with active verbs: "Change how the widget looks with CSS, and change how the widget works, including modifying text labels, with JavaScript."  It doesn't seem like we use the phrases "stylistic customizations" or "Customizations of widget features" anywhere; if so, they can be skipped.

_source/_code/javascript/okta_sign-in_widget.md
```diff
@@ -357,50 +349,44 @@ is up to you. Customization of the widget itself will be done on
...
-present in the [okta-theme.css](https://ok1static.oktacdn.com/assets/js/sdk/okta-signin-widget/1.4.0/css/okta-theme.css) file, other elements in the widget
-will be subject to change and will cause your styling to break in
+present in the [okta-theme.css](https://ok1static.oktacdn.com/assets/js/sdk/okta-signin-widget/1.6.0/css/okta-theme.css) file as other stylistic elements in the widget may be subject to change and may cause your styling to break in
 future versions of the Okta Sign-In Widget.
 
 ### Example CSS styling for the Okta Sign-In Widget
 
 Below is some example CSS, which will give you an idea of how you
-can style the Okta Sign-In Widget. You can try out this CSS yourself
+can style the Okta Sign-In Widget. You can try out the CSS style below
```
mystiberry-okta: Let's keep the sentences as simple as possible:  "Try the following CSS style."

_source/_code/javascript/okta_sign-in_widget_ref.md
```diff
@@ -100,12 +98,13 @@ The following sections provide examples and option details for the last four con
     }
 ~~~
 
+The code above ensures that in case the current user clicked on the 'Forgot Password' or 'Unlock account' link, the sign-in widget will only proceed with the requested operation if and only if the user provided a username that ends with `@acme.com`. Otherwise, the sign-in widget will automatically append
```
mystiberry-okta: Sign-In Widget

mystiberry-okta: and "only proceeds" don't need "will". Also "if and only if" can be "if" since you already said "only". Best practice is to move "only" closer to the thing is modifies, so "...the Sign-In Widget sends the request only if the user provided..." would be the easiest read (if still technically correct).

_source/_code/javascript/okta_sign-in_widget_ref.md
```diff
@@ -118,18 +117,16 @@ Enable or disable widget functionality with the following options for `features`
...
--   `autoPush` `(default: false)`
+ -   `autoPush` `(default: false)`
     Display a checkbox to enable "Send push automatically" functionality in the MFA required flow.
         
 -   `smsRecovery` `(default: false)`       
     Allow users with a configured mobile phone number to recover their password using an SMS message.
     
+-   `callRecovery` `(default: false)`       
+    Allow users with a configured mobile phone number to recover their password using a voice call.
+    Note: only available starting v1.6.0 of the Sign-In Widget.
```
mystiberry-okta: Note: This option is available with version 1.6.0 and later of the Sign-In Widget.

_source/_code/javascript/okta_sign-in_widget_ref.md
mystiberry-okta: Can't merge until this is resolved.

_source/_code/javascript/okta_sign-in_widget_ref.md
mystiberry-okta: Not sure of the ordering of these new paragraphs, they aren't that well related. I'd like to see us use some headings to make the transitions less rough.

_source/_code/javascript/okta_sign-in_widget_ref.md
mystiberry-okta: Best practice is to just say "Important:" and capitalize "To" since that's a complete sentence.

_source/_code/javascript/okta_sign-in_widget_ref.md
```diff
@@ -256,30 +320,16 @@ oktaSignIn.renderEl({
   });
 ~~~
 
-#### OIDC Parameters
+#### Social Authentication Parameters
+
+The table below only highlights the __additional__ parameters necessary for social authentication, on top of the [OpenID Connect/OAuth parameters](#openid-connect-parameters).
```
mystiberry-okta: on top of or in addition to? Unless the additional parameters are a layer, I'd use "in addition to." I'd also get rid of the bold, and simplify:  "In addition to the OpenID Connect and OAuth parameters <link>, additional parameters are necessary for social authentication:

That way we get rid of the unnecessary self-reflecting language, avoid "only" which is a wee bit negative.

_source/_code/javascript/okta_sign-in_widget_ref.md
```diff
@@ -329,7 +376,7 @@ failureCallback       | Function  | Yes           | Failure callback to invoke o
 
 Manage your Okta session with session functions in the Sign-in widget SDK.
 
-#### Check for an existing session
+#### Checking for an existing session
```
mystiberry-okta: Please leave the all heading verbs in simple active form.

_source/_code/javascript/okta_sign-in_widget_ref.md
```diff
@@ -256,30 +320,16 @@ oktaSignIn.renderEl({
...
-
- authParams Parameter       | Description                              |  Type    |  Required    | Default   | Valid Values
---------------------------  | ---------------------------------------  | ------   | -----------  | -------   | ----------------
-`display`                   | Specify how to display authentication UI. | String  | No           | `none` (for Okta) and `popup` (External Idp) | `none`, `popup`, `page`
-`responseMode`              | Specify how the authorization response should be returned. | String | Yes | `okta_post_message` | `query`, `fragment`, `form_post`, `okta_post_message`
-`responseType`              | Specify the response type for OIDC authentication. The authorization code returned can later be exchanged for Access token or Refresh token. | String | Yes | `id_token` | `code`, `token`, `id_token`
-`scope`                     | Specify what information to make available in the `id_token`. `openid` is required. | Array | Yes | `['openid', 'email']` | `openid`, `email`, `profile`, `address`, `phone`    
-
-## Render the widget
+## Public functions
```
mystiberry-okta: You need a heading for .renderEl() to keep it parallel with the other functions, so I suggest *adding* "Public functions" instead of replacing "Render the widget,"

_source/_code/javascript/okta_sign-in_widget_ref.md
varrunramani-okta: Yea, this is kinda bad, we shouldn't be allowing iframe embedding for the whole org just for the sign-in widget. Ideally, we should shift to whitelisting origins. We have an item on a plate to support this and integrate it with the trusted origins feature. FYI @cwu-okta 

@rchild-okta We can try, but even if we override the XFrameOptions policy for the auth pages, I believe any pages post-auth cannot be viewed in an iFrame without the org setting. But all we need is the auth page in an iFrame and then do a redirect or open in new tab/window, this setting won't have to be enabled.

rchild-okta: @varrunramani-okta, yep - we don't care about any other Okta pages. We have an OIDC flow where we exchange a sessionToken for an idToken, and then postMessage that back to the parent frame. Once that's done, we close out the iframe (also, it's a hidden iframe on the page so the user is not interacting with it).

Do you have a jira for that roadmap item, and is there a timeline for working on it?

_source/_code/javascript/okta_sign-in_widget_ref.md
cwu-okta: Nit: Social Auth now supports Microsoft. Maybe instead of listing out the individual providers it become a comprehensive list, or say "for example" to let the reader know there are more providers than those listed.

rchild-okta: @cwu-okta, we'll need to think of a good way to keep in sync on these changes - currently, we haven't added Microsoft to our list of supported IDP's (there's a little bit of work we need to do on our end, i.e. adding icons for it, etc).

I created a jira to add this on our end:
https://oktainc.atlassian.net/browse/OKTA-102612

_source/_code/javascript/okta_sign-in_widget_ref.md
varrunramani-okta: @rchild-okta @tthompson-okta In that case we should have the OIDC endpoints that the sign-in widget uses have the X-Frame-Options override to prevent having to require all OIDC customers from having to turn on this flag(Federations). Since there is no UI element involved, there is no risk in clickjacking. In the mean time Eng-Security will work on enabling whitelisting and once thats done, we can go back and remove the overrides.

_source/_code/javascript/okta_sign-in_widget_ref.md
cwu-okta: We added it in March 2016 and I thougth that it had been communicated to the widget, but looks like it slipped through the cracks. Currently no future plans to add a new provider but, if we do, I'll make sure you're in the loop @rchild-okta

_source/_code/javascript/okta_sign-in_widget_ref.md
raphaellondner-okta: @cwu-okta I wasn't able to actually replicate that requirement in my standard OSW code sample and I can't remember the context in which I stumbled upon this as being a requirement. Agree to scratch that at this point, until I find what the specific use case is here.

rchild-okta: Also got confirmation from @lboyette-okta that we're fine going through the /authorize flow in an embedded iframe.

So, @raphaellondner-okta action item here is to remove this note in the docs since it's no longer an issue.

_source/_code/javascript/okta_sign-in_widget.md
raphaellondner-okta: @mystiberry-okta +1

_source/_code/javascript/okta_sign-in_widget.md
raphaellondner-okta: @mystiberry-okta We're actually moving from using CORS in our Admin UI and using "trusted origins" instead: https://oktawiki.atlassian.net/wiki/pages/viewpage.action?pageId=92834241

raphaellondner-okta: But changing "origin" to "endpoint" nonetheless

_source/_code/javascript/okta_sign-in_widget.md
raphaellondner-okta: @mystiberry-okta Thanks for the grammar lesson! 👍

_source/_code/javascript/okta_sign-in_widget_ref.md
raphaellondner-okta: @rchild-okta @mystiberry-okta I'll let you improve that part of the doc after we're done with that PR.

_source/_code/javascript/okta_sign-in_widget.md
```diff
@@ -131,10 +131,7 @@ var oktaSignIn = new OktaSignIn({baseUrl: orgUrl});
 ~~~
 
 Finally, the lines below actually render the Okta Sign-In
-Widget. Note that the value for `el` can be whatever `id` that you 
-specify, also note that we only define a "SUCCESS" callback in which we create an Okta session and redirect the browser to the Okta organization, which will log the user straight into her Okta dashboard. In a
-production environment, you will want to handle statuses beyond
-"SUCCESS", you will likely also want to define an "ERROR"
+Widget. Note that the value for `el` can be any selector of your choice - "#okta-login-container" is a selector for an element in the html that has an id attribute of "okta-login-container". Also note that we only define a "SUCCESS" callback in which we create an Okta session and redirect the browser to the Okta organization. This logs the user directly into the Okta dashboard. In a production environment, you will want to handle statuses beyond "SUCCESS", you will likely also want to define an "ERROR"
```
mystiberry-okta: We should capitalize HTML and ID (I know it's not very consistent yet), or if we mean a parameter or property named id, use the "tick marks" `id`

_source/_code/javascript/okta_sign-in_widget.md
```diff
@@ -173,15 +170,21 @@ organization, or do not have an Okta organization, you will need to
...
 
 You can enable CORS by following the steps in our guide for
-[guide for Enabling CORS](http://developer.okta.com/docs/api/getting_started/enabling_cors.html). Configure CORS using the same URL for the
-web server you are using to host the HTML for the Okta Sign-In Widget (see below for instructions). For example, if you plan to host your Sign-In Widget in a page accessible under `http://localhost:8000`, you will need to add `http://localhost:8000` as a trusted CORS origin in Okta.
+[guide for Enabling CORS](http://developer.okta.com/docs/api/getting_started/enabling_cors.html). Configure CORS using the same base url of the
+web server you are using to host the HTML for the Okta Sign-In Widget (see below for instructions). For example, if you plan to host your Sign-In Widget in a page accessible under `http://localhost:8000`, you will need to add `http://localhost:8000` as a trusted CORS endpoint in Okta.
+
+To do so, please follow the steps below:
+
+1. Navigate to the `Admin -> Security -> API` page of your Okta dashboard
```
mystiberry-okta: Full sentence needs a period.

_source/_code/javascript/okta_sign-in_widget.md
```diff
@@ -173,15 +170,21 @@ organization, or do not have an Okta organization, you will need to
...
-This step is necessary for Okta to accept authentication requests from your application through the Sign-In Widget.
+This step is necessary for Okta to accept authentication requests from an application through the Sign-In Widget.
 
 You can enable CORS by following the steps in our guide for
-[guide for Enabling CORS](http://developer.okta.com/docs/api/getting_started/enabling_cors.html). Configure CORS using the same URL for the
-web server you are using to host the HTML for the Okta Sign-In Widget (see below for instructions). For example, if you plan to host your Sign-In Widget in a page accessible under `http://localhost:8000`, you will need to add `http://localhost:8000` as a trusted CORS origin in Okta.
+[guide for Enabling CORS](http://developer.okta.com/docs/api/getting_started/enabling_cors.html). Configure CORS using the same base url of the
+web server you are using to host the HTML for the Okta Sign-In Widget (see below for instructions). For example, if you plan to host your Sign-In Widget in a page accessible under `http://localhost:8000`, you will need to add `http://localhost:8000` as a trusted CORS endpoint in Okta.
+
+To do so, please follow the steps below:
```
mystiberry-okta: We leave out the "please" and "thank you" phrases--they're absolutely correct in spoken English but weigh down written instructions. Why do we tell them to go somehwere else and then give them the instructions anyway?

_source/_code/javascript/okta_sign-in_widget.md
```diff
@@ -173,15 +170,21 @@ organization, or do not have an Okta organization, you will need to
...
+
+1. Navigate to the `Admin -> Security -> API` page of your Okta dashboard
+2. In the `CORS` tab, add the url of the site that will host the Okta Sign-In Widget. For instance, if you want to test it on a development machine on a web site hosted at `http://localhost:8081`, just add `http://localhost:8081` in the CORS multi-line text box.
```
mystiberry-okta: Delete "just".

_source/_code/javascript/okta_sign-in_widget.md
raphaellondner-okta: (thumbsup)

PR #416, opened 2016-09-26
36881 cb toc updates

codyrobert:
Hi Joël,

These are the CSS & JS changes for the responsive ToC as outlined. There are a few caveats:

1. On pages with extra large ToCs, there will be some lag in loading. It won't block page load, but will just won't load the ToC immediately. Ex. /docs/api/resources/authn.html
2. If an h1 is not on the page, it won't be added to the toc as detailed in the provided spec. Fix is just to add an h1 to that specific page Ex. /use_cases/integrate_with_okta/
3. I saw a few doc pages that were not using the `docs_page` layout and as a result do not load the docs.js file. This is where the new ToC code is located, so those will need to be updated. I did not make changes to any of these content pages myself. Ex. /use_cases/authentication/
4. As the site is built in jekyll, it doesn't really afford the possiblity to have css transitions between the pages in the left nav. That said, there is a class `.is-active` that can be added to any of the links in the left nav for the requisite highlight. I didn't adjust any of the generated / static content here in the sidebar.

Let me know if there's anything else we can address.

Thanks,
Cody Baker
Flickerbox

### DIFF
--- a/_source/_includes/sidebar.html
+++ b/_source/_includes/sidebar.html
@@ -1,6 +1,18 @@
-        <aside class="Sidebar">
-        <h2 class="Sidebar-location Sidebar-toggle h6">HTTP Verbs</h3>
-           <div class="Sidebar-close Sidebar-toggle"></div>
+<aside class="Sidebar">
+	
+    <div class="Sidebar-toggle MenuToggle">
+		<svg width="28px" height="28px" viewBox="0 0 28 28" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+		    <g stroke="none" stroke-width="1" fill="none" fill-rule="evenodd">
+		        <circle class="MenuToggle-circle" stroke="#000000" cx="14" cy="14" r="13.5"></circle>
+		        <rect class="MenuToggle-line1" fill="#000000" x="5.5" y="8" width="17" height="2"></rect>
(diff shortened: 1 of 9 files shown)


joelfranusic-okta:
Looking at this now.

johngronberg-okta:
Cody,
We reviewed the changes, and have a few things that need updating. Here are the issues we see. 

1.  Only the currently selected section's direct children should be showing in the TOC. (So, if we are in an h1, we should see only the h2's inside of it. For whichever h2 we are currently looking at, we should also see h3's, but no other h3's. Etc.) 
2. The TOC should be independently scrollable from the page (with logic for scrolling the TOC to the currently viewed page section, if the user scrolls the page, described in the wiki).
3. The highlighting in the nav does not always seem to line up with the part of the page I'm looking at at a given moment.
4. The page title should also be in the TOC, with the other sections indented from it.
5. The gray bar on the left of the TOC is not displaying correctly. Sometimes it floats without connecting to all the TOC sections.

Let me know if you can make those updates, or if you need further clarification. 

Thanks!

codyrobert:
Hi John,

I've made the requested changes and rebased via master to get this branch cleaned up again. I think I was able to accommodate everything but let me know if there's anything awry or if I can provide any clarity on specifics.

Thanks,
Cody

johngronberg-okta:
these changes were incorporated as part of #457

PR #420, opened 2016-10-03
Add access tokens and auth codes in okta-auth-js

lboyette-okta:
Resolves: OKTA-95163

@rchild-okta @raphaellondner-okta @mystiberry-okta

### DIFF
--- a/_source/_code/javascript/okta_auth_sdk.md
+++ b/_source/_code/javascript/okta_auth_sdk.md
@@ -6,7 +6,7 @@ excerpt: Easily add sign-in capabilities to your fully branded web site.
 
 # Introduction
 
-The Okta Auth SDK builds on top of our [Authentication API](/docs/api/resources/authn.html) to enable you to create a custom login experience using JavaScript.
+The Okta Auth SDK builds on top of our [Authentication API](/docs/api/resources/authn.html) and [OAuth 2.0 API](/docs/api/resources/oidc.html) to enable you to create a fully branded sign-in experience using JavaScript.
 
 # Prerequisites
 
@@ -18,7 +18,7 @@ The Okta Auth SDK builds on top of our [Authentication API](/docs/api/resources/
 Include the following script tag in your target web page:
 
 ~~~ html
-<script src="https://ok1static.oktacdn.com/assets/js/sdk/okta-auth-sdk-1.0.0.min.js" type="text/javascript"></script>
+<script src="https://ok1static.oktacdn.com/assets/js/sdk/okta-auth-js/1.5.0/okta-auth-js.min.js" type="text/javascript"></script>
 ~~~
 
 ## Authentication Flow
@@ -384,9 +384,9 @@ Returns a transaction with a user's previous state.
 authClient.tx.resume();
 ~~~
 
-## Social Authentication and OIDC
+## Social Authentication and OpenID Connect
 
-If you want to obtain an ID Token via [OIDC](/docs/api/resources/oidc.html) or [social authentication](/docs/guides/social_authentication.html), it's possible to manage the flow with the SDK. To do so, instantiate your client with a `clientId` and `redirectUri`.
+If you want to create a token via [OpenID Connect](/docs/api/resources/oidc.html) or [social authentication](/docs/guides/social_authentication.html), it's possible to manage the flow with this SDK. To do so, instantiate your client with a `clientId` and `redirectUri`. If you don't have either of those, [set up OpenID Connect with these instructions](https://github.com/oktadeveloper/okta-oauth-spa-authjs-osw/blob/master/Okta-OIDC_SPA_JS-OSW_DevSetupGuide.pdf).
 
 ~~~ javascript
 var authClient = new OktaAuth({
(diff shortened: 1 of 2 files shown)


_source/_code/javascript/okta_auth_sdk.md
```diff
@@ -396,46 +396,151 @@ var authClient = new OktaAuth({
...
+Multiple tokens
+
 ~~~ javascript
-authClient.idToken.authorize({
-  sessionToken: 'testSessionToken'
+authClient.token.getWithPopup({
+  responseType: ['token', 'id_token']
+})
+.then(function(arrayOfTokens) {
+  var accessToken = arrayOfTokens[0];
```
rchild-okta: I think it'd be nice to change "arrayOfTokens" to "tokenArray" and add a comment here about the order:
"The tokenArray is returned in the order that is specified in responseType - in this case, because 'token' is first and 'id_token' is second, the first element is the accessToken and the second token is the ID Token" <-- but less verbose

_source/_code/javascript/okta_auth_sdk.md
```diff
@@ -396,46 +396,151 @@ var authClient = new OktaAuth({
...
 ~~~
 
-### Get an ID Token from an IDP
+Each `responseType` has its own format:
+
 ~~~ javascript
-authClient.idToken.authorize({
+// id_token
+{
+  idToken: 'theIdTokenString',
```
rchild-okta: nice, I'm glad we have this section

_source/_code/javascript/okta_auth_sdk.md
```diff
@@ -396,46 +396,151 @@ var authClient = new OktaAuth({
...
+authClient.token.getWithPopup(options)
+.then(function(tokenOrTokens) {
+  // Manage token or tokens
+})
+.catch(function(err) {
+  // Handle OAuthError
+});
+~~~
+
+This is commonly used to create a token with an idp.
```
rchild-okta: with an **external** idp?

_source/_code/javascript/okta_auth_sdk.md
```diff
@@ -396,46 +396,151 @@ var authClient = new OktaAuth({
...
+authClient.token.parseFromUri()
+.then(function(tokenOrTokens) {
+  // Manage token or tokens
+})
+.catch(function(err) {
+  // Handle OAuthError
+});
+~~~
+
+This is commonly used to create an authorization code.
```
rchild-okta: I wonder if we should remove this line - after the thirdparty cookie discussions, seems like the redirect flow could also be perfectly valid for normal id_token flows. Would probably replace this with a line saying that you'll use this when you call "getWithRedirect"

_source/_code/javascript/okta_auth_sdk.md
```diff
@@ -396,46 +396,151 @@ var authClient = new OktaAuth({
...
+
+### Exchange a sessionToken for a token
+
+When you've obtained a sessionToken, or a session already exists, you can obtain a token or tokens without prompting the user to log in.
 
 ~~~ javascript
-authClient.idToken.refresh();
+authClient.token.getWithoutPrompt({
+  responseType: 'id_token', // or array of types
+  sessionToken: 'testSessionToken' // optional if a session already exists
```
rchild-okta: Would be nice to be more explicit in the comment, i.e. something along the lines of "If you've just authenticated using the authN API's and acquired a sessionToken, pass it in here. If the user already has an existing session with Okta, you don't need to pass anything in". Or something along those lines.

_source/_code/javascript/okta_auth_sdk.md
```diff
@@ -18,7 +18,7 @@ The Okta Auth SDK builds on top of our [Authentication API](/docs/api/resources/
 Include the following script tag in your target web page:
 
 ~~~ html
-<script src="https://ok1static.oktacdn.com/assets/js/sdk/okta-auth-sdk-1.0.0.min.js" type="text/javascript"></script>
+<script src="https://ok1static.oktacdn.com/assets/js/sdk/okta-auth-js/1.5.0/okta-auth-js.min.js" type="text/javascript"></script>
```
lboyette-okta: I can do that!

_source/_code/javascript/okta_auth_sdk.md
lboyette-okta: Yup. Will change

_source/_code/javascript/okta_auth_sdk.md
lboyette-okta: I'm not sure I understand this change. I'm trying to give some guidance about how to create an authorization code, without breaking out into another section. Would changing the wording to "This is often used to create an authorization code." be more clear? Saying it's often used doesn't bar people from using it to create id_tokens.

The third party cookie decision will change our default recommendation, but won't change this use case (trying to create an auth code).

_source/_code/javascript/okta_auth_sdk.md
lboyette-okta: I'll work something out

_source/_code/javascript/okta_auth_sdk.md
rchild-okta: Oh I see - I wasn't looking at the context of the change (thought you were just talking about the parseFromUri method). Your wording makes sense 👍

rchild-okta:
lgtm 🚀

mystiberry-okta:
@lboyette-okta Can you resolve the conflicts today so I can merge and ship with the .40 release? (I'd do it but am concerned about making a mistake in unfamiliar files).

_source/_code/javascript/okta_auth_sdk.md
```diff
@@ -6,7 +6,7 @@ excerpt: Easily add sign-in capabilities to your fully branded web site.
 
 # Introduction
 
-The Okta Auth SDK builds on top of our [Authentication API](/docs/api/resources/authn.html) to enable you to create a custom login experience using JavaScript.
+The Okta Auth SDK builds on top of our [Authentication API](/docs/api/resources/authn.html) and [OAuth 2.0 API](/docs/api/resources/authn.html) to enable you to create a custom login experience using JavaScript.
```
raphaellondner-okta: The link to the OAuth 2.0 API seems wrong (points to the AuthN API page). Also, don't we want to mention the Auth SDK is only built on top of our OpenID Connect implementation?

Last, I would recommend using "fully branded sign-in experience" instead of "custom login experience", because the OSW already provides a custom login experience, even if constrained by some stylistic and flow limitations.

_source/_code/javascript/okta_auth_sdk.md
```diff
@@ -386,7 +386,7 @@ authClient.tx.resume();
 
 ## Social Authentication and OIDC
```
raphaellondner-okta: I personally try to shy away from acronyms such as OIDC, which I don't see widely adopted. Can we stick to the longer but more widely accepted "OpenID Connect" and make sure we stick to this rule (or the other one with "OIDC"). I feel like "OpenID Connect" would also improve our SEO vs. "OIDC".

_source/_code/javascript/okta_auth_sdk.md
```diff
@@ -386,7 +386,7 @@ authClient.tx.resume();
...
 
-If you want to obtain an ID Token via [OIDC](/docs/api/resources/oidc.html) or [social authentication](/docs/guides/social_authentication.html), it's possible to manage the flow with the SDK. To do so, instantiate your client with a `clientId` and `redirectUri`.
+If you want to create a token via [OIDC](/docs/api/resources/oidc.html) or [social authentication](/docs/guides/social_authentication.html), it's possible to manage the flow with the SDK. To do so, instantiate your client with a `clientId` and `redirectUri`.
```
raphaellondner-okta: Where do the clientId and redirectUri come from? At this point, I would add point to our code samples, especially this one for a full implementation of the Auth SDK: https://github.com/oktadeveloper/okta-oauth-spa-authjs-osw

_source/_code/javascript/okta_auth_sdk.md
mystiberry-okta: API docs aren't consistent yet, I'll take an item to make it OpenID Connect. The only reason to use the abbreviation is if the noun phrases start to get too long, but I haven't seen any too-long noun phrases on brief review of this and api doc. Sorry for setting a bad example.

_source/_code/javascript/okta_auth_sdk.md
mystiberry-okta: What about "this SDK?" Seems weird to use the formal name this far into the doc.

_source/_code/javascript/okta_auth_sdk.md
```diff
@@ -396,46 +396,157 @@ var authClient = new OktaAuth({
...
+  scopes: [] // The scopes requested
+}
+
+// code
+{
+  authorizationCode: 'theAuthorizationCodeString'
+}
+~~~
+
+### Create token with a popup
```
raphaellondner-okta: "popup window" instead of just "popup"? Just wondering about the proper phrasing.

_source/_code/javascript/okta_auth_sdk.md
raphaellondner-okta: The "options" object is not explicit here, so it's difficult to figure out which flow you're using and want to demonstrate. It looks like it is the authorization code flow ('code') or the hybrid 'code id_token' flow but we have no way of knowing this here and understanding why we need to call the parseFromUri function. Just my 2 cents.

_source/_code/javascript/okta_auth_sdk.md
lboyette-okta: ack on "this SDK".
re: linking to sample - I'm not sure why linking to the code sample is better than linking to the docs that walk through how to set it up (which I guess is only available at https://github.com/oktadeveloper/okta-oauth-spa-authjs-osw/blob/master/Okta-OIDC_SPA_JS-OSW_DevSetupGuide.pdf). BTW, @mystiberry-okta, we probably want to incorporate the contents of that pdf into our support docs (or reference OIDC somewhere in the support docs).

_source/_code/javascript/okta_auth_sdk.md
lboyette-okta: I think I'm going stick with "popup" as "popup window" seems redundant

_source/_code/javascript/okta_auth_sdk.md
lboyette-okta: I'll add a note that the `parseFromUri` method is not necessary if only an authorization code is requested

_source/_code/javascript/okta_auth_sdk.md
```diff
@@ -396,46 +396,157 @@ var authClient = new OktaAuth({
...
+~~~
+
+Multiple tokens
+
 ~~~ javascript
-authClient.idToken.authorize()
-.then(function(res) {
-  // Raw ID Token
-  res.idToken;
+authClient.token.getWithPopup({
```
lboyette-okta: I don't think it's necessary, because we just trying to demonstrate how multiple tokens can be created, not exhaust the possible ways to create them.

mystiberry-okta:
I'm taking a stab at the conflict resolution now since Len is in meetings.

_source/_code/javascript/okta_auth_sdk.md
lboyette-okta: @mystiberry-okta looks like https://support.okta.com/help/articles/Knowledge_Article/Using-OpenID-Connect exists, so we should talk about modifying that guy

PR #421, opened 2016-10-03
removed lots of dead code and fixed anchor scroll (jquery 3 issue)

karlmcguinness-okta:
I switched to fixed height to match the header + page-content padding. 

The selector `a[href*=#]:not([href=#])` is not valid with recent jquery which caused the filed to fail to finish loading.


mystiberry-okta:
Yay! Should I test in my local environment before merging, or is it ready to go?

joelfranusic-okta:
LGTM :shipit:

PR #444, opened 2016-10-14
With Hector's changes

eklein-okta:
@cwu-okta @danielzeller-okta @vladdha-okta @mystiberry-okta Please check this PR.  As soon as it's okay, we can rebuild the site and send it out for a more general review.

### DIFF
--- a/_source/_docs/how-to/updating_saml_cert.md
+++ b/_source/_docs/how-to/updating_saml_cert.md
@@ -308,7 +308,7 @@ If you have OpenSSL installed, from the command line run:
 Where:
 
 `<your certificate>` is the certificate filename relative to the current directory. The certificate must be in PEM format. Use a plain text editor
-line Notepad or Textedit to save the certificate text from the `x5c` element returned from an API call, and add the **Begin Certificate** and **End Certificate** lines with the hyphens to the top and bottom of the file, as shown below. (The certificate shown below has been altered and is not valid.)
+like Notepad or Textedit to save the certificate text from the `x5c` element returned from an API call, and add the **Begin Certificate** and **End Certificate** lines with the hyphens to the top and bottom of the file, as shown below. (The certificate shown below has been altered and is not valid.)
 
 ~~~~
 -----BEGIN CERTIFICATE-----
--- a/_source/_docs/how-to/updating_saml_cert.md
+++ b/_source/_docs/how-to/updating_saml_cert.md
@@ -311,7 +311,7 @@ If you have OpenSSL installed, from the command line run:
 Where:
 
 `<your certificate>` is the certificate filename relative to the current directory. The certificate must be in PEM format. Use a plain text editor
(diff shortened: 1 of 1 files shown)


_source/_docs/how-to/updating_saml_cert.md
```diff
@@ -307,9 +307,34 @@ If you have OpenSSL installed, from the command line run:
 
 Where:
 
-`<your certificate>` is the certificate filename relative to the current directory.
-
-Use a free CSR and Certificate Decoder service and enter the contents of your certificate. These tools are readily found through a web search.
+`<your certificate>` is the certificate filename relative to the current directory. The certificate must be in PEM format. If you saved the
```
danielzeller-okta: At this point we have not called out how to save the certificate. It must be saved in plaintext with the Begin Certificate and End Certificate lines added. We should call out that they should use a plaintext editor like notepad/textedit. If they were to use something like Word/Openoffice openssl will not accept it.

_source/_docs/how-to/updating_saml_cert.md
```diff
@@ -307,9 +307,34 @@ If you have OpenSSL installed, from the command line run:
 
 Where:
 
-`<your certificate>` is the certificate filename relative to the current directory.
-
-Use a free CSR and Certificate Decoder service and enter the contents of your certificate. These tools are readily found through a web search.
+`<your certificate>` is the certificate filename relative to the current directory. The certificate must be in PEM format. If you saved the 
+certificate text from the `x5c` element returned from an API call, you must add the **Begin Certificate** and **End Certificate** lines with the 
+hyphens to the top and bottom of the file, as shown below. (The certificate shown below has been altered and is not valid.)
```
danielzeller-okta: Is there a specific reason that the cert's that we're showing in the document are invalid? To me this detracts from the authenticity of the document.

eklein-okta: As I understand, it's a temporary measure. But I'll leave that call to Cindy.

_source/_docs/how-to/updating_saml_cert.md
eklein-okta: I just added the plaintext editor part...

danielzeller-okta:
A couple of comments:

- In step one the admin is going to see "Use the _List Apps_ to return a list of all apps." where _List Apps_ is a link to the apps api. I'd suggest changing the link text to "List Apps Api" or just "Apps Api"
- There are no instructions on how to make the api calls. We need to either link to existing postman instructions or provide it inline.
- The request on step 3 where we are updating the app needs to be `PUT` instead of `POST`.
- The link in the last section, "Obtaining the Certificate for an App from a URL", is incorrect. The subdomin should be "\<theirOrgSubdomin\>-admin"

eklein-okta:
@danielzeller-okta Done!

cwu-okta:
When describing how to create a certificate to use with openssl we indicate to use a plain text editor and also add the begin/end cert tags. We should also let users know you cannot add trailing white spaces (space, carriage return, etc) at the end of the file. Adding a trailing carriage return (for example) will result in an invalid certificate which openssl will not be able to use.

eklein-okta:
@cwu-okta Done!

cwu-okta:
🚢

eklein-okta:
Yay!

PR #455, opened 2016-10-24
Table of Contents update: Phase 1

joelfranusic-okta:
Changes staged here: https://frozen-ridge-22271.herokuapp.com/

@mystiberry-okta do not merge until this has been peer reviewed.


mystiberry-okta:
1. In narrow (tablet/phone) width, the hamburger to open the left-nav disappears. I think John G. said this is a known issue being fixed in some other place or by some other person.

2. In desktop width, one topic shows odd alignment of the vertical progress bar, perhaps because it has more level 2 entries than the others. This doc is scheduled to be redesigned in the next two weeks, so if it's the only one with this problem, no worries.
odd_alignment

This same topic in a narrow (phone/tablet) width, fails to show the right-hand nav at the top, unless you click the back button once (it doesn't resize the window, but does display the right-hand nav at the top as expected). Weird! 

This topic isn't in the left-hand nav (for historical reasons). I couldn't replicate these behaviors in any other topic (I tried all the long ones with many H2 headings).

joelfranusic-okta:
Yeah, going to have to make some big changes here before we go live. Closing this PR for now.

PR #456, opened 2016-10-25
Fix google custom search bug

codyrobert:
A portion of the time when loading `/search/`, no results would be returned and an error would display in the console such as:

```
Uncaught TypeError: google.search.Search.apiaryXXX is not a function
```

The problem was the google custom search script was being loaded twice, via `master.js` and `search.js`. I've resolved this issue by removing the instance in `master.js`.

---

Thanks,
Cody Baker
Flickerbox, Inc.


johngronberg-okta:
@rchild-okta can you take a look?

assets/js/search.js
```diff
@@ -1,11 +1,17 @@
...
-	gcse.async = true;
-	gcse.src = (document.location.protocol == 'https:' ? 'https:' : 'http:') +
-	    '//cse.google.com/cse.js?cx=' + cx;
-	var s = document.getElementsByTagName('script')[0];
-	s.parentNode.insertBefore(gcse, s);
-    }
-    initSearch();
+(function($) {
+	
+	$(document).ready(function() {
```
rchild-okta: nit, just use:
```javascript
$(function () {
  // code
});
```

No need for the $(document).ready() part, and that pattern is deprecated in newer versions of jquery.

assets/js/search.js
```diff
@@ -1,11 +1,17 @@
...
+		gcse.type = 'text/javascript';
+		gcse.async = true;
+		gcse.src = document.location.protocol + '//cse.google.com/cse.js?cx=' + cx;
+	
+		var s = document.getElementsByTagName('script')[0];
+		s.parentNode.insertBefore(gcse, s);
+    
+    });
+
+})(jQuery);
```
rchild-okta: nit: newline

zdevil:
Hi all,

I see that the changes were approved.
Is there anything we can help with to make this live? The search is broken and we would like to fix that asap.

Thanks a lot.
Hamza.

johngronberg-okta:
@zdevil @codyrobert  we need a those couple of fixes that @rchild-okta mentioned. Then we can merge and deploy!

codyrobert:
@johngronberg-okta Thanks - formatting fixed.

Cody

PR #471, opened 2016-11-04
Tweaks to document order and re-direct changes

johngronberg-okta:
This PR contains the following navigational changes:

* on http://developer.okta.com/documentation/:
        - In the Use Cases section the order is now Authentication, MFA, Integrate with Okta (more logical order than before). 
        - In the reference section, Getting started now shows up first.
* Going to the API getting started page now directs to the API test client setup instead of the design principles page
* Sidebar is re-arranged to match the changes to the /documentation nav
* /docs re-directs to /documentation now instead of docs/api/getting_started/design_principles.html

@mystiberry-okta do these seem OK to you?

### DIFF
--- a/_source/_docs/api/getting_started/api_test_client.md
+++ b/_source/_docs/api/getting_started/api_test_client.md
@@ -1,5 +1,6 @@
 ---
 layout: docs_page
+weight: 1
 title: API Test Client
 redirect_from: "/docs/getting_started/api_test_client.html"
 ---
--- a/_source/_docs/api/getting_started/design_principles.md
+++ b/_source/_docs/api/getting_started/design_principles.md
@@ -1,5 +1,6 @@
 ---
 layout: docs_page
+weight: 4
 title: Design Principles
 redirect_from: "/docs/getting_started/design_principles.html"
 ---
--- a/_source/_docs/api/getting_started/enabling_cors.md
+++ b/_source/_docs/api/getting_started/enabling_cors.md
@@ -1,5 +1,6 @@
 ---
 layout: docs_page
+weight: 3
 title: Enabling CORS
 redirect_from: "/docs/getting_started/enabling_cors.html"
 scripts:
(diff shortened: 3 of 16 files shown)


mystiberry-okta:
How much testing did you do? Are we sure all the existing links to Getting Started page are compatible with the new topic? Not sure about the redirect change (because I don't understand how we implemented redirects very well)--should we test all existing links that point to /docs? I'll check the two navs as they aren't parallel, so not sure exactly what you changed. Monday at the earliest for checks.

I don't love that our getting started topic is pitching Postman hard and Okta not at all, would be happier if we made this change with a revision (just a sentence or two, at least) to that opening topic so the discussion doesn't start out so jarring. I can add something.

Full up today, will be full of RN tasks Monday, so Tuesday earliest.

johngronberg-okta:
* I tested where we were relying on /docs/api/getting_started/ as a re-direct, to make sure none of those were relying on that URL resolving to the design_principles.html page, and we aren't. Though, we do have some old blog posts with hardcoded links to design_principles.html, which should be using dynamic paths. 

* I did NOT add a new redirect_from value to /docs/api/getting_started/api_test_client.md. I am not sure if I need that? 

* I tested where we relied on the /docs basepath. Changing that to go to /documentation presents no issues. We weren't using it as a proxy for design_principles.html anywhere. It makes more sense that it goes to an index page instead of design principles. 

* I see what you mean about pushing Postman. It probably makes sense to hold off on changing the base page for "getting started with the Okta API" until we have the re-written getting started REQ complete. 

* The /documentation and sidebar changes were that I updated the order of 'Use Cases' on /documentation, and updated the sidebar to match (both now have an order of Authentication, MFA, Integrate with Okta)

mystiberry-okta:
Fantastic! Let's go ahead and merge this now, and I'll add a JIRA to fix that intro. It shouldn't take long, and I like the improvements overall. Thanks for being such a thorough tester.

PR #476, opened 2016-11-08
Added section related to updating the phone using 'updatePhone' option

srinivasanagandla-okta:
Documentation about enrolling SMS factor using an existing phone via "/api/v1/users/{userId}/factors" API. 

Covers the following scenarios: 
- *update the phone number*
- *use existing phone number*

Primary Reviewer: @mystiberry-okta 

Optional Reviewers: @karlmcguinness-okta @tthompson-okta 

Note: We have similar gap in authentication api for the SMS factor. I will create another PR for that.

### DIFF
--- a/_source/_docs/api/resources/factors.md
+++ b/_source/_docs/api/resources/factors.md
@@ -703,6 +703,17 @@ curl -v -X GET \
           ]
         }
       }
+    },
+    "_embedded": {
+      "phones": [
+        {
+          "id": "mblldntFJevYKbyQQ0g3",
+          "profile": {
+            "phoneNumber": "+14081234567"
+          },
+          "status": "ACTIVE"
+        }
+      ]
     }
   },
   {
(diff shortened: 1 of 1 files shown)


_source/_docs/api/resources/factors.md
```diff
@@ -736,6 +747,9 @@ curl -v -X GET \
 ]
 ~~~
 
+> Notice that the `sms` factor type includes an existing phone number in `_embedded`. You can either use the existing phone number or update it with a new number. See [Enroll Okta SMS factor](#enroll-okta-sms-factor) for more information.
```
srinivasanagandla-okta: @mystiberry-okta Can you help me with rephrasing this to clarify that "the phone number would be included in '_embedded' object only if there is one". I don't want developers to get the impression that this will "always" have the phone number.

mystiberry-okta: You mean you're worried that people will expect embedded to contain something like Phone number = "" ?   Because I think this is very clear, "includes an existing phone number" otherwise it doesn't include a phone number...?

srinivasanagandla-okta: If you think its clear then its good.

PR #485, opened 2016-11-14
Doc 104948

mystiberry-okta:
@federations-okta making scopes required in doc per Cindy's request, added labels for beta so you can tell OIDC from OAuth 2.0.


mystiberry-okta:
@federations-okta Can someone review this?

danielzeller-okta:
Why are we putting the beta tags around OpenId Connect? OIDC is EA.

mystiberry-okta:
Great question! I was hiding just the label "OpenID Connect" because if you aren't viewing the Beta, you don't need it (there's only one endpoint you can use). You only need the label if you're viewing the beta version.

danielzeller-okta:
Makes sense. LGTM 🍧

mystiberry-okta:
Merging now, will be published tomorrow.

PR #497, opened 2016-11-16
site refresh for 2016.46

mystiberry-okta:
@jpf or @johngronberg-okta do you want to review before I update the site?

### DIFF
--- a/feed.xml
+++ b/feed.xml
@@ -79,7 +79,7 @@
 
 &lt;p&gt;&lt;strong&gt;Happy Okta’ing!&lt;/strong&gt;&lt;/p&gt;
 </description>
-        <pubDate>Tue, 22 Mar 2016 00:00:00 -0700</pubDate>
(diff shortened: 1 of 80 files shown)


feed.xml
```diff
@@ -79,7 +79,7 @@
 
 &lt;p&gt;&lt;strong&gt;Happy Okta’ing!&lt;/strong&gt;&lt;/p&gt;
 </description>
-        <pubDate>Tue, 22 Mar 2016 00:00:00 -0700</pubDate>
+        <pubDate>Tue, 22 Mar 2016 00:00:00 +0000</pubDate>
```
johngronberg-okta: Why is this timezone changing?

johngronberg-okta:
Other than the timezone question, LGTM. We HAVE to fix those newlines that get inserted though...

feed.xml
mystiberry-okta: Safe to publish? The release is out on preview already.

mystiberry-okta: JF says go for it.

PR #499, opened 2016-11-18
table of oidc vs oauth2 endpoints

mystiberry-okta:
@cwu-okta and @federations-okta  team, Here's the table Cindy suggested showing OIDC and OAuth 2.0 endpoints. One problem in the last two rows, I can't link the normal way (markdown doesn't parse [{ correctly), I hope to figure it out tomorrow.

Meanwhile, please let me know if this is what people wanted and if you see any issues. Thanks!

### DIFF
--- a/_source/_docs/api/resources/oauth2.md
+++ b/_source/_docs/api/resources/oauth2.md
@@ -272,6 +272,20 @@ The actions in a Rule define which scopes can be granted to the requests, thus d
 
 ## Endpoints
 
+{% beta %}
+ Quick Reference
+
+| Desscription   | OpenID Connect | OAuth 2.0 |
+|:---|:---------------|:----------|
+|  [Authentication](#authentication-request)  | {% api_operation get /oauth2/v1/authorize %} {% api_lifecycle ea %}               | {% api_operation get /oauth2/:authorizationServerId/v1/authorize %} {% api_lifecycle beta %}          |
+|  [Token](#token-request) | {% api_operation post /oauth2/v1/token %} {% api_lifecycle ea %}               | {% api_operation post /oauth2/:authorizationServerId/v1/token %} {% api_lifecycle beta %}          |
+|  [Introspection Request](#introspection-request)  | {% api_operation post /oauth2/v1/introspect %} {% api_lifecycle ea %}               | {% api_operation post /oauth2/:authorizationServerId/v1/introspect %} {% api_lifecycle beta %}          |
+|  [Revocation Request](#revocation-request)  | {% api_operation post /oauth2/v1/revoke %} {% api_lifecycle ea %}               | {% api_operation post /oauth2/:authorizationServerId/v1/revoke %} {% api_lifecycle beta %}          |
+|  Get Keys  | {% api_operation get /oauth2/v1/keys %} (% api_lifecycle ea %}               |  {% api_operation get /oauth2/:authorizationServerId/v1/keys %} {% api_lifecycle beta %}         |
+|  Authorization Server Metadata Well-Known Discovery Document  | {% api_operation get /.well-known/openid-configuration %} {api_lifecycle ea %}               | {% api_operation get /oauth2/:authorizationServerId/.well-known/oauth-authorization-server %} {% api_lifecycle beta %}          |
+
+{% endbeta %}
+
 ### Authentication Request
 {:.api .api-operation}
 
--- a/_source/_docs/api/resources/oauth2.md
+++ b/_source/_docs/api/resources/oauth2.md
@@ -668,6 +668,7 @@ OpenID Connect:
 {% api_operation post /oauth2/v1/revoke %} {% api_lifecycle ea %}
 
 {% beta %}
+OAuth 2.0: 
 {% api_operation post /oauth2/:authorizationServerId/v1/revoke %} {% api_lifecycle beta %}
 {% endbeta %}
 
@@ -794,7 +795,7 @@ Standard open-source libraries are available for every major language to perform
 
 OpenID Connect: {% api_opertion get /.well-known/openid-configuration %} {% api_lifecycle ea %}
 
-Oauth 2.0: {% api_operation get /oauth2/:authorizationServerId/.well-known/oauth-authorization-server %} {% api_lifecycle beta %}
+OAuth 2.0: {% api_operation get /oauth2/:authorizationServerId/.well-known/oauth-authorization-server %} {% api_lifecycle beta %}
 
 This API endpoint returns metadata related to an Authorization Server that can be used by clients to programmatically configure their interactions with Okta.
 This API doesn't require any authentication and returns a JSON object with the following structure.
--- a/_source/_docs/api/resources/oauth2.md
+++ b/_source/_docs/api/resources/oauth2.md
@@ -272,20 +272,6 @@ The actions in a Rule define which scopes can be granted to the requests, thus d
 
 ## Endpoints
 
-{% beta %}
- Quick Reference
-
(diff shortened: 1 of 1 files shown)


_source/_docs/api/resources/oauth2.md
```diff
@@ -272,6 +272,20 @@ The actions in a Rule define which scopes can be granted to the requests, thus d
...
+|  [Token](#token-request) | {% api_operation post /oauth2/v1/token %} {% api_lifecycle ea %}               | {% api_operation post /oauth2/:authorizationServerId/v1/token %} {% api_lifecycle beta %}          |
+|  [Introspection Request](#introspection-request)  | {% api_operation post /oauth2/v1/introspect %} {% api_lifecycle ea %}               | {% api_operation post /oauth2/:authorizationServerId/v1/introspect %} {% api_lifecycle beta %}          |
+|  [Revocation Request](#revocation-request)  | {% api_operation post /oauth2/v1/revoke %} {% api_lifecycle ea %}               | {% api_operation post /oauth2/:authorizationServerId/v1/revoke %} {% api_lifecycle beta %}          |
+|  Get Keys  | {% api_operation get /oauth2/v1/keys %} (% api_lifecycle ea %}               |  {% api_operation get /oauth2/:authorizationServerId/v1/keys %} {% api_lifecycle beta %}         |
+|  Authorization Server Metadata Well-Known Discovery Document  | {% api_operation get /.well-known/openid-configuration %} {api_lifecycle ea %}               | {% api_operation get /oauth2/:authorizationServerId/.well-known/oauth-authorization-server %} {% api_lifecycle beta %}          |
```
larsjohansen-okta: We have a well-known discovery document for OpenID Connect at ```/oauth2/:authorizationServerId/.well-known/openid-configuration ``` as well, see my overall comment for details.

larsjohansen-okta:
This is good! I still think it's weird that we're not more explicit about the fact that you can use OAuth 2.0 endpoints for OpenID Connect as well. This table makes it look like you can only do OpenID Connect for the specified endpoints.

cwu-okta:
I did not mean to create a table; instead, I meant to weave the content into oauth2.md itself.

For example we have this 
image
I can't tell which is OIDC and which is OAuth. I would hope for something like

```
Token Request

OpenID Connect: /oauth2/v1/token [early access liquid tag]
OAuth: /oauth2/:authorizationServerId/v1/token [beta liquid tag]
```

At the top of oauth2 I would also like some introductory text explaining that OAuth and OIDC are very similar and one should follow the OIDC endpoints if they are following the OIDC use case and OAuth if they are following the OAuth use case. I suggest following up with Darko if you need help on the intro paragraph(s)

@tthompson-okta @yuliu-okta  fyi

mystiberry-okta:
@cwu-okta Got it, easy to fix--but what about Lars' point that you can use OAuth 2.0 endpoints for OIDC? Darko's working on the intro as we speak.

_source/_docs/api/resources/oauth2.md
mystiberry-okta: Got it!

mystiberry-okta: @larsjohansen-okta in oauth2.md, there is only /oauth2/:authorizationServerId/.well-known/oauth-authorization-server, and in oidc, there is only ./well-known/openid-configuration .  We haven't doc'd /oauth2/:authorizationServerId/.well-known/openid-configuration

??

mystiberry-okta:
@cwu-okta and @larsjohansen-okta Here are my updates. I don't have answers to Lars question (shouldn't we point out that you can use OAuth 2.0 endpoints for OIDC?) and no sign of the mystery endpoint Lars said exists: 

We have a well-known discovery document for OpenID Connect at /oauth2/:authorizationServerId/.well-known/openid-configuration as well, see my overall comment for details.

_source/_docs/api/resources/oauth2.md
```diff
@@ -776,7 +792,9 @@ Standard open-source libraries are available for every major language to perform
 ### Authorization Server Metadata
 {:.api .api-operation}
 
-{% api_operation get /oauth2/:authorizationServerId/.well-known/oauth-authorization-server %} {% api_lifecycle beta %}
+OpenID Connect: {% api_opertion get /.well-known/openid-configuration %} {% api_lifecycle ea %}
+
+Oauth 2.0: {% api_operation get /oauth2/:authorizationServerId/.well-known/oauth-authorization-server %} {% api_lifecycle beta %}
```
larsjohansen-okta: Oauth -> OAuth

_source/_docs/api/resources/oauth2.md
```diff
@@ -653,7 +662,11 @@ Content-Type: application/json;charset=UTF-8
 ### Revocation Request
 {:.api .api-operation}
 
+{% beta %}
+OpenID Connect: 
+{% endbeta %}
 {% api_operation post /oauth2/v1/revoke %} {% api_lifecycle ea %}
+
 {% beta %}
 {% api_operation post /oauth2/:authorizationServerId/v1/revoke %} {% api_lifecycle beta %}
```
larsjohansen-okta: Shouldn't we have ```OAuth 2.0:``` within this beta section too?

mystiberry-okta: Fixed!

mystiberry-okta:
Since Cindy is gone, can anyone in @federations-okta  answer these two questions, so I can ship this?

1. I don't have answers to Lars question (shouldn't we point out that you can use OAuth 2.0 endpoints for OIDC?) 

2. I can't find the mystery endpoint Lars said exists:

"We have a well-known discovery document for OpenID Connect at /oauth2/:authorizationServerId/.well-known/openid-configuration as well, see my overall comment for details."

_source/_docs/api/resources/oauth2.md
```diff
@@ -282,8 +282,7 @@ OpenID Connect:
 
 {% beta %}
 
-OAuth 2.0:
-
+OAuth 2.0:
```
yuliu-okta: @mystiberry-okta As the overview section says, OIDC, API Access Management and Social Auth are features built on top of OAuth 2.0. So I'd suggest we change the tag from OAuth 2.0 to API Access Management, it is really that feature is Beta, not all OAuth 2.0 API. And also we can add Social Auth in the authorization endpoint.

mystiberry-okta: I wish we had an Okta feature name for OIDC to make the naming truly parallel. Like this suggestion. I'll fix these tags, but wondering if we should make this change everywhere in our docs?

mcguinness:
I think is future revision it may be beneficial to break out the Organization Authorization Server and Custom Authorization Server.  OIDC only lives on the Org Authorization Server.   We don't support custom scopes for the Okta API be when we do they will be available on the Org Authorization Server.  Its confusing to see multiple endpoints that appear to do the same if you don't grok the different between Org and Custom Authorization Servers.    I think it may also be beneficial to introduce the term security domain and show the relationship of Resources Servers and Authorization Server as security domain.  The Org AS is bound by the org security domain and tokens issued by the Org AS should only be used by the client (id_token) or an protected resource for the Okta API where as Custom AS security domain is defined by the customer.

mystiberry-okta:
I'll create a JIRA for Karl's comments and role back the changes I made (except for the typo fixes) as I don't believe this is the right solution for the issue Cindy raised. Also a JIRA to fix the OAuth 2.0 vs. API Access Management. @federations-okta

mystiberry-okta:
Consensus is that this approach doesn't work. Created new JIRA to get this all worked out: https://oktainc.atlassian.net/browse/OKTA-107474

PR #507, opened 2016-11-23
SWA-only OAN apps special process

mystiberry-okta:
@vladdha-okta or @karanharibhakti-okta  can you verify this change is correct? I didn't include a "why" because I don't know the "why," but not sure you need it. Happy to add that, though, if you have it.

Also, please use the github review function by clicking the green "Review Changes" button instead of just adding comments. Okta wants to be able to find all reviews in the same place for compliance.


mystiberry-okta:
No response, so I'm merging this.

PR #512, opened 2016-11-29
how to configure your custom error page

mystiberry-okta:
@mattloughran-okta  can you review this using the review button (Vicky etc. asked us to do this).

I can tackle any edits tomorrow morning.


mystiberry-okta:
@larsjohansen-okta Can you look at this? Is this dialog really meant only for failed IWA logins? 
And how does the default IdP field work? Say they've got three IdPs defined, what would they choose for the Default IP? 

If you aren't the right developer to review this, can you let me know who is? Thanks!

mystiberry-okta:
@mattloughran-okta If this needs to be more generic (not IWA focused), do you have a place to put all the IWA-specific details? Wherever you link to this doc from?

mattloughran-okta:
What if the section you put in your doc has a bullet or paragraph that describes the IdP case and then below that a bullet or paragraph that describes the IWA login failure case? That's how I would approach it.

mystiberry-okta:
Trying to verify it applies to social auth at all. If not, this topic should be in the product docs.

mystiberry-okta:
@mattloughran-okta  as soon as you have what you need from here (if anything), let me know, and I'll close this. Note to self: close, don't merge.

PR #531, opened 2016-12-08
point users to OSW guide instead of reference

johngronberg-okta:
The existing link was re-directing users to the Sign In Widget Reference, when we really want them to go to the guide page.


mystiberry-okta:
Ship it.

PR #534, opened 2016-12-12
update key rotation doc

mystiberry-okta:
I also pulled the beta tags in this one. @yuliu-okta and @tthompson-okta please review. This must be in docs Wed 12/14.


mystiberry-okta:
Nevermind, this is a dupe of OKTA-97850, will close.

PR #550, opened 2016-12-16
Update Liquid template code with hack to remove whitespace 

joelfranusic-okta:
This is a "medium sized" update to remove whitespace from the HTML that Liquid template generates. The main issue here is that the current version of this site make extensive use of [Jekyll Collections](https://jekyllrb.com/docs/collections/) to generate pages and sidebars. 

This should be considered a temporary fix, until we can move back to generating sidebars and pages from [Jekyll Data Files](https://jekyllrb.com/docs/datafiles/), instead of collections. Meaning: Use a `config.yml` or similar kind of file to drive what shows up in the sidebar, which pages show up in the "Use Cases", "Reference", and "Code" sections, and so on.

### DIFF
--- a/_source/_includes/blog_post.html
+++ b/_source/_includes/blog_post.html
@@ -3,42 +3,33 @@
 	<article class="post-block">
 	  <header class="post-title-block">
 	    <h1><a href="{{ post.url }}">{{ post.title }}</a></h1>
-	    <div class="attribution">
-	      {% for post_author in post.author %}
-	      {% assign author = site.data.authors[post_author] %}
-	      {% if forloop.first %}
+	    <div class="attribution">{%
+          for post_author in post.author %}{% assign author = site.data.authors[post_author] %}{% if forloop.first %}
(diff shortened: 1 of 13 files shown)


joelfranusic-okta:
@johngronberg-okta and @mystiberry-okta for your review. 

Mysti, let's make sure we discuss the pros/cons of moving back to a `config.yml` type file and have a plan before we move forward with this PR.

joelfranusic-okta:
@mystiberry-okta I posted my view of the pros/cons of moving back to a `config.yml` file in our internal wiki page.

mystiberry-okta:
@joelfranusic-okta if it weren't for the white-space issue, I'd ask if we can fix how we're using collections--seems like there'd be solutions to the other issues in your excellent analysis. I read http://ben.balter.com/2015/02/20/jekyll-collections/ and wondered if there was metadata to control ordering? But with the white-space issue on top of everything, seems like moving back to YAML is a good idea.

PR #559, opened 2016-12-23
Add lint.sh with code to check documentation for URL consistency

joelfranusic-okta:
- Install ruby version found in .ruby-version
- Use rvm as a function instead of a command
- Install bundler again, since we'll have a new version of ruby

### DIFF
--- a/scripts/lint.sh
+++ b/scripts/lint.sh
@@ -1,0 +1,54 @@
+#!/bin/bash 
+# lint.sh
+#
+# "Linting is the process of running a program that will analyse code for potential errors."
+#
+# Author: Joël Franusic (joel.franusic@okta.com)
+# Copyright 2016 Okta, Inc.
+
+# `cd` to the path where Okta's build system has this repository
+cd ${OKTA_HOME}/${REPO}
+
+source "scripts/common.sh"
+
+function url_consistency_check() {
+    interject "Checking ${GENERATED_SITE_LOCATION} to make sure documentation does not include '/api/v1' in example URLs"
+    if [ ! -d "$GENERATED_SITE_LOCATION" ]; then
+       echo "Directory ${GENERATED_SITE_LOCATION} not found";
+       return 1;
+    fi
+
+    url_consistency_check_file=`mktemp`
+    # Search the _site directory for all files (-type f) ending in .html (-iname '*.html')
+    find $GENERATED_SITE_LOCATION -type f -iname '*.html' | \
+        # 'grep' all found files for 'api-uri-template', printing line numbers on output
+        xargs grep --line-number api-uri-template | \
+        # Search for the 'api/v' string, so we match "api/v1", "api/v2", etc
+        grep api/v | \
+        # The 'sed' command below pulls out the filename (\1), the line number (\2) and the URL path (\3)
+        # For example, this:
+        # _site/docs/api/resources/authn.html:2278:<p><span class="api-uri-template api-uri-post"><span class="api-label">POST</span> /api/v1/authn</span></p>
+        # becomes this:
+        # _site/docs/api/resources/authn.html:2278:/api/v1/authn
+        sed -e 's/^\([^:]*\):\([^:]*\).*<\/span> \(.*\)<\/span>.*/\1:\2:\3/' | \
+        # Write the results to STDOUT and the $url_consistency_check_file
+        tee $url_consistency_check_file
+    interject "Done checking $GENERATED_SITE_LOCATION for URLs with '/api/v1'"
+    # Return "True" if the file is empty
+    return `[ ! -s $url_consistency_check_file ]`
+}
+
+interject "Running lint.sh in $(pwd)"
+if ! generate_html;
+then
+    echo "Error building site"
+    exit $BUILD_FAILURE;
+fi
+
+if ! url_consistency_check;
+then
+    echo "FAILURE!"
+    exit $BUILD_FAILURE;
+fi
+
+exit $SUCCESS;


scripts/lint.sh
```diff
@@ -0,0 +1,94 @@
...
+cd ${OKTA_HOME}/${REPO}
+
+GENERATED_SITE_LOCATION="_site"
+
+# Print an easily visible line, useful for log files.
+function interject {
+    echo "----- ${1} -----"
+    }
+
+function setup {
```
joelfranusic-okta: This should be `setup()`

scripts/lint.sh
```diff
@@ -0,0 +1,94 @@
...
+}
+
+interject "Running lint.sh in $(pwd)"
+if ! setup;
+then
+    echo "Error building site"
+    exit $BUILD_FAILURE;
+fi
+
+if url_consistency_check;
```
joelfranusic-okta: The logic of `url_consistency_check` should be inverted so that this line reads `if ! url_consistency_check` as above.

scripts/lint.sh
```diff
@@ -0,0 +1,94 @@
+#!/bin/bash 
+# lint.sh
+#
+# "Linting is the process of running a program that will analyse code for potential errors."
+#
+# Author: Joël Franusic (joel.franusic@okta.com)
+# Copyright 2016 Okta, Inc.
+
+# Quit if any of the commands in this script fail
+set -e
```
jmaldonado-okta: In general, this is not a bad practice but we prefer adding error handlers since this script will be wrapped up by the bacon flow. Thus, it is encouraged by us to disable flag `-e` to avoid killing the script before sending the correct error code and logs to bacon. (The failure code is still raised by the error handler)

scripts/lint.sh
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+function setup {
+    interject 'Installing dependencies'
+    # Install the version of Ruby that we all use.
+    #
+    # Note: The file `.ruby-version` has the current ruby version and is used by rbenv
+    # https://rvm.io/workflow/projects#project-file-ruby-version
+    # For example, this file might contain this line: "ruby-2.0.0-p643"
+    rvm install `cat .ruby-version`
```
rlucero-okta: Double check with Eng-Prod, this shouldn't be required for CI machines.
@tsecor-okta @ajinkyasuryawanshi-okta @kelvinzhu-okta

scripts/lint.sh
```diff
@@ -0,0 +1,94 @@
...
+    # Install the gems needed for Jekyll
+    bundle install
+    interject 'Done installing dependencies'
+    interject 'Building Jekyll website'
+    # Build the Jekyll website
+    bundle exec jekyll build
+    local status=$?
+    interject 'Done building Jekyll website'
+    return $status
+    }
```
rlucero-okta: nit. Inconsistent bracket spacing.

scripts/lint.sh
```diff
@@ -0,0 +1,94 @@
...
+
+function url_consistency_check() {
+    interject "Checking ${GENERATED_SITE_LOCATION} to make sure documentation does not include '/api/v1' in example URLs"
+    if [ ! -d "$GENERATED_SITE_LOCATION" ]; then
+       echo "Directory ${GENERATED_SITE_LOCATION} not found";
+       return 1;
+    fi
+
+    url_consistency_check_file=`mktemp`
```
rlucero-okta: nit. Can this be a named file that can be pulled up and posted during CI runs?

scripts/lint.sh
```diff
@@ -0,0 +1,94 @@
...
+    interject "Done checking $GENERATED_SITE_LOCATION for URLs with '/api/v1'"
+    # Set the exit status to "1" if no lines were selected.
+    # From the man page for grep(1):
+    # EXIT STATUS
+    #     The grep utility exits with one of the following values:
+    # 
+    #     0     One or more lines were selected.
+    #     1     No lines were selected.
+    #     >1    An error occurred.
+    grep . $url_consistency_check_file > /dev/null
```
jmaldonado-okta: I am confused by what is the PASS criteria. I have a few comments but they depend on what the criteria is since is not clear to me what we are testing.

For example, in the current implementation there could be entries in file `url_consistency_check_file ` that will not necessarily match the regex in the `sed` above or even the file could contain only non-valid entries. Is this ok or are you just loooking for a non-empty file at the end? You probably want to print only matches via `-n` and raise if the regex does not match via `q`

scripts/lint.sh
jmaldonado-okta: Are you guys planning on using Albatross for development? It would be nice if you do so because that would also mean you would be aligned with CI when building locally then `setup()` should not be required for ruby since all docker containers already include a version.

scripts/lint.sh
```diff
@@ -0,0 +1,94 @@
...
+    # https://rvm.io/rvm/basics#post-install-configuration
+    source $(rvm `cat .ruby-version` do rvm env --path)
+    # Install bundler
+    gem install bundler
+    # Install the gems needed for Jekyll
+    bundle install
+    interject 'Done installing dependencies'
+    interject 'Building Jekyll website'
+    # Build the Jekyll website
+    bundle exec jekyll build
```
jmaldonado-okta: nit: This is the actual build. Thus, consider moving this part to a different function or perhaps a more descriptive name than `setup` ?

scripts/lint.sh
joelfranusic-okta: I just remembered, the reason why I use `tee` below is so that the contents of this file show up in the logs.

joelfranusic-okta:
@jmaldonado-okta and @rlucero-okta thanks for your detailed reviews. I've made changes based on your feedback and added comments where I didn't make any changes. Can you take another look at the code?

rchild-okta:
@joelfranusic-okta Which commits should I be looking at? 

FYI, usually we don't rebase until we're ready to merge (and it would be a rebase, vs. merging in all the latest commits which is what is happening here?). If you did want to pull in the latest changes, better to rebase off master and then push.

scripts/lint.sh
joelfranusic-okta: As far as the `find | xargs grep | grep | sed` command that I'm using. You mentioned the possibility of false positives. The likelihood of an `.html` file in the codebase containing the string `api-uri-template` and the string `api/v` are pretty rare. That said, the next version of this code will very likely want to use a real HTML parser, which would remove a lot of the risk of false positives.

joelfranusic-okta:
@jmaldonado-okta @rlucero-okta @rchild-okta - I should have addressed all of your comments. Additionally, I've squashed this topic branch down to one commit, how does it look to you now?

scripts/common.sh
```diff
@@ -0,0 +1,73 @@
...
+    echo "----- ${1} -----"
+}
+
+function check_for_protractor_dependencies() {
+    interject 'Checking Protractor dependencies'
+    command -v npm > /dev/null 2>&1 || { echo "This script requires 'npm', which is not installed"; exit 1; }
+    if ! command -v protractor > /dev/null 2>&1; then
+        interject 'Installing Protractor'
+        # http://www.protractortest.org/#/tutorial
+        npm install -g protractor
```
rchild-okta: We should install this locally as well, i.e. add it to package.json like this:
https://github.com/okta/samples-python-flask/blob/master/package.json#L33

You can then expose it via an npm run script like this:
https://github.com/okta/samples-python-flask/blob/master/package.json#L15

Which means that you would be able to run the command with `npm run test:e2e` (although in this case better to just name it "test")

In general, all npm dependencies should be listed in package.json, and we shouldn't be using some special mechanism to install it globally on ci/the user's system.

scripts/common.sh
```diff
@@ -0,0 +1,73 @@
...
+        interject 'Done installing Protractor'
+    fi
+    interject "Running 'webdriver-manager update'"
+    webdriver-manager update
```
rchild-okta: And in this case, webdriver-manager update would be handled by that test command as well, so you wouldn't need it here.

scripts/common.sh
```diff
@@ -0,0 +1,73 @@
...
+        bundle exec jekyll build
+        local status=$?
+        interject 'Done generating HTML'
+        return $status
+    else
+        interject 'HTML already generated'
+        return 0
+    fi
+}
+
```
rchild-okta: nit: don't need this extra line at the end

scripts/lint.sh
rchild-okta: Docs here:
https://oktawiki.atlassian.net/wiki/display/eng/Using+Bacon+to+build+your+artifact#UsingBacontobuildyourartifact-SettingUpAdditionalLogFiles

An example in the angular repo is here:
https://github.com/okta/samples-js-angular-1/blob/master/scripts/logs

scripts/common.sh
```diff
@@ -0,0 +1,73 @@
...
+    command -v npm > /dev/null 2>&1 || { echo "This script requires 'npm', which is not installed"; exit 1; }
+    if ! command -v protractor > /dev/null 2>&1; then
+        interject 'Installing Protractor'
+        # http://www.protractortest.org/#/tutorial
+        npm install -g protractor
+        interject 'Done installing Protractor'
+    fi
+    interject "Running 'webdriver-manager update'"
+    webdriver-manager update
+    interject 'Done checking Protractor dependencies'
```
rchild-okta: Also, was expecting to see a new package.json file (with jasmine-reporter from the last time I looked at this PR) + a command somewhere here to run `npm install`

rchild-okta:
lgtm - squash and 🚢

PR #564, opened 2016-12-27
add title to scim topic

mystiberry-okta:


mystiberry-okta:
adding title which was left out.

PR #578, opened 2017-01-07
updates for POST sessions

mystiberry-okta:
@tthompson-okta, @lboyette-okta , is this update accurate? I'm addressing Sohaib's request, not sure if my changes are enough or if we need more examples in the doc (Srini said we shouldn't be encouraging use of this api):

For session APIs, if we can mention the purpose of API call e.g. POST /sessions to create session id to be used for single logout later DELETE /sessions/sessionId. Any examples or flow of calls (e.g. Primary Auth followed by $
{orgname}
.okta.com/login/sessionCookieRedirect?token=<sessionToken>….) to establish session with Okta will be helpful
Sample to call CORS enabled /sessions/me endpoints from browser/JS http://developer.okta.com/docs/api/getting_started/enabling_cors.html

### DIFF
--- a/_source/_docs/api/resources/sessions.md
+++ b/_source/_docs/api/resources/sessions.md
@@ -35,15 +35,13 @@ Explore the Sessions API: [![Run in Postman](https://run.pstmn.io/button.svg)](h
 
 {% api_operation post /sessions %}
 
-> Don't use this API unless you need a session `id`. Instead, use the [Authentication API](./authn.html#primary-authentication) to obtain a `sessionToken`,
- or use one of the following flows to obtain a SSO session with a `sessionToken`:
+Creates a new session for a user with a valid session token. Use this API if, for example, you want to set the session cookie yourself instead of allowing Okta to set it, or want to hold the session ID in order to delete a session via the API instead of visiting the logout URL.
+
+> Don't use this API unless you need a session `id`. Instead, use one of the following flows to obtain a SSO session with a `sessionToken`:
 
 - [Retrieving a session cookie by visiting the OpenID Connect Authorization Endpoint](/docs/examples/session_cookie.html#retrieving-a-session-cookie-via-openid-connect-authorization-endpoint)
 - [Retrieving a session cookie by visiting a session redirect link](/docs/examples/session_cookie.html#retrieving-a-session-cookie-by-visiting-a-session-redirect-link)
 - [Retrieving a session cookie by visiting an application embed link](/docs/examples/session_cookie.html#retrieving-a-session-cookie-by-visiting-an-application-embed-link)
-
-Creates a new session for a user with a valid session token. Use this API if, for example, you want to set the session cookie yourself instead of
-allowing Okta to set it, or hold the session ID to delete a session via the API instead of visiting the logout URL.
 
 > This operation can be performed anonymously without an API Token.
 
@@ -54,8 +52,6 @@ Parameter        | Description                                                  
 ---------------- | ------------------------------------------------------------- | ---------- | ------------------------------- | -------- | -------
 additionalFields | Optional [session properties](#optional-session-properties)   | Query      | String (comma separated values) | FALSE    |
 sessionToken     | Session token obtained via [Authentication API](./authn.html) | Body       | String                          | TRUE     |
-
-> Creating a session with `username` and `password` is not recommended.  
 
 ##### Response Parameters
 {:.api .api-response .api-response-params}
--- a/_source/_docs/api/resources/sessions.md
+++ b/_source/_docs/api/resources/sessions.md
@@ -35,11 +35,15 @@ Explore the Sessions API: [![Run in Postman](https://run.pstmn.io/button.svg)](h
 
 {% api_operation post /sessions %}
 
-Creates a new session for a user with a valid session token. Only use this operation if you need the session `id`, otherwise you should use one of the following flows to obtain a SSO session with a `sessionToken`:
+> Don't use this API unless you need a session `id`. Instead, use the [Authentication API](./authn.html#primary-authentication) to obtain a `sessionToken`,
+ or use one of the following flows to obtain a SSO session with a `sessionToken`:
 
 - [Retrieving a session cookie by visiting the OpenID Connect Authorization Endpoint](/docs/examples/session_cookie.html#retrieving-a-session-cookie-via-openid-connect-authorization-endpoint)
 - [Retrieving a session cookie by visiting a session redirect link](/docs/examples/session_cookie.html#retrieving-a-session-cookie-by-visiting-a-session-redirect-link)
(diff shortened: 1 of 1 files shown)


mystiberry-okta:
@tthompson-okta would love your comments but not making you a required reviewer.

_source/_docs/api/resources/sessions.md
```diff
@@ -51,7 +55,7 @@ Parameter        | Description
 additionalFields | Optional [session properties](#optional-session-properties)   | Query      | String (comma separated values) | FALSE    |
 sessionToken     | Session token obtained via [Authentication API](./authn.html) | Body       | String                          | TRUE     |
 
-> Creating a session with `username` and `password` has been deprecated.  Use the [Authentication API](./authn.html) to obtain a `sessionToken`.
+> Creating a session with `username` and `password` is not recommended.
```
johngronberg-okta: Do we need this line at all here? Isn't this already in the page describing the preferred method?

_source/_docs/api/resources/sessions.md
```diff
@@ -35,12 +35,16 @@ Explore the Sessions API: [![Run in Postman](https://run.pstmn.io/button.svg)](h
...
+> Don't use this API unless you need a session `id`. Instead, use the [Authentication API](./authn.html#primary-authentication) to obtain a `sessionToken`,
+ or use one of the following flows to obtain a SSO session with a `sessionToken`:
 
 - [Retrieving a session cookie by visiting the OpenID Connect Authorization Endpoint](/docs/examples/session_cookie.html#retrieving-a-session-cookie-via-openid-connect-authorization-endpoint)
 - [Retrieving a session cookie by visiting a session redirect link](/docs/examples/session_cookie.html#retrieving-a-session-cookie-by-visiting-a-session-redirect-link)
 - [Retrieving a session cookie by visiting an application embed link](/docs/examples/session_cookie.html#retrieving-a-session-cookie-by-visiting-an-application-embed-link)
 
+Creates a new session for a user with a valid session token. Use this API if, for example, you want to set the session cookie yourself instead of
+allowing Okta to set it, or hold the session ID to delete a session via the API instead of visiting the logout URL.
```
tthompson-okta: Nit: seems like an odd line break here.

tthompson-okta:
It's unrelated to this PR, but I noticed the screenshot for configuring CORS is way out of date: http://developer.okta.com/docs/api/getting_started/enabling_cors.html#granting-cross-origin-access-to-websites

We should update this with the latest: 
screen shot 2017-01-08 at 4 13 13 pm

_source/_docs/api/resources/sessions.md
```diff
@@ -35,12 +35,16 @@ Explore the Sessions API: [![Run in Postman](https://run.pstmn.io/button.svg)](h
 
 {% api_operation post /sessions %}
 
-Creates a new session for a user with a valid session token. Only use this operation if you need the session `id`, otherwise you should use one of the following flows to obtain a SSO session with a `sessionToken`:
+> Don't use this API unless you need a session `id`. Instead, use the [Authentication API](./authn.html#primary-authentication) to obtain a `sessionToken`,
```
tthompson-okta: This guidance is not wrong, but it's also not contextually correct. Specifically, using the Auth API to obtain a sessionToken should actually feed into this API (notice one of the request params for this API is `sessionToken`); it's not an either/or.

I think you should keep everything after `otherwise you should use one of the following flows...` as it was.

_source/_docs/api/resources/sessions.md
```diff
@@ -35,12 +35,16 @@ Explore the Sessions API: [![Run in Postman](https://run.pstmn.io/button.svg)](h
...
+ or use one of the following flows to obtain a SSO session with a `sessionToken`:
 
 - [Retrieving a session cookie by visiting the OpenID Connect Authorization Endpoint](/docs/examples/session_cookie.html#retrieving-a-session-cookie-via-openid-connect-authorization-endpoint)
 - [Retrieving a session cookie by visiting a session redirect link](/docs/examples/session_cookie.html#retrieving-a-session-cookie-by-visiting-a-session-redirect-link)
 - [Retrieving a session cookie by visiting an application embed link](/docs/examples/session_cookie.html#retrieving-a-session-cookie-by-visiting-an-application-embed-link)
 
+Creates a new session for a user with a valid session token. Use this API if, for example, you want to set the session cookie yourself instead of
```
tthompson-okta: The flow of paragraphs feels a little backwards if you put this here. It seems like first we're saying "Don't use this flow unless..." and then down here were actually explaining what "this flow" is.

I would just take the second sentence and insert it into the existing structure up above (i.e. just add in `Use this API if you want to set the session cookie yourself instead of allowing Okta to set it, or if you need to delete the session via API instead of visiting the logout URL`.

_source/_docs/api/resources/sessions.md
mystiberry-okta: line breaks are ignored in the HTML, but I'll fix it.

_source/_docs/api/resources/sessions.md
mystiberry-okta: Done.

_source/_docs/api/resources/sessions.md
mystiberry-okta: Done.

_source/_docs/api/resources/sessions.md
mystiberry-okta: Done.

_source/_docs/api/resources/sessions.md
mystiberry-okta: Great suggestion, thanks!

_source/_docs/api/resources/sessions.md
mystiberry-okta: will do the CORS doc in a separate JIRA.

_source/_docs/api/resources/sessions.md
```diff
@@ -51,8 +53,6 @@ Parameter        | Description
 additionalFields | Optional [session properties](#optional-session-properties)   | Query      | String (comma separated values) | FALSE    |
 sessionToken     | Session token obtained via [Authentication API](./authn.html) | Body       | String                          | TRUE     |
 
-> Creating a session with `username` and `password` has been deprecated.  Use the [Authentication API](./authn.html) to obtain a `sessionToken`.
```
tthompson-okta: Hmm, this part we need to leave in place (in some shape or form). We need to strongly discourage people from creating a session with username/password. This API doesn't not support MFA. We provide a much richer alternative via the Auth API which does support MFA along with recovery flows and the whole nine yards.

johngronberg-okta: got it. Maybe we could add a line in here explaining that the Auth API offers much more functionality.

PR #610, opened 2017-01-27
OKTA-108631 factors in authn

mystiberry-okta:
@srinivasanagandla-okta and @johngronberg-okta Please review.


_source/_docs/api/resources/authn.md
```diff
@@ -1214,7 +1216,9 @@ curl -v -X POST \
 
 Enrolls a user with the Okta verify `push` factor. The factor must be [activated on the device](#activate-push-factor) by scanning the QR code or visiting the activation link sent via email or sms.
 
-> Use the published activation links to embed the QR code or distribute an activation `email` or `sms`
+Use the published activation links to embed the QR code or distribute an activation `email` or `sms`.
+
+> The `/api/v1/authn/factors` endpoint suppports only the Okta Verify and Google Authenticator factors at this time.
```
johngronberg-okta: Wouldn't this end point technically support any OTP app? For instance, I can setup Authy in our end user flow.

_source/_docs/api/resources/authn.md
```diff
@@ -2399,7 +2403,10 @@ curl -v -X POST \
 
 ##### Send Activation Links
 
-An activation email or sms can be sent when when the user is unable to scan the QR code.
+Sends an activation email or SMS when when the user is unable to scan the QR code provided as part of an Okta Verify transaction.
+If for any reason the user can't scan the QR code, they can use the link provided in email or SMS to complete the transaction.
+
+> The user must click the link from the same device as the one hosting the Okt Verify app.
```
johngronberg-okta: I don't think "hosting" is the right word here. Also- this says "Okt Verify". Should be "Okta Verify" 

What about "... the one with the Okta Verify app installed" ?

_source/_docs/api/resources/authn.md
```diff
@@ -3796,12 +3803,18 @@ Content-Type: application/json
...
 #### Forgot Password with Email Factor
 
-Starts a new password recovery transaction with a user identifier (`username`) and asynchronously sends a recovery email to the user's primary and secondary email address with a [recovery token](#recovery-token) that can be used to complete the transaction.
+Starts a new password recovery transaction for the email factor:
+ 
+* You must specify a user identifier (`username`) but no password in the request.
+* If the request is successful, Okta sends a recovery email asynchronously to the user's primary and secondary email address with a [recovery token](#recovery-token) so the user can complete the transaction.
 
-Since the recovery email is distributed out-of-band and may be viewed on a different user agent or device, this operation does not return a [state token](#state-token) and does not have a `next` link.
+Primary authentication of a user's recovery credential (e.g `EMAIL` or `SMS`) has not completed when this request is sent.
```
johngronberg-okta: Would it help to say something describing the state the user would be in at this point like "Recovery credential pending validation"?

_source/_docs/api/resources/authn.md
```diff
@@ -5098,9 +5122,17 @@ curl -v -X POST \
 
 {% api_operation post /api/v1/authn/skip %}
 
-Skips the current [transaction state](#transaction-state) and advances the state machine to the next state.
+Send a skip link to skip the current [transaction state](#transaction-state) and advance to the next state.
+
+When the user sees a skip link in the response, that means it's okay to skip that stage, clicks the skip link in the response.
```
johngronberg-okta: I don't understand the last bit of this statement "clicks the skip link in the response"

_source/_docs/api/resources/authn.md
```diff
@@ -5098,9 +5122,17 @@ curl -v -X POST \
...
+
+For example, after being warned that a password will soon expire, the user can skip the change password
```
johngronberg-okta: "... the user can skip the change password prompt by clicking a skip link"

_source/_docs/api/resources/authn.md
johngronberg-okta: @srinivasanagandla-okta This is probably a question for you.

karlmcguinness-okta: For `token:software:totp` factors, yes any OTP app implementing the TOTP algo should work.  We provide the QR code using param names that google authenticator or okta verify expects.  We provide the raw values in case you have a custom use case.

The gap is that we don't support branding so the Sign-In UI will display Google Authenticator or Okta Verify images and text.  We would need to add support for branding to make our TOTP usable for IT customers

push factor is Okta proprietary and won't work with anyone else

_source/_docs/api/resources/authn.md
karlmcguinness-okta: user should be developer.  This is just saying that response is hypermedia which publishes links dynamically based on status.  If response returns a skip link, then you can advance the state machine without completing the current state (such as changing your password)

_source/_docs/api/resources/authn.md
mystiberry-okta: @srinivasanagandla-okta can you respond? John's original question:

It's unclear which factors the Activate TOTP factor operation applies to. We should clarify. (I think John means which providers are available for TOTP factor? He mentioned these two for example:
Both the Okta Verify factor and Google Authenticator factor are TOTP factors.
e.g.,
http://developer.okta.com/docs/api/resources/authn.html#enroll-okta-verify-totp-factor
http://developer.okta.com/docs/api/resources/authn.html#enroll-google-authenticator-factor

I don't know how to address John's original question with the information provided here.

_source/_docs/api/resources/authn.md
mystiberry-okta: Done.

_source/_docs/api/resources/authn.md
mystiberry-okta: I'm a bit worried that there may be a few different states the user could be in. The point of the comment is that the primary authentication isn't finished until the user responds to the email. @srinivasanagandla-okta can you advise whether it's safe to say as John suggests?

_source/_docs/api/resources/authn.md
mystiberry-okta: Done.

_source/_docs/api/resources/authn.md
mystiberry-okta: Done.

_source/_docs/api/resources/authn.md
srinivasanagandla-okta: @johngronberg-okta Yes, the implementation is same. In that sense our TOTP factor endpoints work with any standard TOTP apps. As @karlmcguinness-okta mentioned the difference is the branding and messaging.

mystiberry-okta: I added this to the relevant endpoints (and one object table that talks about TOTP):

> This API implements [the TOTP standard](https://tools.ietf.org/html/rfc6238).
> This object implements [the TOTP standard](https://tools.ietf.org/html/rfc6238).

_source/_docs/api/resources/authn.md
mystiberry-okta: Done.

mystiberry-okta:
@srinivasanagandla-okta @johngronberg-okta and optionally @karlmcguinness-okta , I made the requested changes (I also cleaned up the tables and one section that had collapsed examples and non-example text). If you want to look at only the outstanding changes you requested, the line numbers are: 1124, 1840, 3817, 5133, 5619.

_source/_docs/api/resources/authn.md
```diff
@@ -5124,9 +5130,8 @@ curl -v -X POST \
 
 Send a skip link to skip the current [transaction state](#transaction-state) and advance to the next state.
 
-When the user sees a skip link in the response, that means it's okay to skip that stage, clicks the skip link in the response.
-
-For example, after being warned that a password will soon expire, the user can skip the change password
+If the response returns a skip link, then you can advance to the next state without completing the current state (such as changing your password).
```
srinivasanagandla-okta: nit: not sure using "you" is appropriate here.

mystiberry-okta: good catch. will fix.

PR #617, opened 2017-02-01
Add scrollTop offset for deep linking to IDs from other pages.

flickerbox-tom:
This is to fix linking directly to ID hashes from other pages or outside of the developer to account for the fixed header.

id under the fixed header

### DIFF
--- a/_source/assets/js/docs.js
+++ b/_source/assets/js/docs.js
@@ -331,9 +331,7 @@ $(function() {
 		onScroll();
 
 		if (window.location.hash) {
-			$body.animate({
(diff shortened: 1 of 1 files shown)


johngronberg-okta:
@flickerbox-tom the behavior is a bit odd when I test this locally, the page loads with the viewport a bit too far down and then the browser scrolls up to the correct location. Is there a way just to have this load directly to the proper location?

flickerbox-tom:
Updated the behavior to go to the ID without using .animate. Let me know if that works better.

mystiberry-okta:
@johngronberg-okta did you re-review?

PR #622, opened 2017-02-03
Describe groups claim

richardmateosian-okta:
This resolves OKTA-88192


mystiberry-okta:
@federations-okta Question: Richard's content mentions the Access Token, which isn't doc'd in this topic. The easy fix is to link to that section of oauth2.md, is that okay? Can someone on Fed review this?

PR #623, opened 2017-02-03
Changes for OKTA-113564

richardmateosian-okta:
Do we need to add anything in oidc.md?

### DIFF
--- a/_source/_docs/api/resources/apps.md
+++ b/_source/_docs/api/resources/apps.md
@@ -2620,7 +2620,9 @@ curl -v -X POST \
 
 <span class="api-uri-template api-uri-post"><span class="api-label">POST</span> /api/v1/apps/*:id*/lifecycle/deactivate</span>
 
-Deactivates an active application.
+Deactivate an active application. Invalidate all tokens issued by that application
+immediately and permanently. Reactivating the application does not restore the
+validity of those tokens.
 
 ##### Request Parameters
 {:.api .api-request .api-request-params}
--- a/_source/_docs/api/resources/oauth2.md
+++ b/_source/_docs/api/resources/oauth2.md
@@ -86,7 +86,9 @@ This authentication method produces a `sessionToken` which can be passed into th
 
 An Access Token is a [JSON web token (JWT)](https://tools.ietf.org/html/rfc7519) encoded in base64URL format that contains [a header](#jwt-header), [payload](#jwt-payload), and [signature](#jwt-signature). A resource server can authorize the client to access particular resources based on the [scopes and claims](#scopes-and-claims) in the Access Token.
 
-The lifetime of Access Token can be configured in the [Access Policies](#access-policies).
+The lifetime of Access Token can be configured in the [Access Policies](#access-policies). If the application that
+issued the token is deactivated, the token is immediately and permanently invalidated. Reactivating the application
+does not make the token valid again.
 
 ### JWT Header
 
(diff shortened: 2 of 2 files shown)


mystiberry-okta:
@federations-okta , do we also need to add a comment in oidc.md?

_source/_docs/api/resources/apps.md
```diff
@@ -2620,7 +2620,9 @@ curl -v -X POST \
 
 <span class="api-uri-template api-uri-post"><span class="api-label">POST</span> /api/v1/apps/*:id*/lifecycle/deactivate</span>
 
-Deactivates an active application.
+Deactivate an active application. Invalidate all tokens issued by that application
```
cwu-okta: This is the general apps api so discussion of tokens doesn't make any sense here. This content should go elsewhere, ie OIDC Clients API or where we talk about tokens.

richardmateosian-okta: @cwu-okta I understand that point. On the other hand, invalidating tokens is a direct effect of invoking the deactivation API. Wouldn't people contemplating deactivating an app want to know that before proceeding? 

Perhaps we need information in both places.

cwu-okta: We should probably mention this in the clients API, minimally.  For now I would probably not include this information in the public apps API. We don't have precedence for how to handle documenting what happens for specific apps in the general apps API. In this case, only one app takes specific action (different than the rest of the apps) when this endpoint is called. I would suggest creating a follow up Jira for working through how to deal with specific actions by an app. This might not be the first example where specific action is taken for only one app.

richardmateosian-okta: @cwu-okta I may be confused about the difference between clients and apps, so I'll talk with Darko to see if I can get clear on the difference.

richardmateosian-okta: @cwu-okta Would it be sufficient to back out the changes to apps.md, leave the changes to oauth2.md, and add similar changes to oidc.md?

mystiberry-okta: FYI, the convention is to use "Deactivates" not "Deactivate" for the sentence fragment that briefly explains the endpoint.

cwu-okta: @richardmateosian-okta The changes should be backed out of apps.md. Instead of adding the changes to oidc.md I would add it to oauth-clients.md, which is the CRUD for oauth clients API

richardmateosian-okta: @cwu-okta Just to be sure I understand, you're saying
 * Back changes out of apps.md 
 * Leave changes to oauth2.md as they are in this PR
 * Add similar changes to oauth-clients.md

richardmateosian-okta: @mystiberry-okta It's not just that I prefer the imperative here, but I swear I saw an Okta engineering document that says to do it that way. Unfortunately, I can't find it.  :-)

cwu-okta: @richardmateosian-okta correct, that is my suggestion.

mystiberry-okta: @richardmateosian-okta Imperative is really not linguistically correct here, as we're telling the reader what the resource does in  phrase,. We're not telling the user to do the thing that the resource does in an imperative sentence. One could argue that the imperative represents a deleted/unspoken phrase like "With this resource, you can deactive an app." But I would argue the opposite, that the phrase is really "my/cool/resource" deactivates clients (except that sometimes we don't show the resource first).  So, please follow the convention currently being used for now.  If you firmly believe there's a linguistic reason for going with imperative, file a JIRA please. We'd want to change all resource descriptions if we adopt this convention.

PR #627, opened 2017-02-07
Fix typo

mincong-h:


mystiberry-okta:
Reman showed me about switching bases. This is a three-character plain text file change.

PR #634, opened 2017-02-13
OKTA-113564: Describe the fate of tokens issued by subsequently deact…

richardmateosian-okta:
…ivated clients.


cwu-okta:
:shipit:

PR #635, opened 2017-02-14
Add support for transclusion via import_external_markdown()

joelfranusic-okta:
The `import_external_markdown()` function appends files from external repositories into Markdown files in the _source directory that use the `external_source:` frontmatter option.

The plan is to start including documentation from projects like `okta/okta-signin-widget` and `okta/okta-auth-js` into developer.okta.com by including the npm packages for those projects into the `package.json` file in this repository, then we can set up transclusion in files in this repository via the `external_source:` frontmatter option.

### DIFF
--- a/scripts/common.sh
+++ b/scripts/common.sh
@@ -5,12 +5,14 @@
 # Where the generated Jekyll site will be placed
 GENERATED_SITE_LOCATION="_site"
 
+# Define these ENV vars if they aren't defined already,
+# so these scripts can be run outside of CI
 if [[ -z "${BUILD_FAILURE}" ]]; then
     export BUILD_FAILURE=1
 fi
 
 if [[ -z "${SUCCESS}" ]]; then
-    export SUCCESS=1
+    export SUCCESS=0
 fi
 
 source "scripts/import_external_markdown.sh"
--- a/scripts/common.sh
+++ b/scripts/common.sh
@@ -5,6 +5,14 @@
 # Where the generated Jekyll site will be placed
 GENERATED_SITE_LOCATION="_site"
 
+if [[ -z "${BUILD_FAILURE}" ]]; then
+    export BUILD_FAILURE=1
+fi
+
+if [[ -z "${SUCCESS}" ]]; then
+    export SUCCESS=1
+fi
+
 # Print an easily visible line, useful for log files.
 function interject() {
     echo "----- ${1} -----"
@@ -65,5 +73,11 @@ function generate_html() {
     fi
 }
 
-
-
+function require_env_var() {
+    local env_var_name=$1
+    eval env_var=\$$env_var_name
+    if [[ -z "${env_var}" ]]; then
+        echo "Environment variable '${env_var_name}' must be defined, but isn't.";
+        exit 1
+    fi
+}
--- a/scripts/common.sh
+++ b/scripts/common.sh
@@ -13,16 +13,18 @@ if [[ -z "${SUCCESS}" ]]; then
     export SUCCESS=1
 fi
 
+source "scripts/import_external_markdown.sh"
+
 # Print an easily visible line, useful for log files.
 function interject() {
     echo "----- ${1} -----"
 }
 
-function check_for_protractor_dependencies() {
-    interject 'Checking Protractor dependencies'
+function check_for_npm_dependencies() {
+    interject 'Checking NPM dependencies'
     command -v npm > /dev/null 2>&1 || { echo "This script requires 'npm', which is not installed"; exit 1; }
     npm install --only=dev
-    interject 'Done checking Protractor dependencies'
+    interject 'Done checking NPM dependencies'
 }
 
 function check_for_jekyll_dependencies() {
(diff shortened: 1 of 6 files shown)


scripts/common.sh
```diff
@@ -5,16 +5,26 @@
 # Where the generated Jekyll site will be placed
 GENERATED_SITE_LOCATION="_site"
 
+if [[ -z "${BUILD_FAILURE}" ]]; then
+    export BUILD_FAILURE=1
+fi
+
+if [[ -z "${SUCCESS}" ]]; then
+    export SUCCESS=1
```
rchild-okta: Should this be 0? I.e. seems like build.sh will return a non-zero exit code if we are not on the deploy branch, failing the test suite.

scripts/import_external_markdown.sh
```diff
@@ -0,0 +1,63 @@
+#!/bin/bash
+# include_external_markdown.sh
```
rchild-okta: nit: **import** not include

scripts/import_external_markdown.sh
```diff
@@ -0,0 +1,63 @@
...
+#     title: Testing
+#     external_source: /tmp/contents.md
+#     ----
+#
+#   And "/tmp/contents.md" contained this:
+#
+#     Hello World!
+#
+#   Then here is what "/tmp/hello.md" would contain after
+#   running the import_external_documentation() function:
```
rchild-okta: import_external_markdown

scripts/import_external_markdown.sh
```diff
@@ -0,0 +1,63 @@
...
+#
+#     ----
+#     title: Testing
+#     external_source: /tmp/contents.md
+#     ----
+#     <!-- Imported from external_source file -->
```
rchild-okta: General: We should add a selenium test that visits the new sign-in-widget & authjs pages and verifies that it has been built correctly (i.e. check for some string that is coming from the readme's). My general thought is that we're adding code, but I see no tests to validate this code.

It seems like we're not actually including them yet - is there a reason for that? I think it'd be nice to start using it here, and perhaps preface the title with "devOnly" or something like that.

scripts/import_external_markdown.sh
```diff
@@ -0,0 +1,63 @@
...
+
+
+# This is the string written into transcluded files, to makes this script idempotent 
+TRANSCLUDE_WAS_HERE="<!-- Imported from external_source file -->"
+# This is the key used in the front matter of Markdown files, used to point to the file to be transcluded
+EXTERNAL_SOURCE="external_source"
+
+# Given a filename, read the filename of a file to transclude
+function transclude_file() {
+    filename=$1
```
rchild-okta: How will this work i.e. the frontmatter file isn't going to know the absolute path to the file is it? Should this be relative to this directory, or even more specific to node_modules? In general I am also wondering if we want to limit where we're pulling things from. For example, is there ever a case where I can run this against something like Travis and pull in file information for something I shouldn't?

scripts/common.sh
joelfranusic-okta: Oh, good catch. Thanks.

scripts/import_external_markdown.sh
joelfranusic-okta: Regarding tests to validate this code, we've got several options for this code (in order of preference):

1. [basht](https://github.com/progrium/basht)

joelfranusic-okta: > It seems like we're not actually including them yet - is there a reason for that? 

Only reason for that is that the stylistic changes needed in okta-auth-js haven't been included in a recent NPM package.

>  We should add a selenium test that visits the new sign-in-widget & authjs pages and verifies that it has been built correctly

Do you want that now, or just take a note to do this later?

scripts/import_external_markdown.sh
joelfranusic-okta: This is a good point, I propose that we do the following:

- Scope relative to the directory, the root of the okta.github.io repo.
- Make a path whitelist for transcluded files, with `node_modules` as the only item currently on the whitelist.

Here is what the Markdown frontmatter would look like under this proposal:
```
----
title: Example
external_source: node_modules/\@okta/okta-auth-js/README.md
----
```

If the path in `external_source` did not start with `node_modules/` the script would give an error on the order of "The path for ${filename} is not whitelisted by the import_external_markdown script"

Thoughts?

scripts/import_external_markdown.sh
joelfranusic-okta: Here is the commit for the Selenium tests I added after our conversation: 9d22844974d33f63894ab41c4eca83046aeabeaa - how do these look?

scripts/import_external_markdown.sh
joelfranusic-okta: Implemented here: 017474e60b058083027635baba51ae3ce77c871f

joelfranusic-okta:
@rchild-okta could you take another look at the updates I made based on your PR, I added more code than just the basic fixes you asked for.

PR #639, opened 2017-02-17
Added missing parenthesis to the javascript example

mlaursen:


mystiberry-okta:
@lboyette-okta or @rchild-okta can you confirm this small change is correct? Thanks!

PR #644, opened 2017-02-23
correct steps in saml example app topic

mystiberry-okta:


mystiberry-okta:
Corrected content to match UI.

PR #692, opened 2017-03-24
better URL

mystiberry-okta:


mystiberry-okta:
@jpf I created a static redirect, is that the current process?

PR #711, opened 2017-03-30
React-Sign-In-Widget Post from Lee

leebrandt:
Let me know if there is anything else I need to do.


jmelberg-okta:
Tested in Bacon with PR: [#712](https://github.com/okta/okta.github.io/pull/712). We can probably close this PR in favor or [#712](https://github.com/okta/okta.github.io/pull/712)

PR #716, opened 2017-04-03
Fixed small typo and fixed the last header getting merged into quote about it.

leebrandt:


mystiberry-okta:
@rchild-okta I can't see the last two commits in bacon. How can we test?

PR #253, opened 2016-05-26
Attempts to remove the last group target from a UserAdmin role now re…

brianzuzga-okta:
…sult in an error.

API doc update that goes with the code change.

Relates-to: OKTA-90514

PR #729, opened 2017-04-10
removing last app target now throws an exception

yuanxue-okta:
this is a similar change to #253 

@mystiberry-okta 
@alifatolahi-okta 
@brianzuzga-okta

### DIFF
--- a/_source/_docs/api/resources/roles.md
+++ b/_source/_docs/api/resources/roles.md
@@ -480,7 +480,7 @@ HTTP/1.1 204 No Content
 
 Removes an app target from an `APP_ADMIN` role assignment.
 
-> Removing the last app target changes the scope of the role assignment from only applying to specific app targets to applying to **all** app targets.
+> Don't remove the last app target from a role assignment, as this causes an exception.  If you need a role assignment that applies to all apps, the API consumer should delete the `APP_ADMIN` role assignment and recreate it.
 
 ##### Request Parameters
 {:.api .api-request .api-request-params}


johngronberg-okta:
tagging @jakubvul for review since Mysti is out this week

_source/_docs/api/resources/roles.md
```diff
@@ -480,7 +480,7 @@ HTTP/1.1 204 No Content
 
 Removes an app target from an `APP_ADMIN` role assignment.
 
-> Removing the last app target changes the scope of the role assignment from only applying to specific app targets to applying to **all** app targets.
+> Don't remove the last app target from a role assignment, as this causes an exception.  If you need a role assignment that applies to all apps, the API consumer should delete the `APP_ADMIN` role assignment and recreate it.
```
alifatolahi-okta: "If you need a role assignment that applies to all apps, the API consumer should delete" this is a bit confusing. I'd change to something along the line of "If a role assignment is required to apply to all apps, the API consumer should delete"

yuanxue-okta: this is the exact same wording as user admin's group target removal

jakubvul: @alifatolahi-okta Isn't the correct term here "target"? That is, "If a role assignment should target all apps".

alifatolahi-okta: yeah that makes sense, i.e. "If a role assignment should target all apps, the API consumer should delete".

@yuanxue-okta I find it confusing as it says "if you ..." and then assert "API consumer should ..." it's not clear whether "you" and "API consumer" are the same or different people. I'd consult with someone who has more experience and is a better domain expert; then decide as they advise.

yuanxue-okta: changed to "If a role assignment should target all apps", ill loop in @brianzuzga-okta  and get his opinion as well

jakubvul: I would rewrite this as follows:

"Removing the last app target will cause an exception. If a role assignment should target all apps, then you should delete the `APP_ADMIN` role assignment and recreate it."

However, it is not obvious what the connection between these two sentences is. They only make sense if you know that previously the documentation mentioned removing the last app target as a way of creating a role assignment that targets all apps. So right now it looks like we have two unrelated pieces of information here. Which is fine, but we should treat them that way.

Also, how do you create an `APP_ADMIN` role that targets all apps?

yuanxue-okta: changed to the suggested wording.

all new APP_ADMIN will by default targets all apps, you can create an APP_ADMIN through an api POST call to  /api/v1/users/:uid/roles with a body of 
```
{
      "type": "APP_ADMIN"
}
```

yuanxue-okta:
@alifatolahi-okta @jakubvul @brianzuzga-okta 
do you guys have any other concerns with the wording?

_source/_docs/api/resources/roles.md
```diff
@@ -480,7 +480,7 @@ HTTP/1.1 204 No Content
 
 Removes an app target from an `APP_ADMIN` role assignment.
 
-> Removing the last app target changes the scope of the role assignment from only applying to specific app targets to applying to **all** app targets.
+> Removing the last app target will cause an exception. If a role assignment should target all apps, then you should delete the `APP_ADMIN` role assignment and recreate it.
```
brianzuzga-okta: You may want to mirror the language that we have for group scoping of the UserAdmin role.

> Don’t remove the last group target from a role assignment, as this causes an exception. If you need a role assignment that applies to all groups, the API consumer should delete the USER_ADMIN role assignment and recreate it.

jakubvul: Looks good to me!

PR #752, opened 2017-04-20
Fix blog layout so its responsive

mraible:
Code listings with long lines causes the blog to be a fixed width and not squish when the browser is squished. This fixes that problem. Tested in Chrome and Firefox on Mac.

Before:

screen shot 2017-04-20 at 3 22 48 pm

After:

screen shot 2017-04-20 at 3 23 25 pm

### DIFF
--- a/_source/_sass/okta/components/_PageContent.scss
+++ b/_source/_sass/okta/components/_PageContent.scss
@@ -186,7 +186,7 @@
 			overflow: auto;
 			code {
 				white-space: pre;
-				// fix wide cost listings that don't overflow automatically
+				// fix wide code listings that don't overflow automatically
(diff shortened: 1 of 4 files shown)


_source/_sass/okta/components/_PageContent.scss
```diff
@@ -186,6 +186,9 @@
 			overflow: auto;
 			code {
 				white-space: pre;
+				// fix wide cost listings that don't overflow automatically
```
mystiberry-okta: @mraible Did you mean to say "fix wide cost listings" or "fix wide code listings" ?

mraible: @mystiberry-okta I meant to say "fix wide code listings". Good catch! Will fix.

PR #741, opened 2017-04-17
Fix individual blog layout

mraible:
This makes it so individual posts don't stretch to fit the page. It also fixes syntax highlighting on individual posts.

Before:

before

After:

after

PR #757, opened 2017-04-21
Make blog_index more like blog_post

mraible:
This reverts my previous changes:

- https://github.com/okta/okta.github.io/pull/741
- https://github.com/okta/okta.github.io/pull/752

Rather than changing `blog_post.html` to use `class="PageContent"`, it seems easier to make `blog_index.html` use styles associated with `blog_post.html` (because they're responsive and stretch with the page). It also minimizes the impact on `.PageContent` since that class seems to pertain to the rest of the site.

### DIFF
--- a/_source/assets/css/page-blog.scss
+++ b/_source/assets/css/page-blog.scss
@@ -10,13 +10,13 @@
 ;
 
 
-#main-container.blog {
-  width: 800px;
+.main-container > .blog {
+  max-width: 1360px; // result of calculation in _PageContent.scss
   margin: 0 auto;
 }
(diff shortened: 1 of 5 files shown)


rchild-okta:
Yep, @lboyette-okta

_source/assets/css/page-blog.scss
```diff
@@ -10,13 +10,13 @@
 ;
 
 
-#main-container.blog {
```
lboyette-okta: hmm, guess this selector wasn't matching anything before.

PR #758, opened 2017-04-21
simple hal links heads up

mystiberry-okta:
I word-smithed Ed's excellent write-up. @edjohns-okta and @dbebout-okta please review.


mystiberry-okta:
oops, wrong content. ignore.

PR #520, opened 2016-12-06
API doc for U2F factor enroll and challenge in authentication API

srinivasanagandla-okta:
- *Added section for enroll and verify API details for U2F factor*
- *Also fixed references to rain.okta1.com and changed them to your-domain.okta.com*

Please review @mystiberry-okta @tthompson-okta

PR #761, opened 2017-04-21
override settings in apps api

mystiberry-okta:
@annaluo-okta , @tthompson-okta , @swastimishra-okta , please review this doc for OKTA-123576.

1. I added a note explaining when to use the overrides which is too generic. Please provide additional explanation of when someone should use these. 

2. I guessed Nillable=True and Unique=False for all four, and that the datatype is string.

3. Can I use trevor's example from the JIRA as the request example?

### DIFF
--- a/_source/_docs/api/resources/apps.md
+++ b/_source/_docs/api/resources/apps.md
@@ -699,11 +699,15 @@ Parameter             | Description                                             
 --------------------- | ----------------------------------------------------------------------------------------------------------------- | ---------------------------------------------------- | -------- | ----- | ----------------------------------------
 defaultRelayState     | Identifies a specific application resource in an IDP initiated SSO scenario.                                      | String                                               | TRUE     | FALSE |
 ssoAcsUrl             | Single Sign On Url                                                                                                | String                                               | FALSE    | FALSE |  [URL](http://tools.ietf.org/html/rfc3986)
+ssoAcsUrlOverride     | Overrides the `ssoAcsUrl` setting                                                                                 | String                                               | TRUE     | FALSE |  [URL](http://tools.ietf.org/html/rfc3986)
 recipient             | The location where the app may present the SAML assertion                                                         | String                                               | FALSE    | FALSE |  [URL](http://tools.ietf.org/html/rfc3986)
+recipientOverride     | Overrides the `recipient` setting                                                                                 | String                                               | TRUE     | FALSE |  [URL](http://tools.ietf.org/html/rfc3986)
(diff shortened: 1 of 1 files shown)


_source/_docs/api/resources/apps.md
```diff
@@ -713,9 +717,9 @@ honorForceAuthn       | Prompt user to re-authenticate if SP asks for it
 authnContextClassRef  | Identifies the SAML authentication context class for the assertion's authentication statement                     | String                                               | FALSE    | FALSE |
 attributeStatements   | Check [here](http://docs.oasis-open.org/security/saml/Post2.0/sstc-saml-tech-overview-2.0-cd-02.html) for details | [Attribute Statements](#attribute-statements-object) | FALSE    | FALSE |
 
-> Fields that require certificate uploads can't be enabled through the API, such as Single Log Out and Assertion Encryption. These must be updated through the UI.
-
-> Either (or both) "responseSigned" or "assertionSigned" must be TRUE.
+* Fields that require certificate uploads can't be enabled through the API, such as Single Log Out and Assertion Encryption. These must be updated through the UI.
+* Either (or both) "responseSigned" or "assertionSigned" must be TRUE.
+* Use the override settings (`ssoAcsUrlOverride`, `recipientOverride`, `destinationOverride`, or `audienceOverride` if the corresponding fields are causing errors.
```
mystiberry-okta: This is the sentence that needs some improvement!

annaluo-okta:
@swastimishra-okta @tthompson-okta FYI, this is for updating doc http://developer.okta.com/docs/api/resources/apps.html .

@mystiberry-okta , few suggestions -
1. ssoAcurlOverride, recipientOverride and destinationOverride Validation should be URL. All Nillable and Unique looks good to me.

2. There are different scenarios, if we can make clear in the doc -
 - In SAML 1.1, like O365 app, destinationOverride is not available.
 - In SAML 2.0, like Box app, all 4 override attributes are available.
 - In App Wizard SAML App, no override attribute available.
 * You can see examples from the second set of screenshots from https://github.com/okta/okta-core/pull/19317 .

3. For the description, I'd like to add something similar as we have in AppInstance.java #520 - https://github.com/okta/okta-core/blob/ba796a0a5ddb21f669213c5e07b8ee56839a72b6/client/api/src/main/java/com/okta/client/api/admin/dto/AppInstance.java#L520 .

"These override properties were added to provide an alternative way of persisting post back and similar other urls. One example we currently have is where ssoAcsUrlOverride is being used is to support cloud access security broker(CASB) use case for Office365 app instances."

PR #772, opened 2017-04-26
Doc login_hint for oidc/oauth2

royalchan-okta:
@federations-okta 

https://github.com/okta/okta-core/pull/19965


_source/_docs/api/resources/oauth2.md
```diff
@@ -349,6 +349,7 @@ prompt         | Can be either ``none`` or ``login``. The value determines if Ok
 nonce          | Specifies a nonce that is reflected back in the ID Token. It is used to mitigate replay attacks. | Query        | String   | TRUE     |
 code_challenge | Specifies a challenge of [PKCE](#parameter-details). The challenge is verified in the Access Token request.  | Query        | String   | FALSE    |
 code_challenge_method | Specifies the method that was used to derive the code challenge. Only S256 is supported.  | Query        | String   | FALSE    |
+login_hint | A username to prepopulate if prompting for authentication.  | Query        | String   | FALSE    |
```
yuliu-okta: Do we need to mentioned we will ignore the session if it does not match the login_hint?

royalchan-okta: It's actually not quite the whole truth, we will ignore the session if prompt=login AND login_hint doesn't match. I'm hesitant to be super explicit about this behavior, but maybe @karlmcguinness-okta has an opinion

PR #776, opened 2017-04-27
Fix text overflow on mobile

### LINKED ISSUE #775: Text overflows to right on mobile
I see there's lots of space on the right side on my phone. Problem seems to be this line:

```java
Kentor.AuthServices.Configuration.Options.GlobalEnableSha256XmlSignatures();
```

And videos need:

```css
style="max-width: 100%"
```

Steps to reproduce: open http://developer.okta.com/blog in your browser and squish it so it's as narrow as it can be. Then scroll to the right. You'll see the blank space and can scroll down to see the elements that overflow.

mraible:
Fixes #775.

### DIFF
--- a/_source/_posts/2015-05-08-software-engineering-design-principles.md
+++ b/_source/_posts/2015-05-08-software-engineering-design-principles.md
@@ -185,5 +185,5 @@ performing code:
 
 ### Reference
 
-[^the-knowledge-creating-company]: Ikujiro Nonaka, and Hirotaka Takeuchi. The Knowledge Creating Company. Oxford University Press, 1995. Print. https://books.google.com/books/about/The_Knowledge_creating_Company.html?id=B-qxrPaU1-MC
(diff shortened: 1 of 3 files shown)


mraible:
@jmelberg-okta Just a test. How does GitHub's markdown handles long strings w/o spaces/dashes in single backticks:

`"Thequickbrownfoxjumpsoverthelazydog"isanEnglishlanguagepangramasentencethatcontainsallofthelettersofthealphabet.`

Same string with three backticks:

```
"Thequickbrownfoxjumpsoverthelazydog"isanEnglishlanguagepangramasentencethatcontainsallofthelettersofthealphabet.
```

It looks like GitHub handles single backticks (by wrapping) better than we do.

jmelberg-okta:
@mraible: Good find. After giving this a bit more thought, as long as we use the three backticks for long strings, we should be fine. Plus, our reviews should catch overhanging text outside of the container.

I've filed a JIRA to investigate a long-term fix for this.

👍

PR #779, opened 2017-04-27
Mark `scope` param as optional for /authorize requests to Custom ASs.

danielzeller-okta:
- Not going to elaborate on WHY it's optional until we have UI support for default scopes.
- For the /token endpoint we don't specify what params are required/optional (http://developer.okta.com/docs/api/resources/oauth2.html#request-parameters-1)

Primary reviewers: @larsjohansen-okta @mystiberry-okta 
@federations-okta


larsjohansen-okta:
🚢

PR #789, opened 2017-05-03
Adds Travis CI

rchild-okta:
## Description:

Two commits:

1. **Adds TravisCI with build and e2e testing**

    - TravisCI config file
    - SauceLabs for master/topic branches, phantomjs for pull requests, and Chrome
      for local testing
    - Swapped out default CLI reporter with one that's more informative
    - Fixed broken selenium tests, and added function for ignoring tests that
      will never pass in phantomjs

2. **Updates bacon ci scripts to work with Travis changes**

    - Moves ci-update-package and ci-pkginfo out of dependencies
    - Refactors linting logic and shared functions into helpers.sh
    - Fixes linting logic, which didn't actually error on linting mistakes
    - Adds linting to Travis script

## Type of Pull Request:
- [x] Functional - adds build infrastructure (travis & bacon changes)

## Resolves:
* [OKTA-124739](https://oktainc.atlassian.net/browse/OKTA-124739)

## How Has This Been Tested?
- [x] Local build - passes
- [x] Travis push build (master, topic branches) - passes
- [x] Travis pull build (PR, external) - passes

## Primary Reviewers:
- @lboyette-okta @sergiymelnyk-okta

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,13 +2,13 @@
 # Okta Developer Site
 
 Okta developer site (developer.okta.com) is a custom [Jekyll](http://jekyllrb.com/) site deployed on [GitHub
-pages](https://pages.github.com/)
+pages](https://pages.github.com/).
 
 All API documentation submissions are welcome. To submit a change, fork this repo, commit your changes, and send us a
-[pull request](http://help.github.com/send-pull-requests/)
+[pull request](http://help.github.com/send-pull-requests/).
 
 
-## Setup Environment
+## Set Up Environment
 
 Most of the content of the developer site is written in a format called
 [kramdown](http://kramdown.gettalong.org/syntax.html) which is a flavor of Markdown with some enhanced features like the
@@ -18,7 +18,7 @@ last resort.
 
 Though kramdown gives us benefits of readability it comes at the cost of requiring a build step to preview the changes to
 the site. Jekyll, handles this nicely by detecting changes to the file system and refreshing automatically. To setup
-Jekyll to be able to build and view the site, follow the steps below: 
+Jekyll to be able to build and view the site, follow the steps below:
 
 1. Install RVM if you don't already have it. On OSX [install the stable version](https://rvm.io/rvm/install) or on linux `yum install rvm`. To ensure you have
    rvm installed properly run `rvm list` and ensure that there is a version marked as 'current'.
@@ -78,15 +78,9 @@ If you are making changes in a fork, here is how to make a clean Pull Request ag
 
 Okta uses the
 [GitHub Flow](https://guides.github.com/introduction/flow/)
-workflow for contributions.
+workflow for contributions. At a high level:
 
-Please read the linked
-[GitHub Flow](https://guides.github.com/introduction/flow/)
-document to learn the details of that workflow.
-
-Here is a high-level overview of the GitHub flow:
-
-1.  (Optional) Fork this repository.
+1.  Fork this repository.
 2.  *Create a branch* using the `git checkout -b $BRANCH_NAME` command.
     Replace `$BRANCH_NAME` with your branch name.
 3.  *Add commits* to your branch using the `git add` and `git commit -m ""` commands.
@@ -94,6 +88,8 @@ Here is a high-level overview of the GitHub flow:
     example: `git push origin $BRANCH_NAME`).
 4.  *Open a Pull Request*
     [using the GitHub UI](https://help.github.com/articles/using-pull-requests/).
+    - For most changes, set `base` to the `weekly` branch. Changes merged to this branch are published with our weekly Preview release.
+    - If you need your change to be published on a different cadence (i.e. for blog posts), set `base` to `master`.
 5.  *Discuss and review your code* using the GitHub UI.
 6.  Once your Pull Request has been reviewed and approved, one of the
     project owners for this site will *merge and deploy* your Pull
--- a/package.json
+++ b/package.json
@@ -9,9 +9,12 @@
     "_site"
   ],
   "scripts": {
-    "ci-update-package": "ci-update-package",
-    "ci-pkginfo:dataload": "ci-pkginfo -t dataload",
-    "test": "webdriver-manager update && protractor tests/selenium/conf.js"
+    "start": "bundle exec jekyll serve",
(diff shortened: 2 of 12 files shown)


package.json
```diff
@@ -9,9 +9,12 @@
     "_site"
   ],
   "scripts": {
-    "ci-update-package": "ci-update-package",
-    "ci-pkginfo:dataload": "ci-pkginfo -t dataload",
-    "test": "webdriver-manager update && protractor tests/selenium/conf.js"
+    "start": "bundle exec jekyll serve",
+    "stop": "if-env CI=true && exit 0 || kill $(lsof -t -i:4000; lsof -t -i:4444)",
```
lboyette-okta: So this won't kill the server if it's in CI, because kill requires permissions we don't have in CI?

package.json
```diff
@@ -9,9 +9,12 @@
...
+    "driver": "if-env PHANTOMJS=true && (phantomjs --webdriver 4444 & sleep 3) || webdriver-manager update",
+    "pretest": "npm start & wait-on http://localhost:4000 && npm run driver",
```
lboyette-okta: argh, the single `&` + `wait-on` is definitely a better solution than `forever` + `wait-on`

scripts/lint.sh
```diff
@@ -1,61 +1,27 @@
...
-
 interject "Running lint.sh in $(pwd)"
 if ! generate_html;
 then
     echo "Error building site"
     exit $BUILD_FAILURE;
 fi
 
-if ! url_consistency_check && duplicate_slug_in_url;
+if ! url_consistency_check || ! duplicate_slug_in_url;
```
lboyette-okta: The naming makes this logic seem trickier than it is (and may have caused the original error). What do you think about renaming them to something like:

- `url_consistency_check` - `has_consistent_urls`
- `duplicate_slug_in_url` - `has_single_slug_in_urls`

tests/selenium/conf.js
```diff
@@ -1,12 +1,59 @@
...
+    print: () => {},
+    defaultTimeoutInterval: 10000
+  },
+
+  specs: ['spec/*.js'],
+  capabilities: {},
+  troubleshoot: true
 };
+
+
```
lboyette-okta: nit: extra space or intended?

package.json
rchild-okta: Couple issues with running stop in CI:
- In Travis with Sauce, there's an error that pops up because they handle killing the sauce connection on their own.
- In Bacon, there's no lsof command

Not a big problem since these machines go away on their own.

package.json
rchild-okta: Yep, simplifies it quite a bit. I originally tried out forever, but it's a bit heavy for this repo.

scripts/lint.sh
rchild-okta: Totally agree. I thought about it, but decided that the real fix is to rewrite these checks in node (or better yet, pull in libraries that will do this and more checks for us). With tests (since we weren't catching anything here anyway). That'll go in with the work that @vijetmahabaleshwar-okta is doing.

tests/selenium/conf.js
rchild-okta: Unintended - will fix.

README.md
```diff
@@ -2,10 +2,10 @@
 # Okta Developer Site
 
 Okta developer site (developer.okta.com) is a custom [Jekyll](http://jekyllrb.com/) site deployed on [GitHub
-pages](https://pages.github.com/)
+pages](https://pages.github.com/).
 
 All API documentation submissions are welcome. To submit a change, fork this repo, commit your changes, and send us a
```
mystiberry-okta: We have two distinct audiences here. We want internal devs to branch, not fork, right? Or is that changing? I thought there was some distinction between the two, but I see in Setup Environment that they seem to be the same thing?

README.md
```diff
@@ -2,10 +2,10 @@
...
-[pull request](http://help.github.com/send-pull-requests/)
+[pull request](http://help.github.com/send-pull-requests/).
 
 
 ## Setup Environment
```
mystiberry-okta: should be Set Up Environment

README.md
```diff
@@ -78,22 +78,18 @@ If you are making changes in a fork, here is how to make a clean Pull Request ag
...
 3.  *Add commits* to your branch using the `git add` and `git commit -m ""` commands.
     Push your commits to your branch with the `git push` command (for
     example: `git push origin $BRANCH_NAME`).
 4.  *Open a Pull Request*
     [using the GitHub UI](https://help.github.com/articles/using-pull-requests/).
+    - For most changes, set `base` to the `weekly` branch. Changes merged to this branch are published with our weekly Preview release.
+    - If you need your change to be published on a different cadence (i.e. for blog posts), set `base` to `master`.
 5.  *Discuss and review your code* using the GitHub UI.
 6.  Once your Pull Request has been reviewed and approved, one of the
     project owners for this site will *merge and deploy* your Pull
```
mystiberry-okta: If they are submitting to 'weekly', we want them to merge, yes?

package.json
mystiberry-okta: Dumb question here: is it safe to use jekyll serve given the localhost:4000 problem? should it be "build" not "serve"?

README.md
mystiberry-okta: The structure of this page is a bit wack, and I don't think we should show the "how to publish to D.O.C." just how to publish locally for testing. But that can follow this PR if you like.

README.md
rchild-okta: I'd like to simplify the model (and align it more closely with how we do our other OSS repos), which is to move to forks for everything, and trim down the number of people with push permissions.

README.md
rchild-okta: Will change - as I was editing this readme, I noticed a lot of text that could be tightened up/is not great. We should followup after this PR with another task to rewrite it.

README.md
rchild-okta: Yep, let's update it in a followup PR. Re us wanting them to merge - yes, I'd consider "project maintainers" to be anyone with push permissions.

rchild-okta:
@mystiberry-okta re "is it safe to use jekyll serve given the localhost:4000 problem? should it be "build" not "serve"?"

This is the command to run it locally and for tests - it has to use localhost:4000 because that's where the tests will run against. 

In a followup PR I'll create another command to make it simpler to build the prod version of the site.

PR #804, opened 2017-05-05
Add pagination for blog entries

mraible:
Add pagination for blog entries. Set to 10 per page in `_config.yml`.

Learned how to do it from https://jekyllrb.com/docs/pagination/.

## Type of Pull Request:

## How Has This Been Tested?
- [x] For functional changes, I have tested the change across Chrome, Firefox, Safari, and mobile.

## Primary Reviewers:

- DevEx

## Before Merging:

- [ ] I have had this PR reviewed and approved on GitHub.
- [ ] I have squashed my commits into `1` commit or will use the **Squash and Merge button**.

### DIFF
--- a/_source/assets/css/page-blog.scss
+++ b/_source/assets/css/page-blog.scss
@@ -22,6 +22,22 @@
   }
 }
 
+#blog {
+    .pagination {
+        text-align: center;
+        margin: 20px 0;
+        
+        a, span, em { 
+          margin: 0 5px;
+        }
+        
+        em {
+          font-style: normal;
+          font-weight: bold;
+        }
+    }
+}
+
 .wrap {
   padding: 0 10px;
 }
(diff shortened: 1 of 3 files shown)


_source/assets/css/page-blog.scss
```diff
@@ -22,6 +22,22 @@
   }
 }
 
+#blog {
```
rchild-okta: nit: let's use the same spacing as in the rest of the file

mraible: Good catch - fixed!

rchild-okta:
Btw I'm looking into why these two tests sometimes fail (we have some code that pulls in the README's from the widget/authjs and creates pages from them). I might just remove them since we're not actually displaying these pages yet on d.o.c.

For now, it's okay to just push again to this branch (I usually just do a rebase to get a new sha and then force push) so that you can get a passing travis run.

PR #819, opened 2017-05-08
Site updated with new blog layout

jakubvul-okta:
## Description:
- Updated blog layout so that only excerpts are included on the main `/blog` page, instead of the entire post.

## Type of Pull Request:

- [X] Functional (CSS changes, JS updates, or layout modifications)

### Resolves:
* N/A

## How Has This Been Tested?
- [X] I have built this locally and verified that it does not break existing functionality.
- [X] For functional changes, I have tested against Chrome, Firefox, and Safari, and mobile.

## Primary Reviewers:

- @rchild-okta


rchild-okta:
Hmm, the strange thing is that the test that is failing is being run in phantom (which runs locally, so should usually be fast). Just rebase and push again - let me know if you need help with that. End state should be that there's 1 pr in this branch.

PR #846, opened 2017-05-15
Imports OSW and AuthJs docs from module README's

rchild-okta:
## Description:
- Added jekyll tag to run npm scripts and replace tag with output
- Added two temporary "external reference" doc pages with the output
  from the OSW and AuthJS readme's
- Created an import-external node script which takes markdown from a module
  and formats it for rendering in d.o.c.

Minor:
- Fixed an issue with invalid asset tags
- Fixed an issue where running npm test results in 143 exit code

## Type of Pull Request:

### Resolves:
* [OKTA-125357](https://oktainc.atlassian.net/browse/OKTA-125357)

## How Has This Been Tested?
- [x] For functional changes, I have added tests.

## Primary Reviewers:
@lboyette-okta 
@vijetmahabaleshwar-okta to check out the protractor changes


_source/_plugins/NpmTag.rb
```diff
@@ -0,0 +1,28 @@
...
+      params = text.split(" ")
+      @run_script = params[0]
+      @arg = params[1]
+      super
+    end
+
+    def render(context)
+      out = `npm run import-external --silent -- #{@arg}`
+      if $?.to_i == 0
+        out
```
lboyette-okta: I think I get it. `npm run import-external` writes output to the console, which is actually written to the document?

tests/selenium/framework/page-objects/BasePage.js
```diff
@@ -16,7 +16,7 @@ class BasePage {
 
   get() {
     browser.ignoreSynchronization = true;
-    browser.get(util.formatUrl(this.url, true));
+    browser.get(this.url);
```
lboyette-okta: do we not need to format the url anymore?

edit: nope, we removed `util.formatUrl` because we don't want to hide the nav anymore

_source/_code/javascript/okta_auth_sdk_external_ref.md
```diff
@@ -0,0 +1,9 @@
+---
+layout: docs_page
+title: Okta Auth SDK
+weight: 100
+hide_from_layout: 1
+excerpt: Reference information for customizing the Okta Sign-In Widget.
```
rchild-okta: Ah, yup will change. I kept on seeing Auth SDK (it's also in the AuthJs github README), so we can stick with it for now. No preference on whether we switch to AuthJs in the future.

package.json
```diff
@@ -12,11 +12,13 @@
     "clean": "rm -rf dist && rm -rf _source/.asset-cache",
     "build-prod": "npm run clean && JEKYLL_ENV=production bundle exec jekyll build --verbose && rsync -av ./dist/ ./",
     "start": "npm run clean && JEKYLL_ENV=production bundle exec jekyll serve --incremental",
-    "stop": "if-env CI=true && exit 0 || kill $(lsof -t -i:4000; lsof -t -i:4444)",
+    "stop": "if-env CI=true && exit 0 || (pkill -f clean && kill $(lsof -t -i:4000; lsof -t -i:4444))",
```
rchild-okta: It's not that it doesn't complete, but rather that when we kill the jekyll process, the npm run script that spawned it will show an ugly 143 exit code error.

So, this is a roundabout way to:
1. Kill the npm process first (that's the one that matches clean)
2. Kill the jekyll process

_source/_plugins/NpmTag.rb
rchild-okta: yep

tests/selenium/framework/page-objects/BasePage.js
rchild-okta: Yeah, that was a courage-only query param

PR #853, opened 2017-05-16
Adding new tests to protractor test suite

vijetmahabaleshwar-okta:
## Description:
Added the following tests for D.O.C deployment verification
* Blog page tests - Verify pagination and other basic features
* Search bar test - Verify "search" functionality on home page

## Type of Pull Request:

- [ ] Blog Post
- [x] Tests added

### Resolves:
* [OKTA-125563](https://oktainc.atlassian.net/browse/OKTA-125563)

## How Has This Been Tested?
- [x] For functional changes, I have added tests.

## Primary Reviewers:

- *Tag reviewer(s)*

- [Blog Post Google Doc](https://docs.google.com)

### DIFF
--- a/tests/selenium/framework/page-objects/BasePage.js
+++ b/tests/selenium/framework/page-objects/BasePage.js
@@ -7,7 +7,7 @@ const baseUrl = 'http://localhost:4000';
 
 class BasePage {
   constructor(relativeURL) {
-    this.pageLoadElement = "";
+    this.$pageLoad = null;
     if(relativeURL) {
       this.url = baseUrl + relativeURL;
     } else {
@@ -15,22 +15,18 @@ class BasePage {
     }
   }
 
-  setPageLoadElement(element) {
-    this.pageLoadElement = element;
+  setPageLoad(element) {
+    this.$pageLoad = element;
   }
 
   load() {
-    this.get();
+    browser.ignoreSynchronization = true;
+    browser.get(this.url);
     return this.waitForPageLoad();
   }
 
   waitForPageLoad() {
-    return util.wait(this.pageLoadElement);
-  }
-
-  get() {
-    browser.ignoreSynchronization = true;
-    browser.get(this.url);
+    return util.wait(this.$pageLoad);
   }
 
   setWindowSize(width, height) {
@@ -63,17 +59,18 @@ class BasePage {
   }
 
   getCurrentURL() {
-    return browser.getCurrentUrl().then(function(url) {
-      // return url relative to baseURL
-      return url.replace(baseUrl, '');
-    })
+    return browser.getCurrentUrl().then(url => url.replace(baseUrl, ''));
   }
-
-  getBaseURL() {
-    return baseUrl;
-  }
-
-  // FIXME: Take sizes from the CSS that we use to define sizes.
+  
+  // These are values used in css for managing different browser sizes -
+  // max-width: 1599px -> xxLarge
+  // max-width: 1399px -> xLarge
+  // max-width: 1199px -> large
+  // max-width: 1023px -> medium
+  // max-width: 899px -> mediumSmall
+  // max-width: 767px -> small
+  // max-width: 479px -> xSmall
+  // max-width: 319px -> xxSmall
   resizeDesktop() {
     browser.driver.manage().window().setSize(1060, 640);
   }
--- a/tests/selenium/framework/page-objects/BasePage.js
+++ b/tests/selenium/framework/page-objects/BasePage.js
@@ -7,11 +7,25 @@ const baseUrl = 'http://localhost:4000';
 
 class BasePage {
   constructor(relativeURL) {
+    this.pageLoadElement = "";
     if(relativeURL) {
       this.url = baseUrl + relativeURL;
     } else {
       this.url = baseUrl;
     }
+  }
+
+  setPageLoadElement(element) {
+    this.pageLoadElement = element;
+  }
+
+  load() {
+    this.get();
+    return this.waitForPageLoad();
+  }
+
+  waitForPageLoad() {
+    return util.wait(this.pageLoadElement);
   }
 
   get() {
@@ -44,22 +58,29 @@ class BasePage {
     })
   }
 
-  elementContainsText(element, expectedTextArray) {
-    return element.getText().then(function(text) {
-      let found = true;
-      for (let i = 0; i < expectedTextArray.length; i++) {
-        if (text.indexOf(expectedTextArray[i]) < 0) {
-          found = false;
-        }
-      }
-      return found;
-    })
-  }
-
   urlContains(str) {
     return EC.urlContains(str)();
   }
 
+  getCurrentURL() {
+    return browser.getCurrentUrl().then(function(url) {
+      // return url relative to baseURL
+      return url.replace(baseUrl, '');
+    })
+  }
+
+  getBaseURL() {
+    return baseUrl;
+  }
+
+  // FIXME: Take sizes from the CSS that we use to define sizes.
+  resizeDesktop() {
+    browser.driver.manage().window().setSize(1060, 640);
+  }
+
+  resizeMobile() {
+    browser.driver.manage().window().setSize(360, 640);
+  }
 }
 
 module.exports = BasePage;
--- a/tests/selenium/framework/page-objects/BasePage.js
+++ b/tests/selenium/framework/page-objects/BasePage.js
@@ -3,7 +3,7 @@
 const util = require('../shared/util');
 const _ = require('lodash');
 const EC = protractor.ExpectedConditions;
-const baseUrl = 'http://localhost:4000/';
+const baseUrl = 'http://localhost:4000';
 
 class BasePage {
   constructor(relativeURL) {
@@ -41,6 +41,18 @@ class BasePage {
       });
     }).then((elementList) => {
       return elementList.length == expectedTextArray.length;
+    })
+  }
+
+  elementContainsText(element, expectedTextArray) {
+    return element.getText().then(function(text) {
+      let found = true;
+      for (let i = 0; i < expectedTextArray.length; i++) {
+        if (text.indexOf(expectedTextArray[i]) < 0) {
+          found = false;
+        }
+      }
+      return found;
     })
   }
 
--- a/tests/selenium/framework/page-objects/BlogPage.js
+++ b/tests/selenium/framework/page-objects/BlogPage.js
@@ -6,42 +6,69 @@ const util = require('../shared/util');
 class BlogPage extends BasePage {
   constructor(url) {
     super(url);
-    this.blog = $('.wrap.blog');
-    this.blogPosts = $$('article.post-block');
-    this.paginationDiv = $('.pagination');
+    this.blogElement = $('.wrap.blog');
+    this.blogPostElements = $$('article.post-block');
+    this.paginationElement = $('.pagination');
+    this.nextLink = element(by.partialLinkText('Next'));
+    this.prevLink = element(by.partialLinkText('Prev'));
+    this.setPageLoadElement(this.blogElement);
   }
-
-  load() {
-    this.get();
-    return this.waitForPageLoad();
-  }
-
-  waitForPageLoad() {
-    return util.wait(this.blog);
-  }
-
+  
   getBlogPostCount() {
-    return this.blogPosts.count();
+    return this.blogPostElements.count();
   }
 
   isPaginationVisible() {
-    return this.paginationDiv.isDisplayed();
+    return this.paginationElement.isDisplayed();
   }
 
   doesPaginationHaveLinks() {
     let links = ['Prev', '1', '2', 'Next'];
-    return this.elementContainsText(this.paginationDiv, links);
+
+    return this.paginationElement.getText().then(function(text) {
+      for (let i = 0; i < links.length; i++) {
+        if (text.indexOf(links[i]) < 0) {
+          return false;
+        }
+      }
+      return true;
+    })
   }
 
   doBlogsHaveReadMoreLink() {
     let blogCount = 0;
-    return this.blogPosts.filter(function (blogPost) {
+    return this.blogPostElements.filter(function (blogPost) {
       blogCount++;
       let readMoreLink = blogPost.element(by.linkText('Read more'));
       return readMoreLink.isPresent();
     }).then(function (elementList) {
       return elementList.length == blogCount;
     })
+  }
+
+  clickNext() {
+    return this.nextLink.click();
+  }
+
+  clickPrevious() {
+    return this.prevLink.click();
+  }
+
+  clickItem(item) {
+    let itemLink = element(by.linkText(item.toString()));
+    return itemLink.click();
+  }
+
+  clickReadMoreOnPost(post) {
+    let blogPost = this.blogPostElements.get(post);
+    let readMoreLink = blogPost.element(by.linkText('Read more'));
+    return readMoreLink.click();
+  }
+  
+  getBlogLink(post) {
+    let blogPost = this.blogPostElements.get(post);
+    let title = blogPost.element(by.css('a'));
+    return title.getAttribute('href');
   }
 }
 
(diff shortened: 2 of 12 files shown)


tests/selenium/framework/page-objects/BlogPage.js
```diff
@@ -0,0 +1,48 @@
...
+
+const BasePage = require('./BasePage');
+const util = require('../shared/util');
+
+class BlogPage extends BasePage {
+  constructor(url) {
+    super(url);
+    this.blog = $('.wrap.blog');
+    this.blogPosts = $$('article.post-block');
+    this.paginationDiv = $('.pagination');
```
rchild-okta: Nit - we should probably come up with some convention for this, but this should be something like "paginationEl" or "paginationElement" (like is used elsewhere, i.e h3Elements) - div is too specific and would change if we decided to change the underlying structure of the page.

And, use that convention for blog and blogPosts, since they don't have any "el" or "element" attached.

tests/selenium/framework/page-objects/BasePage.js
```diff
@@ -44,6 +44,18 @@ class BasePage {
     })
   }
 
+  elementContainsText(element, expectedTextArray) {
```
rchild-okta: Hmm, the naming is a bit confusing, i.e.

- elementsContainText - search all the elements, find the text, and for the expectedTextArray find **exact** matches for the text in the elements 
- elementContainsText - for the given element, see if each of the expectedTextArray strings exist in the text of the element

tests/selenium/framework/page-objects/BlogPage.js
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  load() {
+    this.get();
```
rchild-okta: Seems like we're doing this all the time - why not just have a base load method that navigates to the page and then calls the waitForPageLoad method? Seems like we would never want to just call the page but not wait for it to load.

tests/selenium/framework/page-objects/BlogPage.js
```diff
@@ -0,0 +1,48 @@
...
+    return this.waitForPageLoad();
+  }
+
+  waitForPageLoad() {
+    return util.wait(this.blog);
```
rchild-okta: And for this, could also bump it out to basePage and standardize on the "pageLoadEl", i.e. each view would be responsible for just setting this element, but not having to define a specific method for it.

tests/selenium/framework/page-objects/BlogPage.js
```diff
@@ -0,0 +1,48 @@
...
+
+  getBlogPostCount() {
+    return this.blogPosts.count();
+  }
+
+  isPaginationVisible() {
+    return this.paginationDiv.isDisplayed();
+  }
+
+  doesPaginationHaveLinks() {
```
rchild-okta: I think this is an okay test, but better would be to actually click into the links to verify that the links actually work. I.e. instead make a couple methods like "clickNext", "clickPrev", "clickItem(1)"

tests/selenium/framework/page-objects/BlogPage.js
```diff
@@ -0,0 +1,48 @@
...
+    let links = ['Prev', '1', '2', 'Next'];
+    return this.elementContainsText(this.paginationDiv, links);
+  }
+
+  doBlogsHaveReadMoreLink() {
```
rchild-okta: Same here - would rather the method was "clickPostReadMore" and then pass in a number for which blog post, and then verify that we actually redirect to the correct page.

tests/selenium/framework/page-objects/NavPage.js
```diff
@@ -31,6 +36,36 @@ class NavPage extends BasePage {
   isMobileNavDisplayed() {
     return this.mobileNav.isDisplayed();
   }
+
+  clickSearchIcon() {
+    util.waitTillClickable(this.searchIcon);
```
rchild-okta: nit: would prefer "waitUntilClickable"

Also, am curious why this would even be an issue? I.e. is it not clickable on page load? If it's not, should we wait for a different element before saying that the page has loaded?

Or, if it's about the transition in window size, we should have a different wait that is along the lines of windowResizeFinished or something like that.

tests/selenium/spec/blog-page-tests.js
```diff
@@ -0,0 +1,17 @@
...
+
+describe('blog page tests', function () {
+    const blogPage = new BlogPage('/blog');
+
+    beforeEach(function () {
+        browser.ignoreSynchronization = true;
+        blogPage.load();
+    });
+    
+    it('has blog posts with read more links and pagination', function () {
```
rchild-okta: Related to other comments, probably better to split this out into two:

1. it has blog posts with read more links (where we'd click through to the actual blog post)
2. it has pagination (where we'd click through prev, next, and one of the numbers, and ideally check that we end up on the right page by looking at the url)

tests/selenium/spec/blog-page-tests.js
```diff
@@ -0,0 +1,17 @@
...
+        expect(blogPage.getBlogPostCount()).toBeGreaterThan(0);
+        expect(blogPage.doBlogsHaveReadMoreLink()).toBeTruthy();
```
rchild-okta: Noticed this from before, but better to be explicit - expect(foo).toBe(true). Can result in some subtle errors.

tests/selenium/spec/search-bar-tests.js
```diff
@@ -0,0 +1,35 @@
...
+describe('navigation bar search tests', function() {
+  const navPage = new NavPage();
+
+  beforeEach(function() {
+    browser.ignoreSynchronization = true;
+    navPage.load();
+  });
+
+  it('does search on desktop browser sizes', function () {
+    navPage.setWindowSize(1060, 640);
```
rchild-okta: Now's probably a good time to just stick this on the base page object. Just create a couple methods like "resizeDesktop" and "resizeMobile"

tests/selenium/spec/search-bar-tests.js
```diff
@@ -0,0 +1,35 @@
...
+
+    navPage.clickSearchIcon();
+    expect(navPage.areSearchResultsPresent()).toBeFalsy();
+
+    navPage.enterSearchText('Authentication');
+    navPage.submitSearch();
+
+    expect(navPage.areSearchResultsPresent()).toBeTruthy();
```
rchild-okta: Rather than check for presence, probably better to check that the results we expect are there.

tests/selenium/framework/page-objects/BlogPage.js
vijetmahabaleshwar-okta: Done

tests/selenium/framework/page-objects/BlogPage.js
vijetmahabaleshwar-okta: Done

tests/selenium/framework/page-objects/BlogPage.js
vijetmahabaleshwar-okta: Done

tests/selenium/framework/page-objects/BlogPage.js
vijetmahabaleshwar-okta: Done

tests/selenium/framework/page-objects/BlogPage.js
vijetmahabaleshwar-okta: Done

tests/selenium/framework/page-objects/NavPage.js
vijetmahabaleshwar-okta: It was mainly for the window resize. I saw the test fail sometimes due to that. This was only a safety net to avoid that. But looks like it's not reproducible locally anymore. I can remove this safety net now. Might have been some other issue altogether. 
Will look for failures on travis as well to see if something needs to be added. For now, will remove the method.

tests/selenium/spec/blog-page-tests.js
vijetmahabaleshwar-okta: Done

tests/selenium/spec/blog-page-tests.js
vijetmahabaleshwar-okta: done

tests/selenium/spec/search-bar-tests.js
vijetmahabaleshwar-okta: Done

tests/selenium/spec/search-bar-tests.js
vijetmahabaleshwar-okta: I thought about it but this search is provided by google. I could check for the results but it'd be more like testing google search than anything else. Do you think there's much value added here by checking results?

tests/selenium/framework/page-objects/BasePage.js
vijetmahabaleshwar-okta: Agree.
Actually the method "elementContainsText" is called only once and there I use it to check for link text. For now, I will remove the method and add the logic inside the method it's being used.
In the future, if more methods need this logic, will abstract it out.

tests/selenium/framework/page-objects/BasePage.js
```diff
@@ -48,6 +62,25 @@ class BasePage {
     return EC.urlContains(str)();
   }
 
+  getCurrentURL() {
+    return browser.getCurrentUrl().then(function(url) {
```
rchild-okta: fyi, we can now use [fat arrows](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions), i.e.
```javascript
return browser.getCurrentUrl().then(url => url.replace(baseUrl, ''));
```

tests/selenium/framework/page-objects/BasePage.js
```diff
@@ -48,6 +62,25 @@ class BasePage {
...
+      // return url relative to baseURL
+      return url.replace(baseUrl, '');
+    })
+  }
+
+  getBaseURL() {
+    return baseUrl;
+  }
+
+  // FIXME: Take sizes from the CSS that we use to define sizes.
```
rchild-okta: I think it's probably okay not to take it from the CSS. It would however be nice to add a comment describing what the responsive sizes are.

tests/selenium/framework/page-objects/BlogPage.js
```diff
@@ -0,0 +1,75 @@
+'use strict';
+
+const BasePage = require('./BasePage');
+const util = require('../shared/util');
+
+class BlogPage extends BasePage {
+  constructor(url) {
+    super(url);
+    this.blogElement = $('.wrap.blog');
```
rchild-okta: I probably should have mentioned it before, but I do prefer "EL" rather than "Element" (el is pretty idiomatic to describe elements in JS). I.e. this.blogEl, etc

tests/selenium/framework/page-objects/BlogPage.js
```diff
@@ -0,0 +1,75 @@
...
+  
+  getBlogPostCount() {
+    return this.blogPostElements.count();
+  }
+
+  isPaginationVisible() {
+    return this.paginationElement.isDisplayed();
+  }
+
+  doesPaginationHaveLinks() {
```
rchild-okta: It feels like it's overkill to have a check for this - don't we get the check just by trying to click the next/prev/1 links?

tests/selenium/framework/page-objects/BlogPage.js
```diff
@@ -0,0 +1,75 @@
...
+      for (let i = 0; i < links.length; i++) {
+        if (text.indexOf(links[i]) < 0) {
+          return false;
+        }
+      }
+      return true;
+    })
+  }
+
+  doBlogsHaveReadMoreLink() {
```
rchild-okta: I'd say the same thing here - in general, feels like we can just click on a link in the blog post and verify that this works rather than making an explicit check that all posts have readMore links.

tests/selenium/spec/blog-page-tests.js
```diff
@@ -0,0 +1,34 @@
+const BlogPage = require('../framework/page-objects/BlogPage');
+
+describe('blog page tests', function() {
```
rchild-okta: For this and the filename, can remove "tests" (some of our projects use -spec instead in the fileName)

tests/selenium/spec/blog-page-tests.js
```diff
@@ -0,0 +1,34 @@
...
+  const blogPage = new BlogPage('/blog');
+
+  beforeEach(function() {
+    browser.ignoreSynchronization = true;
```
rchild-okta: Do we need this?

tests/selenium/spec/search-bar-tests.js
```diff
@@ -0,0 +1,35 @@
+const NavPage = require('../framework/page-objects/NavPage');
+const util = require('../framework/shared/util');
+
+describe('navigation bar search tests', function() {
+  const navPage = new NavPage();
+
+  beforeEach(function() {
+    browser.ignoreSynchronization = true;
```
rchild-okta: Don't need this anymore right?

tests/selenium/framework/page-objects/BasePage.js
```diff
@@ -3,17 +3,31 @@
 const util = require('../shared/util');
 const _ = require('lodash');
 const EC = protractor.ExpectedConditions;
-const baseUrl = 'http://localhost:4000/';
+const baseUrl = 'http://localhost:4000';
 
 class BasePage {
   constructor(relativeURL) {
+    this.pageLoadElement = "";
```
vijetmahabaleshwar-okta: Done.

tests/selenium/framework/page-objects/BasePage.js
```diff
@@ -3,17 +3,31 @@
...
       this.url = baseUrl;
     }
   }
 
+  setPageLoadElement(element) {
+    this.pageLoadElement = element;
+  }
+
+  load() {
+    this.get();
```
vijetmahabaleshwar-okta: Yes, not used outside. getting rid of it.

tests/selenium/framework/page-objects/BasePage.js
vijetmahabaleshwar-okta: I'm used to the old school way. Time to unlearn. Old habits die hard. :)

tests/selenium/framework/page-objects/BlogPage.js
vijetmahabaleshwar-okta: Coming from java, this looks different to me. Will change it. :)

tests/selenium/framework/page-objects/BlogPage.js
vijetmahabaleshwar-okta: That's true. Will remove this.

tests/selenium/framework/page-objects/BlogPage.js
vijetmahabaleshwar-okta: Got it

tests/selenium/spec/blog-page-tests.js
vijetmahabaleshwar-okta: Yes, not needed.

tests/selenium/spec/search-bar-tests.js
vijetmahabaleshwar-okta: Nope. Will remove.

tests/selenium/spec/blog-page-tests.js
vijetmahabaleshwar-okta: Done

tests/selenium/framework/page-objects/BasePage.js
```diff
@@ -7,30 +7,26 @@ const baseUrl = 'http://localhost:4000';
 
 class BasePage {
   constructor(relativeURL) {
-    this.pageLoadElement = "";
+    this.$page = null;
```
rchild-okta: nit: I do like calling out "$pageLoad" here since it is an element that is used to determine that the page has finished rendering (i.e. in the case of docs/ it's the sidebar element, which feels distinct from what a $page might refer to)

vijetmahabaleshwar-okta: Makes sense. Changing it to $pageLoad and method to "setPageLoad"

PR #861, opened 2017-05-17
:seedling: Adds localhost url checker

jmelberg-okta:
## Description:
- Adds a build script to check for occurrences of `localhost:4000`

## Type of Pull Request:

- [ ] Blog Post

### Resolves:
* [OKTA-126472](https://oktainc.atlassian.net/browse/OKTA-126472)

## How Has This Been Tested?


jmelberg-okta:
@rchild-okta & @vijetmahabaleshwar-okta for your review

scripts/setup.sh
```diff
@@ -88,3 +88,15 @@ function require_env_var() {
         exit 1
     fi
 }
+
+function check_for_localhost_links() {
```
rchild-okta: Don't really need this right? I.e. setup.sh is already importing helpers.sh.

Rather, what we want to to do is add the function invocation to lint.sh (as well as travis.sh)

scripts/helpers.sh
```diff
@@ -114,3 +114,15 @@ function require_env_var() {
         exit 1
     fi
 }
+
+function check_for_localhost_links() {
+    function grep_for_localhost() {
+        grep -R "localhost:4000" ../* --exclude-dir={node_modules,scripts,tests,dist} --exclude={README.md,package.json}
+    }
+
+    if [ grep_for_localhost ];
```
rchild-okta: I think it's actually better to just return the result of the command, and then check if it's a bad exist code in the place that's calling it.

scripts/helpers.sh
```diff
@@ -114,3 +114,12 @@ function require_env_var() {
         exit 1
     fi
 }
+
+function check_for_localhost_links() {
+    if grep -R "localhost:4000" ../* --exclude-dir={node_modules,scripts,tests,dist} --exclude={README.md,package.json}
+    then
+        echo 1
```
rchild-okta: Better to **echo** a useful error statement (letting the user know why the lint check failed, i.e. because there are hardcoded localhost links), and **return** the result of the function (like in the other functions in this helper file).

scripts/lint.sh
```diff
@@ -27,4 +27,11 @@ then
     exit $BUILD_FAILURE;
 fi
 
+# Run localhost:4000 checker to make sure we removed all references correctly
```
rchild-okta: And in this case, since it's in line with the other lint checks, I would instead just move this as another check in the if condition with url_consistency_check, etc. 

So general idea here is to output error message, have the comment about what it's doing in the helper function, and make the invocation of the function simple.

scripts/helpers.sh
jmelberg-okta: @rchild-okta: Did a slightly different approach:
 - On failure:
    - `echo` the pathname and text of the located `localhost:4000`
    - `echo` error message
    - `return 1`
  - On success:
    - Do nothing

Then we can check for the output of the function the same way we do `url_consistency_check` and `duplicate_slug_in_url`.

scripts/travis.sh
```diff
@@ -34,5 +34,12 @@ if ! url_consistency_check || ! duplicate_slug_in_url; then
   exit 1;
 fi
 
-# 5. Run htmlproofer to validate links, scripts, and images
+# 5. Run localhost:4000 checker to make sure we removed all references correctly
+localhost=$(check_for_localhost_links)
```
rchild-okta: Need to update this here as well

jmelberg-okta: Oops. Oversight there.

PR #874, opened 2017-05-22
Always use UTC for feed and sitemap generation

bdemers:
## Description:
- Always use UTC for feed and sitemap generation

## Type of Pull Request:

- [ ] Blog Post

## How Has This Been Tested?


mraible:
LGTM 👍

bdemers:
Setting the TZ lets makes the dates generate the same between builds and removes the date issues we have been seeing between PRs

This just makes all dates UTC.
How we display the dates is a different issue, but the raw date _should_ have a timezone

PR #882, opened 2017-05-24
Add Disqus for comments on blog posts

mraible:
## Description:

Implementing Disqus with Jekyll for blog post comments. Used tutorial at: http://vcabbage.com/development/2015/06/21/implementing-disqus-with-jekyll.html

I tried to implement comment count on the blog index page, but couldn't get it working right.

## Type of Pull Request:

- [ ] Blog Post

### Resolves:

## How Has This Been Tested?
- [x] For functional changes, I have tested against Chrome, Firefox, and Safari, and mobile.

## Primary Reviewers:

DevEx

### DIFF
--- a/3rd_party_notices/index.html
+++ b/3rd_party_notices/index.html
@@ -37,6 +37,10 @@
   <meta name="description" content="                        Okta Android Mobile product            Okta Android Verify product            Okta Connector Agent Tool product            Okta Frame...">
   <link rel="canonical" href="http://developer.okta.com/3rd_party_notices/index">
   <link rel="alternate" type="application/rss+xml" title="Okta Developer" href="http://developer.okta.com/feed.xml"><!-- GA -->
+  <script type="text/javascript">
+    var disqus_shortname = 'oktadevblog';
+  </script>
+  </head>
 </head>
 
 <body>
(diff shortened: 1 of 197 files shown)


mraible:
@bdemers You are correct.

mystiberry-okta:
Won't this code be overwritten in the generated doc?

3rd_party_notices/index.html
```diff
@@ -37,6 +37,10 @@
   <meta name="description" content="                        Okta Android Mobile product            Okta Android Verify product            Okta Connector Agent Tool product            Okta Frame...">
   <link rel="canonical" href="http://developer.okta.com/3rd_party_notices/index">
   <link rel="alternate" type="application/rss+xml" title="Okta Developer" href="http://developer.okta.com/feed.xml"><!-- GA -->
+  <script type="text/javascript">
```
rchild-okta: We don't need this to show up in 3rd party notices right? Better to add it in a blog only template, or if it must go in head.html, then check for whether some variable defined in the blog front matter is actually defined before injecting this snippet of code.

johngronberg-okta:
I don't think we should use Disqus for this right now. We can embed discourse comments (which would be a slick integration and automatically create topics in the forum for new blog posts). See: https://meta.discourse.org/t/embedding-discourse-comments-via-javascript/31963 It also means we don't need to go through another vendor/security review.

PR #954, opened 2017-06-15
New content for Social Auth page

jakubvul-okta:
## Description:
This is a total refresh of the existing Social Authentication page (which can be found [here](https://developer.okta.com/docs/api/resources/social_authentication.html)). I tried following the instructions on the existing page and I was not able to get social auth working. The instructions in this PR result in working social auth.

Some particular points that I am unsure of:

- I'm not sure if these are _the_ steps to follow to get social authentication working, or just one possible way to get it working. All I know is that this is what I did to get it working.
- I'm not sure if we should mention the Sign-in Widget and Auth.js here at all (even though the mentions are quite brief), or just omit them entirely.

PDF, as requested: [Social Authentication.pdf](https://github.com/okta/okta.github.io/files/1079072/Social.Authentication.pdf)

## Type of Pull Request:

- [ ] Blog Post

### Resolves:
* [OKTA-130265](https://oktainc.atlassian.net/browse/OKTA-130265)

## How Has This Been Tested?


_source/_docs/api/resources/social_authentication.md
```diff
@@ -6,127 +6,551 @@ excerpt: Setting up an Okta Social Authentication provider
 
 # Social Authentication
 
-Okta Social Authentication allows Okta to integrate with a Social Identity Provider such as Facebook, Google, LinkedIn, and Microsoft. These integrations support the following features:
+---
```
rchild-okta: This to the header needs to be removed (its duped on the page)

_source/_docs/api/resources/social_authentication.md
```diff
@@ -6,127 +6,551 @@ excerpt: Setting up an Okta Social Authentication provider
...
+excerpt: Setting up an Okta Social Authentication provider
+---
+
+# Social Authentication
+
+## Overview
+
+### Features
+
+Integrating social login with Okta allows you to use the following features:
```
rchild-okta: General - is there a reason to remove the diagram that used to show the flow?

_source/_docs/api/resources/social_authentication.md
```diff
@@ -6,127 +6,551 @@ excerpt: Setting up an Okta Social Authentication provider
...
+
+- **User Registration:** Capture the Profile attributes from a Social Identity Provider user and store those attributes in Okta's Universal Directory.
+
+- **User Authentication:** After a user is registered, continue to use that Social Identity Provider for user authentication, thus eliminating the need to store an additional username and password for that user.
+
+- **Social Profile Sync:** If a user updates their Social Profile, those changes can be reflected inside Okta the next time they use social login.
```
rchild-okta: Am curious if it only happens when they auth with the social idp, or if it works in some other way

_source/_docs/api/resources/social_authentication.md
```diff
@@ -6,127 +6,551 @@ excerpt: Setting up an Okta Social Authentication provider
...
+
+#### Set-up
+
+To set-up social login, you configure the following: 
+
+1. An OAuth client in your social provider
+2. An Identity Provider in Okta
+3. An OAuth Application in Okta
+
+Every Identity Provider in Okta is linked to an Application. Every time a user signs in with a Social Identity Provider for the first time, an Application User is created for them. This means that, for example, if a user signed in with Facebook and Google, they would have three Users in Okta: a Facebook User, a Google User, and a master Okta User that links these two. For more information about how to configure this behavior see [Identity Provider Settings](#idp-settings).
```
rchild-okta: From their perspective, is this true? Or instead, we create one Okta user (the user they see when they go to the People page), and incidentally create AppUsers for Facebook and Google which can be used to map different profile attributes.

_source/_docs/api/resources/social_authentication.md
```diff
@@ -6,127 +6,551 @@ excerpt: Setting up an Okta Social Authentication provider
 
 # Social Authentication
 
-Okta Social Authentication allows Okta to integrate with a Social Identity Provider such as Facebook, Google, LinkedIn, and Microsoft. These integrations support the following features:
```
mystiberry-okta: Can we have an intro statement before launching into features? The stem over the list is a good start but not quite enough?

_source/_docs/api/resources/social_authentication.md
```diff
@@ -6,127 +6,551 @@ excerpt: Setting up an Okta Social Authentication provider
...
+
+Every Identity Provider in Okta is linked to an Application. Every time a user signs in with a Social Identity Provider for the first time, an Application User is created for them. This means that, for example, if a user signed in with Facebook and Google, they would have three Users in Okta: a Facebook User, a Google User, and a master Okta User that links these two. For more information about how to configure this behavior see [Identity Provider Settings](#idp-settings).
+
+#### The Social Login Process
+
+In general, the social login process works as follows:
+
+1. The user who wishes to authenticate will click a “Log in with x” link.
+2. The user authenticates and is asked by the Provider to accept the permissions required by your app.
+3. Once the user accepts the permissions, Okta handles the login and redirects the user back to your specified redirect URI with a OpenID Connect ID Token.
```
rchild-okta: This would be a nice place to put that (or a) diagram

_source/_docs/api/resources/social_authentication.md
mystiberry-okta: Not sure we should use integration here but maybe implement? And the customer's goals is not usually to use our features but to do something else, (missing from the original intro too!). Something like "make it easy for your customers or end users to log in to Okta by allowing them to log in with their existing social media accounts." That's probably too long and may be too obvious, but starting with a brief statement of the value prop is helpful.

_source/_docs/api/resources/social_authentication.md
```diff
@@ -6,127 +6,551 @@ excerpt: Setting up an Okta Social Authentication provider
...
+- **response_type:** Determines which flow is used. This should be `id_token`.
+- **response_mode:** Determines how the authorization response should be returned.
+- **state:** Protects against cross-site request forgery (CSRF).
+- **nonce:** A string included in the returned ID Token. Use it to associate a client session with an ID Token, and to mitigate replay attacks.
+- **redirect_url:** The location where Okta returns a user after the user has finished authenticating against their Social Authentication provider. This URL must start with “https” and must match one of the Redirect URIs  that you configured previously.
+
+For a full explanation of all these parameters, see here: [OAuth 2.0 Request parameters](https://developer.okta.com/docs/api/resources/oauth2.html#request-parameters).
+
+An example of a complete URL looks like this: `https://dev-263012.oktapreview.com/oauth2/v1/authorize?idp=0oaaq9pjc2ujmFZexample&client_id=GkGw4K49N4UEE1example&response_type=id_token&response_mode=fragment&scope=openid&redirect_uri=http://yourAppUrlHere.com/social_auth&state=foo&nonce=bar`
```
rchild-okta: How do we show these types of urls elsewhere in the docs? Specifically:
- Using foo and bar as state and nonce - would probably be better to use something a little more opaque (to at least suggest that they should think about what they're putting there
- Do we care if we expose a real org with real info (although it looks like you probably deleted the app)

_source/_docs/api/resources/social_authentication.md
```diff
@@ -6,127 +6,551 @@ excerpt: Setting up an Okta Social Authentication provider
...
+
+Okta also offers an easily embeddable JavaScript widget that reproduces the look and behavior of the standard Okta sign-in page. Adding a "Login with Facebook" button is as simple as adding the following code to your configuration:
+
+```
+idps: [
+  {type: 'FACEBOOK', id: '$Your_FB_IDP_ID_Here'}
+]
+```
+
+You can find out more about it [on GitHub](https://github.com/okta/okta-auth-js#introduction). Implementing login with Facebook would use the Widget's [OpenID Connect authentication flow](https://github.com/okta/okta-signin-widget#openid-connect).
```
rchild-okta: First link should be link to the widget, not authJs

cwu-okta:
@johngronberg-okta

_source/_docs/api/resources/social_authentication.md
```diff
@@ -6,127 +6,551 @@ excerpt: Setting up an Okta Social Authentication provider
...
+
+- **User Registration:** Capture the Profile attributes from a Social Identity Provider user and store those attributes in Okta's Universal Directory.
+
+- **User Authentication:** After a user is registered, continue to use that Social Identity Provider for user authentication, thus eliminating the need to store an additional username and password for that user.
+
+- **Social Profile Sync:** If a user updates their Social Profile, those changes can be reflected inside Okta the next time they use social login.
+
+- **Support for Multiple Social Profiles** Multiple Social Profiles can all be linked to one master Okta user.
+
+- **Support for Webhooks:** Make callouts to your application during user registration and/or account linking. For example: prompt a user for additional attributes before linking their account, or enable a customized validation or confirmation flow.
```
mystiberry-okta: Do we support webhooks now? That isn't mentioned anywhere else in this doc.

_source/_docs/api/resources/social_authentication.md
```diff
@@ -6,127 +6,551 @@ excerpt: Setting up an Okta Social Authentication provider
...
+
+- **OAuth Scope Configuration:** Specify OAuth2 scopes to fully control which Social Profile attributes are linked to Okta.
```
mystiberry-okta: We use "OAuth 2.0" instead of OAuth2 in prose (unless referring to the endpoint /oauth2

_source/_docs/api/resources/social_authentication.md
```diff
@@ -6,127 +6,551 @@ excerpt: Setting up an Okta Social Authentication provider
...
+
+- **Social Profile Sync:** If a user updates their Social Profile, those changes can be reflected inside Okta the next time they use social login.
+
+- **Support for Multiple Social Profiles** Multiple Social Profiles can all be linked to one master Okta user.
+
+- **Support for Webhooks:** Make callouts to your application during user registration and/or account linking. For example: prompt a user for additional attributes before linking their account, or enable a customized validation or confirmation flow.
+
+- **OAuth Scope Configuration:** Specify OAuth2 scopes to fully control which Social Profile attributes are linked to Okta. 
+
+### How it Works
```
mystiberry-okta: what is this heading for? "How It Works" Should "The Social Login Process be before the how-to section? That way you can just eliminate "how it works."

_source/_docs/api/resources/social_authentication.md
```diff
@@ -6,127 +6,551 @@ excerpt: Setting up an Okta Social Authentication provider
...
-    steps and give you back a `client_id` which you use in Step 7. For social authentication, choose the following:
+### Authentication Settings
 
-      1. In the **Create a New Application Integration** page of the wizard, select **OpenID Connect** in the **Sign on method** section.
-      2. In the **General Settings** tab, check the **Implicit** checkbox for allowed grant types.
+**IdP username:** The IdP attribute that should map to the Okta username. The dropdown list contains the default value, `idpuser.email`.
 
-7.  Create a Social Auth login (Authorize) URL by replacing the value of the following parameters in the Social Auth "Authorize URL."
+You can enter an expression to reformat the value, if desired. For example, if the username in the SAML assertion is `john.doe@mycompany.okta.com`, you could specify the replacement of `mycompany.okta` with `endpointA.mycompany` to make the transformed username `john.doe@endpointA.mycompany.com`. If you want to enter an expression, follow the [Okta Expression Language link](http://developer.okta.com/docs/getting_started/okta_expression_lang.html) for syntax requirements.
```
rchild-okta: if the username in the SAML assertion -> something like "if the social username" ...

_source/_docs/api/resources/social_authentication.md
```diff
@@ -6,127 +6,551 @@ excerpt: Setting up an Okta Social Authentication provider
...
+**Provisioning Policy:** Determines whether just-in-time provisioning of users should be automatic, disabled, or delegated to an external application via a callout.
 
-8.  Add the Social Auth Authorize URL to the page where you want to
-    enable Social Auth. Using the example URL from step 7, your HTML looks similar to the following:
+### JIT Settings
 
-    ```html
-    <a href="https://example.okta.com/oauth2/v1/authorize?idp=0oa0bcde12fghiJkl3m4&client_id=AbcDE0fGHI1jk2LM34no&scope=openid%20email%20profile&response_type=id_token&response_mode=fragment&state=someState&nonce=someNonce&redirect_uri=https://app.example.com/social_auth">Log in</a>
-    ```
+**Profile Master:** If an authenticated user exists, the user is updated with the information in this SAML assertion when this box is selected. If this box is not selected, the flow continues without updating information.
```
rchild-okta: SAML assertion -> something else

_source/_docs/api/resources/social_authentication.md
```diff
@@ -6,127 +6,551 @@ excerpt: Setting up an Okta Social Authentication provider
...
-8.  Add the Social Auth Authorize URL to the page where you want to
-    enable Social Auth. Using the example URL from step 7, your HTML looks similar to the following:
+### JIT Settings
 
-    ```html
-    <a href="https://example.okta.com/oauth2/v1/authorize?idp=0oa0bcde12fghiJkl3m4&client_id=AbcDE0fGHI1jk2LM34no&scope=openid%20email%20profile&response_type=id_token&response_mode=fragment&state=someState&nonce=someNonce&redirect_uri=https://app.example.com/social_auth">Log in</a>
-    ```
+**Profile Master:** If an authenticated user exists, the user is updated with the information in this SAML assertion when this box is selected. If this box is not selected, the flow continues without updating information. 
 
+**Group Assignments:** Allows you to assign new users to one ore more existing Groups. For example, new Facebook users could be added to a "Facebook" Group.
```
rchild-okta: ore -> or

_source/_docs/api/resources/social_authentication.md
```diff
@@ -6,127 +6,551 @@ excerpt: Setting up an Okta Social Authentication provider
...
+
+- **Support for Multiple Social Profiles** Multiple Social Profiles can all be linked to one master Okta user.
+
+- **Support for Webhooks:** Make callouts to your application during user registration and/or account linking. For example: prompt a user for additional attributes before linking their account, or enable a customized validation or confirmation flow.
+
+- **OAuth Scope Configuration:** Specify OAuth2 scopes to fully control which Social Profile attributes are linked to Okta. 
+
+### How it Works
+
+#### Set-up
```
mystiberry-okta: I mistook this for instructions. I'd move it *after* Social Login Process,  and change the title to something a little clearer, Setting Up Social Logins.

_source/_docs/api/resources/social_authentication.md
mystiberry-okta: Why is this paragraph here? Might deserve its own section heading like "Social Login and Okta Users" or something. Also, I'm unsure how an IdP is different from a Social IdP when I read this paragraph. It seems to go with the content at the end?

_source/_docs/api/resources/social_authentication.md
```diff
@@ -6,127 +6,551 @@ excerpt: Setting up an Okta Social Authentication provider
...
+
+1. An OAuth client in your social provider
+2. An Identity Provider in Okta
+3. An OAuth Application in Okta
+
+Every Identity Provider in Okta is linked to an Application. Every time a user signs in with a Social Identity Provider for the first time, an Application User is created for them. This means that, for example, if a user signed in with Facebook and Google, they would have three Users in Okta: a Facebook User, a Google User, and a master Okta User that links these two. For more information about how to configure this behavior see [Identity Provider Settings](#idp-settings).
+
+#### The Social Login Process
+
+In general, the social login process works as follows:
```
mystiberry-okta: This stem doesn't give me enough context to easily consume the numbered list. Can we say something like "Implementing social authentication flows travel from auth point to to provider to Okta and back" or something that would help set context for the numbered list? Also, "In general" isn't needed.

_source/_docs/api/resources/social_authentication.md
```diff
@@ -6,127 +6,551 @@ excerpt: Setting up an Okta Social Authentication provider
...
+
+1. The user who wishes to authenticate will click a “Log in with x” link.
```
mystiberry-okta: we try to avoid future tense unless it's needed. will click > clicks

_source/_docs/api/resources/social_authentication.md
```diff
@@ -6,127 +6,551 @@ excerpt: Setting up an Okta Social Authentication provider
...
+2. The user authenticates and is asked by the Provider to accept the permissions required by your app.
+3. Once the user accepts the permissions, Okta handles the login and redirects the user back to your specified redirect URI with a OpenID Connect ID Token.
+
+
+## Facebook
+
```
mystiberry-okta: Do we expect to have customers that will set up all four at once? If not, why are all four on this page? The saml-doc.okta versions are linked to from the product, are we planning to link to these instead?

Also, @johngronberg-okta we never decided how we would get out of documenting other people's constantly changing UIs. Jakub is updating all four here, but we'll still have the problem of what to do the next time someone tweaks their UI.

_source/_docs/api/resources/social_authentication.md
mystiberry-okta: Also, Social Login is a product name of Auth0. Or their doc makes it look like it is.

_source/_docs/api/resources/social_authentication.md
```diff
@@ -6,127 +6,551 @@ excerpt: Setting up an Okta Social Authentication provider
...
-    `https://example.okta.com/oauth2/v1/authorize?idp=0oa0bcde12fghiJkl3m4&client_id={clientId}&response_type={responseType}&response_mode={responseMode}&scope={scopes}&redirect_uri={redirectUri}&state={state}&nonce={nonce}`
+**Auth.js**
 
-6.  Register an OAuth client using the App Integration Wizard.
+If you don't want pre-built views, or need deeper levels of customization, then you can use the same AuthJS SDK that the Sign-in Widget is built with. For further information see [the AuthJS GitHub repo](https://github.com/okta/okta-auth-js#install). Implementing login with Microsoft would use the SDK's [OpenID Connect authentication flow](https://github.com/okta/okta-auth-js#openid-connect-options).
 
-    1. Navigate to the Administrator Dashboard.
 
-    2. Select *Applications*.
+# Identity Provider Settings <a name="idp-settings"></a>
```
mystiberry-okta: Is this section a conceptual section (explaining terms), a set of instructions for things I must do, should do, can do, and would I want this info BEFORE setting up social authentication? Is there BOTH a social IdP and an IdP or just one?

_source/_docs/api/resources/social_authentication.md
```diff
@@ -6,127 +6,551 @@ excerpt: Setting up an Okta Social Authentication provider
...
+
+- **Support for Webhooks:** Make callouts to your application during user registration and/or account linking. For example: prompt a user for additional attributes before linking their account, or enable a customized validation or confirmation flow.
+
+- **OAuth Scope Configuration:** Specify OAuth2 scopes to fully control which Social Profile attributes are linked to Okta. 
+
+### How it Works
+
+#### Set-up
+
+To set-up social login, you configure the following:
```
mystiberry-okta: no hyphen when a verb "to set up social authentication" (why are we using social login? if we are going to use the same term as Auth0 and social login=social auth, shouldn't we use just the one term everywhere?)

_source/_docs/api/resources/social_authentication.md
```diff
@@ -6,127 +6,551 @@ excerpt: Setting up an Okta Social Authentication provider
...
+
+## Facebook
+
+##### 1. Set Up a Facebook App 
+
+1.1. Go to <https://developers.facebook.com> and register for a developer account if you haven't already done so.
+
+1.2. Head over to the Facebook App Dashboard: <https://developers.facebook.com/apps>
+
+1.3. Create a Facebook app. Instructions for creating a Facebook application can be found here: <https://developers.facebook.com/docs/apps/register>
```
danielzeller-okta: I know that there's been discussion about this before but I don't know what the resolution was:
1. To what extent do we want to be providing instructions on how to perform actions on the social provider's side of things
2. How are we going to ensure that this document stays up to date with their process? If they were to change their url we've suddenly got a broken link in our docs.

I agree that this is a good balance, but wanted to call this out

_source/_docs/api/resources/social_authentication.md
```diff
@@ -6,127 +6,551 @@ excerpt: Setting up an Okta Social Authentication provider
...
+2.3. Hover your cursor over **Security** until the menu opens, then click on **Identity Providers**.
+
+2.4. On the Identity Providers page, click on **Add Identity Provider** > **Add Facebook**
+
+- **Name:** We suggest using the name you would expect to see on a button, something like “Log in to Facebook”.
+- **IdP Username:** Set to “idpuser.email”.
+- **Match against:** Leave set to the default.
+- **Account Link Policy:** Leave set to “Automatic” for now.
+- **Auto-Link Restrictions:** Leave set to the default.
+- **Provisioning Policy:** Leave set to “Automatic” for now.
```
danielzeller-okta: nit: why say default for 2 of the last 4 items, but specify the default option (Automatic) for two?

_source/_docs/api/resources/social_authentication.md
```diff
@@ -6,127 +6,551 @@ excerpt: Setting up an Okta Social Authentication provider
...
+
+4.1. Back on the Okta Admin site, click on **Applications**
+
+4.2. Click **Add Application**
+
+4.3. On the “Add Application” page, click **Create New App**
+
+4.4. In the modal that pops up, select the appropriate platform for your use case and choose OpenID Connect as your sign-on method. 
+
+4.5. Add a name of your choosing, and optionally a logo, then click **Next**.
```
danielzeller-okta: clicking Next is no longer needed as of next week, it's turning into a one-page modal

_source/_docs/api/resources/social_authentication.md
```diff
@@ -6,127 +6,551 @@ excerpt: Setting up an Okta Social Authentication provider
...
+
+3.1. In your Google API Manager Credentials page, select the OAuth client you just created.
+
+3.2. In the "Restrictions" section, find the "Authorized redirect URIs"
+
+3.3. Paste in the Redirect URI from step 2.6 above.
+
+3.5. Click **Save**.
+
+##### 4. Register an OAuth Application in Okta
```
danielzeller-okta: Is it really necessary to give instructions on how to create an oauth client for each social provider?

eklein-okta:
Nit: There are missing periods at the end of several steps; for example, 3.2. In the "Restrictions" section, find the "Authorized redirect URIs" should end with "...redirect URIs."

_source/_docs/api/resources/social_authentication.md
```diff
@@ -6,127 +6,551 @@ excerpt: Setting up an Okta Social Authentication provider
 
 # Social Authentication
 
-Okta Social Authentication allows Okta to integrate with a Social Identity Provider such as Facebook, Google, LinkedIn, and Microsoft. These integrations support the following features:
+---
+layout: docs_page
+title: Social Authentication
+excerpt: Setting up an Okta Social Authentication provider
```
larsjohansen-okta: image
This is how it looks now, not sure what the intentions were here, but we can't have it like that.

_source/_docs/api/resources/social_authentication.md
larsjohansen-okta: I generally agree with Mysti here, however I don't think implementing is a better word than integrating, but that's just my opinion.

_source/_docs/api/resources/social_authentication.md
```diff
@@ -6,127 +6,551 @@ excerpt: Setting up an Okta Social Authentication provider
...
+---
+
+# Social Authentication
```
larsjohansen-okta: I think we should be consistent with calling it either Social Login or Social Authentication, and not mix those two.

_source/_docs/api/resources/social_authentication.md
larsjohansen-okta: This feature is called profile mastering, and certain apps [read Active Directory / LDAP and IdPs] have the opt-in ability to be the source of truth for a user's profile attributes. If a user is assigned multiple apps with profile mastering enabled, a prioritization in Directory -> Profile Masters will decide which app is the source of truth for the user's attributes. All I could find regarding this feature was here: https://developer.okta.com/doc_reviews/In_Product_Help/Content/Topics/Directory/Directory_Profile_Masters.htm

Regarding this short sentence here, I think it's ok.

_source/_docs/api/resources/social_authentication.md
```diff
@@ -6,127 +6,551 @@ excerpt: Setting up an Okta Social Authentication provider
...
+
+Integrating social login with Okta allows you to use the following features:
+
+- **User Registration:** Capture the Profile attributes from a Social Identity Provider user and store those attributes in Okta's Universal Directory.
+
+- **User Authentication:** After a user is registered, continue to use that Social Identity Provider for user authentication, thus eliminating the need to store an additional username and password for that user.
+
+- **Social Profile Sync:** If a user updates their Social Profile, those changes can be reflected inside Okta the next time they use social login.
+
+- **Support for Multiple Social Profiles** Multiple Social Profiles can all be linked to one master Okta user.
```
larsjohansen-okta: I would just call it "one Okta user", as there's a concept of a Okta-mastered user which has a different meaning.

_source/_docs/api/resources/social_authentication.md
larsjohansen-okta: Question: Should we call it OAuth 2.0 Scope Configuration instead of OAuth Scope Configuration, since we're trying to be consistent with using the term OAuth 2.0 and not OAuth.

_source/_docs/api/resources/social_authentication.md
```diff
@@ -6,127 +6,551 @@ excerpt: Setting up an Okta Social Authentication provider
...
+
+- **OAuth Scope Configuration:** Specify OAuth2 scopes to fully control which Social Profile attributes are linked to Okta. 
+
+### How it Works
+
+#### Set-up
+
+To set-up social login, you configure the following: 
+
+1. An OAuth client in your social provider
```
larsjohansen-okta: OAuth -> OAuth 2.0 ?

_source/_docs/api/resources/social_authentication.md
```diff
@@ -6,127 +6,551 @@ excerpt: Setting up an Okta Social Authentication provider
...
+2. An Identity Provider in Okta
+3. An OAuth Application in Okta
```
larsjohansen-okta: Same here

_source/_docs/api/resources/social_authentication.md
larsjohansen-okta: It would be wrong to call it three Users in Okta; it is only one user and two app users which we talk about other places in our docs, so the audience might be familiar with that term. I would probably have made the whole section slightly shorter:

```
Every Identity Provider in Okta is linked to an Application. Every time a user signs in with a Social Identity Provider for the first time, an Application User is created for them. An Application User represents the external user at the Social Identity Provider and can be used to map attributes to the Okta User. For more information about how to configure this behavior see [Identity Provider Settings](#idp-settings).
```

_source/_docs/api/resources/social_authentication.md
larsjohansen-okta: I don't know whether we should put the focus to the ID Token, as there's nothing preventing the admins to configure this whole flow to eventually return an access token instead. I think I would leave out " with a OpenID Connect ID Token".

_source/_docs/api/resources/social_authentication.md
```diff
@@ -6,127 +6,551 @@ excerpt: Setting up an Okta Social Authentication provider
...
+1. The user who wishes to authenticate will click a “Log in with x” link.
+2. The user authenticates and is asked by the Provider to accept the permissions required by your app.
+3. Once the user accepts the permissions, Okta handles the login and redirects the user back to your specified redirect URI with a OpenID Connect ID Token.
+
+
+## Facebook
+
+##### 1. Set Up a Facebook App 
+
+1.1. Go to <https://developers.facebook.com> and register for a developer account if you haven't already done so.
```
larsjohansen-okta: Clicking on this link takes me away from the current page, can we make it open a new tab?

_source/_docs/api/resources/social_authentication.md
```diff
@@ -6,127 +6,551 @@ excerpt: Setting up an Okta Social Authentication provider
...
+
+1.2. Head over to the Facebook App Dashboard: <https://developers.facebook.com/apps>
```
larsjohansen-okta: Same here, and generally applies to all the links used in this fashion.

_source/_docs/api/resources/social_authentication.md
```diff
@@ -6,127 +6,551 @@ excerpt: Setting up an Okta Social Authentication provider
...
+2.1. Sign in to your Okta org.
+
+2.2. On the main page, click on your name in the upper right and then click on **Admin**. 
+
+2.3. Hover your cursor over **Security** until the menu opens, then click on **Identity Providers**.
+
+2.4. On the Identity Providers page, click on **Add Identity Provider** > **Add Facebook**
+
+- **Name:** We suggest using the name you would expect to see on a button, something like “Log in to Facebook”.
+- **IdP Username:** Set to “idpuser.email”.
```
larsjohansen-okta: Do we want to explain to them that they can use Okta EL expression freely here?

_source/_docs/api/resources/social_authentication.md
```diff
@@ -6,127 +6,551 @@ excerpt: Setting up an Okta Social Authentication provider
...
+- **Match against:** Leave set to the default.
+- **Account Link Policy:** Leave set to “Automatic” for now.
```
larsjohansen-okta: "for now" makes me think that we'll make them change this setup, hmm

_source/_docs/api/resources/social_authentication.md
```diff
@@ -6,127 +6,551 @@ excerpt: Setting up an Okta Social Authentication provider
...
+- **Auto-Link Restrictions:** Leave set to the default.
+- **Provisioning Policy:** Leave set to “Automatic” for now.
+- **Profile Master:** Leave unchecked if you want to be able to edit your user information in Okta, rather than having Facebook be the only source for all user updates.
```
larsjohansen-okta: An important aspect of this is that the profile attributes will not be updated even if they are updated in Facebook once we have created the user in Okta unless this box is checked.

_source/_docs/api/resources/social_authentication.md
```diff
@@ -6,127 +6,551 @@ excerpt: Setting up an Okta Social Authentication provider
...
+
+##### 4. Register an OAuth Application in Okta
+
+4.1. Back on the Okta Admin site, click on **Applications**
+
+4.2. Click **Add Application**
+
+4.3. On the “Add Application” page, click **Create New App**
+
+4.4. In the modal that pops up, select the appropriate platform for your use case and choose OpenID Connect as your sign-on method.
```
larsjohansen-okta: Is it intuitive for the admins what use case they're actually interested in, or do we need to provide them a bit more guidance here?

_source/_docs/api/resources/social_authentication.md
```diff
@@ -6,127 +6,551 @@ excerpt: Setting up an Okta Social Authentication provider
...
+
+4.3. On the “Add Application” page, click **Create New App**
+
+4.4. In the modal that pops up, select the appropriate platform for your use case and choose OpenID Connect as your sign-on method. 
+
+4.5. Add a name of your choosing, and optionally a logo, then click **Next**. 
+
+4.6. Add one or more Redirect URIs. This is where the user will be directed to after they have authenticated with Facebook.
+
+4.7. Click **Finish** and you will arrive on the page for your new Facebook application.
```
larsjohansen-okta: This can quickly be confused with the application user's mentioned earlier, which are _not_ related to this app, but to the application created from the IdP. I would maybe just remove the name "Facebook" from that sentence.

_source/_docs/api/resources/social_authentication.md
```diff
@@ -6,127 +6,551 @@ excerpt: Setting up an Okta Social Authentication provider
...
+
+4.5. Add a name of your choosing, and optionally a logo, then click **Next**. 
+
+4.6. Add one or more Redirect URIs. This is where the user will be directed to after they have authenticated with Facebook.
+
+4.7. Click **Finish** and you will arrive on the page for your new Facebook application. 
+
+4.8. Beside "General Settings", click on **Edit**
+
+4.9. Enable the "Implicit (Hybrid)" grant type and click **Save**.
```
larsjohansen-okta: Step 4.8 and 4.9 is not needed (in fact cannot be done) for Single Page App applications (ref. the platform for their use case mentioned in 4.4.

_source/_docs/api/resources/social_authentication.md
```diff
@@ -6,127 +6,551 @@ excerpt: Setting up an Okta Social Authentication provider
...
+
+4.10. In the "Client Credentials" section, copy your "Client ID", which you will use to complete your Authorize URL in the next step.
+
+4.11. Click on the **Assignments** tab.
+
+4.12. Click **Assign** > **Assign to Groups** 
+
+4.13. Assign the group "Everyone", then click **Done**.
```
larsjohansen-okta: I believe it should be enough to assign it to any of the groups that was selected in step 2.4#Group Assignments. If that was left to default you need to assign to the Everyone group as described.

_source/_docs/api/resources/social_authentication.md
```diff
@@ -6,127 +6,551 @@ excerpt: Setting up an Okta Social Authentication provider
...
+4.13. Assign the group "Everyone", then click **Done**. 
+
+##### 5. Complete Your Authorize URL
+
+The Okta Identity Provider that you created in section 2 above generated an Authorize URL with a number of blank parameters that you must now fill-in:
+
+- **client_id:** use the client_id value you copied in step 4.10.
+- **scope:** Determines the claims that are returned in the ID token. This should have at least `openid`.
+- **response_type:** Determines which flow is used. This should be `id_token`.
+- **response_mode:** Determines how the authorization response should be returned.
```
larsjohansen-okta: Since you're providing a suggestion in the step above, I would provide "This should be `fragment`." here.

_source/_docs/api/resources/social_authentication.md
```diff
@@ -6,127 +6,551 @@ excerpt: Setting up an Okta Social Authentication provider
...
+- **state:** Protects against cross-site request forgery (CSRF).
+- **nonce:** A string included in the returned ID Token. Use it to associate a client session with an ID Token, and to mitigate replay attacks.
+- **redirect_url:** The location where Okta returns a user after the user has finished authenticating against their Social Authentication provider. This URL must start with “https” and must match one of the Redirect URIs  that you configured previously.
```
larsjohansen-okta: The location where Okta returns a user after the user -> The location where Okta redirects the browser after the user

larsjohansen-okta: ... configured previously in step 4.5 (which most likely will become 4.4 after you incorporate Daniel's feedback).

This could be confusing with the callback url you had to copy from the IdP configuration, so trying to make it clear that it's from the OpenID Connect Client setup.

_source/_docs/api/resources/social_authentication.md
larsjohansen-okta: To add to that:
- I believe we generally don't use oktapreview.com in examples, but rather an okta.com org.
- We typically url encode parameters, so in your example, we should encode the redirect_uri and not include it non-encoded.

_source/_docs/api/resources/social_authentication.md
```diff
@@ -6,127 +6,551 @@ excerpt: Setting up an Okta Social Authentication provider
...
+
+For a full explanation of all these parameters, see here: [OAuth 2.0 Request parameters](https://developer.okta.com/docs/api/resources/oauth2.html#request-parameters).
+
+An example of a complete URL looks like this: `https://dev-263012.oktapreview.com/oauth2/v1/authorize?idp=0oaaq9pjc2ujmFZexample&client_id=GkGw4K49N4UEE1example&response_type=id_token&response_mode=fragment&scope=openid&redirect_uri=http://yourAppUrlHere.com/social_auth&state=foo&nonce=bar`
+
+#### Invoking Facebook for Login
```
larsjohansen-okta: Is Using a better word than Invoking?

_source/_docs/api/resources/social_authentication.md
```diff
@@ -6,127 +6,551 @@ excerpt: Setting up an Okta Social Authentication provider
...
+
+4.10. In the "Client Credentials" section, copy your "Client ID", which you will use to complete your Authorize URL in the next step.
+
+4.11. Click on the **Assignments** tab.
+
+4.12. Click **Assign** > **Assign to Groups** 
+
+4.13. Assign the group "Everyone", then click **Done**. 
+
+##### 5. Complete Your Authorize URL
```
larsjohansen-okta: Do we have to provide them with the same instructions for this for each provider, even though they are identical?

_source/_docs/api/resources/social_authentication.md
```diff
@@ -6,127 +6,551 @@ excerpt: Setting up an Okta Social Authentication provider
...
-       This URL **must** start with "https" and **must** match one of the URLs in the `redirect_uris` array that you configured previously.
+**Match against:** The Okta user property against which the IdP username is compared. 
 
-     When complete, your Authorize URL looks something like this:
+More user profile attributes are available for matching as an Early Access feature. To enable more choices, contact Support.
 
-    ```http
-    https://example.okta.com/oauth2/v1/authorize?idp=0oa0bcde12fghiJkl3m4&client_id=AbcDE0fGHI1jk2LM34no&scope=openid%20email%20profile&response_type=id_token&response_mode=fragment&state=someState&nonce=someNonce&redirect_uri=https://app.example.com/social_auth
-    ```
+**Account Link Policy:** Determines whether your Application User should be linked to an Okta user
```
larsjohansen-okta: Missing period to end the sentence to be consistent with the other sections.

_source/_docs/api/resources/social_authentication.md
```diff
@@ -6,127 +6,551 @@ excerpt: Setting up an Okta Social Authentication provider
...
 
-     If you log users into Okta, the Authorize URL looks something like this:
+- Automatic: Link user accounts automatically according to the "Auto-Link" and "Match against" settings.
```
larsjohansen-okta: "Auto-Link" -> "Auto-Link Resitrctions"

_source/_docs/api/resources/social_authentication.md
larsjohansen-okta: I would suggesting to replace this section with something similar to:

"If selected, the Social Identity Provider will be the source of truth for a user's profile attributes. This means that next time the user logs in using the Social Identity Provider, Okta will update the user profile attributes for this user. If a user is assigned multiple applications with profile mastering enabled, a prioritization in Directory -> Profile Masters will decide whether this provider will be mastering the user's attributes."

_source/_docs/api/resources/social_authentication.md
jakubvul-okta: I have yet to talk to anyone who can make heads or tails of it, so I would say it hinders rather than helps.

_source/_docs/api/resources/social_authentication.md
jakubvul-okta: Ah, something went funky during the rebase.

_source/_docs/api/resources/social_authentication.md
jakubvul-okta: I prefer login, but as the page is already called "social_authentication" I'll use that.

_source/_docs/api/resources/social_authentication.md
jakubvul-okta: @larsjohansen-okta "Configuration" doesn't seem accurate. Once you've got it all configured, you'll still need to do something to actually allow a user to log in, whether that's adding a link using the Authorize URL or the sign-in widget or something.

_source/_docs/api/resources/social_authentication.md
jakubvul-okta: It is not documented at all, and it's quite a large feature to document, so Tom A said we should backlog it for now.

_source/_docs/api/resources/social_authentication.md
jakubvul-okta: Fair point. I've removed this header, moved "Process" up, and raised the heading level of "Process" and "Set-up".

_source/_docs/api/resources/social_authentication.md
jakubvul-okta: Changed to "The Set-up Process"

_source/_docs/api/resources/social_authentication.md
jakubvul-okta: Unfortunately this is outside of my ability to specify in Markdown. I've brought it up with the DevEx team though, since I think it would be a good addition to the site as a whole.

_source/_docs/api/resources/social_authentication.md
jakubvul-okta: I think we should, as much as possible, outsource the instructions for their side of things.

In my experience, the URLs don't change as often as the UI does, and even if they do they create redirects. Either way it's still a more reliable approach than trying to write the instructions ourselves.

_source/_docs/api/resources/social_authentication.md
jakubvul-okta: Good point. Will remove from all of them.

_source/_docs/api/resources/social_authentication.md
jakubvul-okta: Will make it more specific. (Ex. Leave set to "None".)

_source/_docs/api/resources/social_authentication.md
jakubvul-okta: I've added that to the IDP Settings section at the bottom.

_source/_docs/api/resources/social_authentication.md
jakubvul-okta: Removed Facebook/Google/etc throughout.

_source/_docs/api/resources/social_authentication.md
jakubvul-okta: Changed this: 

```
4.8. Under "General Settings", check if the "Implicit (Hybrid)" grant type is enabled. If it is not already enabled, click **Edit**, enable it, and click **Save**.
```

_source/_docs/api/resources/social_authentication.md
jakubvul-okta: Changed to: 

4.12. Assign the group that you chose under "Group Assignments" in Step 2.4 above. If you did not choose a Group, assign "Everyone", then click **Done**.

_source/_docs/api/resources/social_authentication.md
jakubvul-okta: * @rchild-okta can you give me an example of "more opaque" values?
* @larsjohansen-okta all of my work in Okta so far has used oktapreview, so I'm actually not sure why we don't use it in our examples. 
* Changed the redirect URI so that it's URL encoded

_source/_docs/api/resources/social_authentication.md
jakubvul-okta: The guiding principle here is to reduce clicking around, and to allow for a simple list of step-by-step instructions that can be carefully followed. The more you're clicking around the more likely you are to get confused or miss something. 

Also @larsjohansen-okta.

_source/_docs/api/resources/social_authentication.md
jakubvul-okta: Response above.

_source/_docs/api/resources/social_authentication.md
rchild-okta: @jakubvul-okta in authJs, we just [generate a random 64 character string](https://github.com/okta/okta-auth-js/blob/f162b052efd664dd0db843032c1970dec4639e66/lib/token.js#L270), so here you go:

YsG76jozcsaZGLcFWM6DDGpEJzdkzwRPEFZa7FRDdTW9bBPiXHUDXjfpJwYTVoGx

_source/_docs/api/resources/social_authentication.md
jakubvul-okta: Amended this as follows:

# Social Identity Provider Settings <a name="idp-settings"></a>

When you are setting up your social Identity Provider (IdP) in Okta, there are a number of settings that allow you to finely control the social login behavior. While the provider-specific instructions show one possible configuration, this section explains each of these in more detail so that you can choose the right configuration for your use case.

jakubvul-okta: Also, As to your Qs:

- It's documenting what the different settings mean when you are creating a social IdP. What I discovered is that these were not documented anywhere up to now. I guess that qualifies as "What you can do".
- In the provider-specific instructions I link to this section.
- There is a Social IdP and a SAML IdP.

_source/_docs/api/resources/social_authentication.md
jakubvul-okta: I'll use a truncated version, just for readability's sake.

_source/_docs/api/resources/social_authentication.md
jakubvul-okta: My assumption is that (1) the audience is developers, and (2) the developer knows what they're building here.

jakubvul-okta:
@mystiberry-okta @danielzeller-okta @larsjohansen-okta 

There are some larger structural points that you've brought up, namely:

- The instructions for the different IdPs shouldn't live on the same page
- We shouldn't repeat the OAuth App set-up instructions four times on one page

I think we should let these slide for now, if for no other reason than because we are planning on moving this page out of the reference section (it's not really like any of the other pages in this section) and into a larger Authentication Guide in the near future.

mystiberry-okta:
@jakubvul-okta For the first bullet, if you feel the customers will not have trouble finding their content or navigating the topic, I'm fine with your current structure.

Not sure about the second bullet, I was just talking about moving all the conceptual content into the same area, and before the tasks. But it's a minor point, up to you!

mystiberry-okta:
Also, hoping if you rerun the tests they'll pass. I don't recognize the error it's reporting.

_source/_docs/api/resources/social_authentication.md
larsjohansen-okta: This could be a shared first step, there's no reason you have to do this as step 4, aside from 4.12 which we could simplify to just say assigning the Everyone group. That way you wouldn't need to click around.

_source/_docs/api/resources/social_authentication.md
```diff
@@ -1,132 +1,541 @@
...
+
+If you don't want pre-built views, or need deeper levels of customization, then you can use the same AuthJS SDK that the Sign-in Widget is built with. For further information see [the AuthJS GitHub repo](https://github.com/okta/okta-auth-js#install). Implementing login with Microsoft would use the SDK's [OpenID Connect authentication flow](https://github.com/okta/okta-auth-js#openid-connect-options).
+
+# Social Identity Provider Settings <a name="idp-settings"></a>
+
+When you are setting up your social Identity Provider (IdP) in Okta, there are a number of settings that allow you to finely control the social login behavior. While the provider-specific instructions show one possible configuration, this section explains each of these in more detail so that you can choose the right configuration for your use case.
+
+### Authentication Settings
+
+**IdP username:** The IdP attribute that should map to the Okta username. The dropdown list contains the default value, `idpuser.email`.
```
larsjohansen-okta: The Okta EL expression leveraging IdP attributes in which the IdP username will be constructed. This IdP username will be used for matching Application User to Okta User. The dropdown list contains the default value, `idpuser.email`.

larsjohansen-okta: As a side note, the Okta username (also known as the login field) is not populated with this value, but with the app user's email regardless of the settings. There's one exception to this; we allow the login/username of the Okta user to be customizable using callouts / IdP transaction API.

_source/_docs/api/resources/social_authentication.md
mdeggies: One point of confusion for me, and something that should probably be noted, is the difference between the Okta IDP redirect URI (generated when you create the IDP, in the format https://dev-<ID>.oktapreview.com/oauth2/v1/authorize/callback) and the redirect URI that the end-user is redirected back to after successful authentication (usually something like http://localhost:3000/callbacks/idp, depending on the library they're using).

_source/_docs/api/resources/social_authentication.md
```diff
@@ -1,132 +1,560 @@
...
-    -   [Facebook](http://saml-doc.okta.com/IdentityProvider_Docs/Facebook_Identity_Provider_Setup.html)
-    -   [Google](http://saml-doc.okta.com/IdentityProvider_Docs/Google_Identity_Provider_Setup.html)
-    -   [LinkedIn](http://saml-doc.okta.com/IdentityProvider_Docs/LinkedIn_Identity_Provider_Setup.html)
-    -   [Microsoft](http://saml-doc.okta.com/IdentityProvider_Docs/Microsoft_Identity_Provider_Setup.html)
+To set-up social login, you configure the following: 
 
-4.  Configure your Social Authentication provider:
+1. An OAuth 2.0 client in your social provider
+2. An Identity Provider in Okta
+3. An OAuth 2.0 Application in Okta
```
johngronberg-okta: Should this say an OpenID Connect Application, since that's what we call them in the UI?

jakubvul-okta: Fixed.

_source/_docs/api/resources/social_authentication.md
```diff
@@ -1,132 +1,560 @@
...
+
+The Okta Identity Provider that you created in section 2 above generated an Authorize URL with a number of blank parameters that you must now fill-in:
+
+* **client_id:** use the client_id value you copied in step 4.10.
+* **scope:** Determines the claims that are returned in the ID token. This should have at least `openid`.
+* **response_type:** Determines which flow is used. This should be `id_token`.
+* **response_mode:** Determines how the authorization response should be returned. This should be `fragment`.
+* **state:** Protects against cross-site request forgery (CSRF).
+* **nonce:** A string included in the returned ID Token. Use it to associate a client session with an ID Token, and to mitigate replay attacks.
+* **redirect_url:** The location where Okta returns a browser after the user has finished authenticating against their social login provider. This URL must start with “https” and must match one of the Redirect URIs that you configured previously in step 4.6.
```
larsjohansen-okta: redirect_url -> redirect_uri

_source/_docs/api/resources/social_authentication.md
```diff
@@ -1,132 +1,560 @@
...
+
+For a full explanation of all these parameters, see here: [OAuth 2.0 Request parameters](https://developer.okta.com/docs/api/resources/oauth2.html#request-parameters).
+
+An example of a complete URL looks like this: `https://dev-263012.oktapreview.com/oauth2/v1/authorize?idp=0oaaq9pjc2ujmFZexample&client_id=GkGw4K49N4UEE1example&response_type=id_token&response_mode=fragment&scope=openid&redirect_uri=http%3A%2F%2FyourAppUrlHere.com%2Fsocial_auth&state=WM6D&nonce=YsG76jo`
```
larsjohansen-okta: I don't know what our preferences are for examples, but using a real org and real idp id etc might be a bit odd if it is not how we show examples other places.

_source/_docs/api/resources/social_authentication.md
```diff
@@ -1,132 +1,560 @@
...
+
+The Okta Identity Provider that you created in section 2 above generated an Authorize URL with a number of blank parameters that you must now fill-in:
+
+* **client_id:** use the client_id value you copied in step 4.10.
+* **scope:** Determines the claims that are returned in the ID token. This should have at least `openid`.
+* **response_type:** Determines which flow is used. This should be `id_token`.
+* **response_mode:** Determines how the authorization response should be returned. This should be `fragment`.
+* **state:** Protects against cross-site request forgery (CSRF).
+* **nonce:** A string included in the returned ID Token. Use it to associate a client session with an ID Token, and to mitigate replay attacks.
+* **redirect_url:** The location where Okta returns a browser after the user has finished authenticating against their social login provider. This URL must start with “https” and must match one of the Redirect URIs that you configured previously in step 4.6.
```
larsjohansen-okta: redirect_url -> redirect_uri

_source/_docs/api/resources/social_authentication.md
```diff
@@ -1,132 +1,560 @@
...
```
larsjohansen-okta: redirect_url -> redirect_uri

_source/_docs/api/resources/social_authentication.md
```diff
@@ -1,132 +1,560 @@
...
```
larsjohansen-okta: redirect_url -> redirect_uri

_source/_docs/api/resources/social_authentication.md
```diff
@@ -109,71 +103,69 @@ In general, the social login process works as follows:
...
-- **redirect_url:** The location where Okta returns a user after the user has finished authenticating against their Social Authentication provider. This URL must start with “https” and must match one of the Redirect URIs  that you configured previously.
+* **client_id:** use the client_id value you copied in step 4.10.
+* **scope:** Determines the claims that are returned in the ID token. This should have at least `openid`.
+* **response_type:** Determines which flow is used. This should be `id_token`.
+* **response_mode:** Determines how the authorization response should be returned. This should be `fragment`.
+* **state:** Protects against cross-site request forgery (CSRF).
+* **nonce:** A string included in the returned ID Token. Use it to associate a client session with an ID Token, and to mitigate replay attacks.
+* **redirect_url:** The location where Okta returns a browser after the user has finished authenticating against their social login provider. This URL must start with “https” and must match one of the Redirect URIs that you configured previously in step 4.6.
```
jmelberg-okta: @larsjohansen-okta: Do we only support social auth web flows? Mobile redirects use a private-use URI scheme that often looks like `com.okta.example:/callback`.

jakubvul-okta: @jmelberg-okta right now we're only covering web flows.

_source/_docs/api/resources/social_authentication.md
```diff
@@ -1,132 +1,562 @@
...
+
+If you don't want pre-built views, or need deeper levels of customization, then you can use the same AuthJS SDK that the Sign-in Widget is built with. For further information see [the AuthJS GitHub repo](https://github.com/okta/okta-auth-js#install). Implementing login with Microsoft would use the SDK's [OpenID Connect authentication flow](https://github.com/okta/okta-auth-js#openid-connect-options).
+
+# Social Identity Provider Settings <a name="social-idp-settings"></a>
+
+When you are setting up your social Identity Provider (IdP) in Okta, there are a number of settings that allow you to finely control the social login behavior. While the provider-specific instructions show one possible configuration, this section explains each of these in more detail so that you can choose the right configuration for your use case.
+
+### Authentication Settings
+
+**IdP username:** This is the expression (written in the Okta Expression Language) that will be used to convert an IdP attribute to the Application user's `username`. This IdP username will be used for matching an Application User to an Okta User.
```
larsjohansen-okta: Application user's -> Application User's (to be consistent with the sentence that comes after)

_source/_docs/api/resources/social_authentication.md
```diff
@@ -1,132 +1,562 @@
...
+
+# Social Identity Provider Settings <a name="social-idp-settings"></a>
+
+When you are setting up your social Identity Provider (IdP) in Okta, there are a number of settings that allow you to finely control the social login behavior. While the provider-specific instructions show one possible configuration, this section explains each of these in more detail so that you can choose the right configuration for your use case.
+
+### Authentication Settings
+
+**IdP username:** This is the expression (written in the Okta Expression Language) that will be used to convert an IdP attribute to the Application user's `username`. This IdP username will be used for matching an Application User to an Okta User. 
+
+For example, the value `idpuser.email` means that it takes the takes the `email` attribute passed by the social IdP and maps it to the Okta Application user's `username` property
```
larsjohansen-okta: Shouldn't this end with a period?

_source/_docs/api/resources/social_authentication.md
jmelberg-okta: @jakubvul-okta: This was more of a product question. Curious if we support it @larsjohansen-okta.

PR #967, opened 2017-06-20
Updated javadoc location to new okta/okta-sdk-java : gh-pages location

bdemers:
Updates javadoc locations to https://developer.okta.com/okta-sdk-java/0.0.4/apidocs/
This is a republication of the javadocs already found at https://developer.okta.com/docs/sdk/core/java_api_sdk/
The difference being the version has been corrected to NOT be a snapshot.

**NOTE:** this does NOT add the new alpha version.  I not sure of a way to do this and keep it clear that this is new version does not include all of the existing features that are in the 0.0.4 version.

Ideas wanted.

## Description:

## Type of Pull Request:

- [ ] Blog Post

### Resolves:

## How Has This Been Tested?
- [ ] I have built this locally and verified that it does not break existing functionality.

## Primary Reviewers:

- *Tag reviewer(s)*

- [Blog Post Google Doc](https://docs.google.com)


bdemers:
@rchild-okta does the dead link checking only work against local files?
the check failed but the link is valid: https://developer.okta.com/okta-sdk-java/0.0.4/apidocs/
I'm guessing I could updated the href to be fully qualified, but is it possible to check locally, and fall back to `https://developer.okta.com` ?

bdemers:
@rchild-okta  Also, I'm fine with waiting, it just sounded like this was _desired_ for the current sprint. 
If we don't want to update the javadoc version the only benefit here is the updating the href to the new location.
Also, we _should_ delete the old generated (incorrect) javadoc at the same time as this, and add rewrite rules (though I'm not sure how to handle that)
Otherwise we can just push this whole issue off until we have a plan for that

rchild-okta:
> dead link checking only work against local files?

Yes, it does a build and runs it against that. Mind checking to see if there's a way to exclude the link? We're using [htmlproofer](https://github.com/gjtorikian/html-proofer), and it's just being run with cli args [here](https://github.com/okta/okta.github.io/blob/master/scripts/travis.sh#L27). Else, would need to do the external link option.

> like this was desired for the current sprint

I'm forgetting now what we had said in the standup - I *think* it was along the lines of just sticking it on the code page, but looking at the PR now it looks like we don't have very much customization options there (can only write it out as part of the sentence).

An option:
- Add a new frontmatter option (like github_url and excerpt) for sdk documentation, and then update the template to loop over it around the same section as the github urls are being displayed. 
- Re the main question of how to tell users that it doesn't have everything - is labeling it as "beta/alpha/whatever" sufficient?

rchild-okta:
Re delete old & rewrite - hmm, good point. There's a jekyll redirectTo plugin (i.e. [here](https://github.com/okta/okta.github.io/blob/83813af796b6c62ca481e456499fd73312c25ace/_source/_docs/index.md#L2)), but wouldn't be fun to do that for all the pages. Also, these pages aren't going through the build - would need to write a script to replace them with the content (i.e. javascript that does a redirect).

We're trying to solve a related problem (the redirects to okta.github.io) with aws edge lambda, maybe this is something we can try to fix with that as well?

bdemers:
We could also create rewrite rules in cloudflare, though it seems this should be handled before that (ideally)

bdemers:
Re alpha/beta.  I don't think simple tag is sufficient here, as the functionality is different between the two.

bdemers:
re re-writes, no rush then, we can just create a new issue to track these
rewrites

bdemers:
Replaced with #1189

PR #980, opened 2017-06-23
OKTA-122113-doc

richardmateosian-okta:
## Description:
Add Cache-Control example to header.

## Type of Pull Request:

- [ ] Blog Post

### Resolves:
https://oktainc.atlassian.net/browse/OKTA-122113

## How Has This Been Tested?

## Primary Reviewers:
@mystiberry-okta 
@larsjohansen-okta 
@johngronberg-okta

### DIFF
--- a/_source/_docs/api/resources/oidc.md
+++ b/_source/_docs/api/resources/oidc.md
@@ -556,7 +556,6 @@ If automatic key rotation is disabled, provide the *client_id* to fetch public k
 ~~~http
 HTTP/1.1 200 OK
 Cache-Control → max-age=3832304, must-revalidate
-. . .
 Content-Type: application/json;charset=UTF-8
 ~~~
 ~~~json
@@ -601,9 +600,7 @@ Content-Type: application/json;charset=UTF-8
 
 >Okta strongly recommends retrieving keys dynamically with the JWKS published in the discovery document.
 
->Okta also recommends cacheing or persisting downloaded keys for performance.
-However, if the client application is pinned to a signing key, the verification might fail because Okta rotates the key automatically.
-Pinned client applications must periodically check the Okta signing keys.
+>Okta also recommends caching or persisting downloaded keys to improve performance by eliminating multiple API requests. If the client application is pinned to a signing key, the verification fails when Okta rotates the key automatically. Pinned client applications must periodically check the cached Okta signing keys.
 
 Any of the two or three keys listed are used to sign tokens. The order of keys in the result doesn't indicate which keys are used.
 
--- a/_source/_docs/api/resources/oidc.md
+++ b/_source/_docs/api/resources/oidc.md
@@ -553,6 +553,12 @@ If automatic key rotation is disabled, provide the *client_id* to fetch public k
 #### Response Example
 {:.api .api-response .api-response-example}
 
+~~~http
+HTTP/1.1 200 OK
+Cache-Control → max-age=3832304, must-revalidate
+. . .
+Content-Type: application/json;charset=UTF-8
+~~~
 ~~~json
 {
   "keys": [
@@ -594,8 +600,9 @@ If automatic key rotation is disabled, provide the *client_id* to fetch public k
 ~~~
 
 >Okta strongly recommends retrieving keys dynamically with the JWKS published in the discovery document.
-It is safe to cache or persist downloaded keys for performance.
-However, if the client application is pinned to a signing key, the verification might fail since Okta rotates the key automatically.
+
+>Okta also recommends cacheing or persisting downloaded keys for performance.
+However, if the client application is pinned to a signing key, the verification might fail because Okta rotates the key automatically.
 Pinned client applications must periodically check the Okta signing keys.
 
 Any of the two or three keys listed are used to sign tokens. The order of keys in the result doesn't indicate which keys are used.


_source/_docs/api/resources/oidc.md
```diff
@@ -575,8 +581,9 @@ If automatic key rotation is disabled, provide the *client_id* to fetch public k
 ~~~
 
 >Okta strongly recommends retrieving keys dynamically with the JWKS published in the discovery document.
-It is safe to cache or persist downloaded keys for performance.
-However, if the client application is pinned to a signing key, the verification might fail since Okta rotates the key automatically.
+
+>Okta also recommends cacheing or persisting downloaded keys for performance.
+However, if the client application is pinned to a signing key, the verification might fail because Okta rotates the key automatically.
```
mystiberry-okta: This is very confusing. Why do we recommend something that might fail?  I don't think it's helpful to change to 'we also recommend'.  What about something like:

Cache or persist your downloaded keys for improved performance (performance of what? we're a cloud app), unless the client application is pinned to a signing key. The verification sometimes fails because Okta rotates the key automatically.

Check w/your SME too.

_source/_docs/api/resources/oidc.md
```diff
@@ -534,6 +534,12 @@ If automatic key rotation is disabled, provide the *client_id* to fetch public k
 #### Response Example
 {:.api .api-response .api-response-example}
 
+~~~http
+HTTP/1.1 200 OK
+Cache-Control → max-age=3832304, must-revalidate
+. . .
+Content-Type: application/json;charset=UTF-8
```
mystiberry-okta: I don't think an ellipsis is helpful. Can we have the full response header here?

_source/_docs/api/resources/oidc.md
```diff
@@ -575,8 +581,9 @@ If automatic key rotation is disabled, provide the *client_id* to fetch public k
 ~~~
 
 >Okta strongly recommends retrieving keys dynamically with the JWKS published in the discovery document.
-It is safe to cache or persist downloaded keys for performance.
-However, if the client application is pinned to a signing key, the verification might fail since Okta rotates the key automatically.
+
+>Okta also recommends cacheing or persisting downloaded keys for performance.
```
larsjohansen-okta: The more common spelling of "cacheing" is "caching".

_source/_docs/api/resources/oidc.md
larsjohansen-okta: We usually only include relevant headers in responses / requests for examples without using an ellipsis, so I think we should not be using it in this example as well.

_source/_docs/api/resources/oidc.md
richardmateosian-okta: OK. I'll go with that one.

_source/_docs/api/resources/oidc.md
richardmateosian-okta: @larsjohansen-okta Is the improved performance the time saved by not hitting the /keys endpoint again? Is there anything else?

_source/_docs/api/resources/oidc.md
richardmateosian-okta: Ellipsis begone!

_source/_docs/api/resources/oidc.md
larsjohansen-okta: @richardmateosian-okta yes, the performance gain is that you'll save time and resources on not hitting the endpoint frequently. My understanding is that if the application is pinned to a key, it will not be rotated and hence it is safe to cache, however if the client application is _not_ pinned to a key, then Okta will rotate the key on a regular schedule (not sure if we want to/have documented the schedule) but I believe it is 4 times a year. I would not mention it unless we've mentioned it other places in our docs.

richardmateosian-okta: I've updated the text without mentioning the rotation schedule.

_source/_docs/api/resources/oauth2.md
```diff
@@ -469,7 +469,7 @@ Content-Type: application/json;charset=UTF-8
 
 {% api_operation post /oauth2/:authorizationServerId/v1/revoke %}
 
-The API takes an Access Token or Refresh Token and revokes it. Revoked tokens are considered inactive at the introspection endpoint. A client may only revoke its own tokens.
+This API takes an Access Token or Refresh Token and revokes it. Revoked tokens are considered inactive at the introspection endpoint. A client can revoke only its own tokens.
```
cwu-okta: This change is not necessary for this Jira. 

Can we not change tone in this PR? The previous one was always to use "may" and that was true in both the OIDC and OAuth developer pages. This will exasperate the mixture of "can" and "may" I'm starting to see throughout this page and the OIDC one. I would like platform doc to discuss what tone they want to use for the developer site then make a holistic change throughout the site. 
@mystiberry-okta

_source/_docs/api/resources/oidc.md
```diff
@@ -575,9 +580,8 @@ If automatic key rotation is disabled, provide the *client_id* to fetch public k
 ~~~
 
 >Okta strongly recommends retrieving keys dynamically with the JWKS published in the discovery document.
-It is safe to cache or persist downloaded keys for performance.
-However, if the client application is pinned to a signing key, the verification might fail since Okta rotates the key automatically.
-Pinned client applications must periodically check the Okta signing keys.
+
+>Okta also recommends caching or persisting downloaded keys to improve performance by eliminating multiple API requests. If the client application is pinned to a signing key, the verification fails when Okta rotates the key automatically. Pinned client applications must periodically check the cached Okta signing keys.
```
cwu-okta: @mystiberry-okta FYI since this looks very similar to work you are doing.

_source/_docs/api/resources/schemas.md
```diff
@@ -757,19 +757,19 @@ The [User Model](./users.html#user-model) schema is defined using [JSON Schema D
...
+| ------------- | ---------------------------------------- | ------------------------------------------------------ |--------- | ------ | -------- | --------- | --------- | ----------  |
+| id            | URI of user schema                       | String                                                 | FALSE    | TRUE   | TRUE     |           |           |             |
+| $schema       | JSON Schema version identifier           | String                                                 | FALSE    | FALSE  | TRUE     |           |           |             |
+| name          | name for the schema                      | `user`                                                 | FALSE    | TRUE   | TRUE     |           |           |             |
+| title         | user-defined display name for the schema | String                                                 | FALSE    | FALSE  | FALSE    |           |           |             |
+| created       | timestamp when schema was created        | [ISO 8601 String](https://tools.ietf.org/html/rfc3339) | FALSE    | FALSE  | TRUE     |           |           |             |
```
cwu-okta: Why are we adding a reference to the rfc here but nowhere else throughout the site?

mystiberry-okta:
@richardmateosian-okta All the PRs started in weekly but not finished when we do the weekly cleanup get extra commits. Try git rebase -i origin/weekly and remove the irrelevant commits. That way you won't have to resolve the conflict.

PR #1023, opened 2017-07-06
Updates PR template to be less noisy

jmelberg-okta:
## Description:
- Updates the PR template to be less noisy and ease external contribution experience.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -4,5 +4,6 @@
 
 ### Resolves:
 * [Issue #X](#X)
-* [OKTA-XXXXXX](https://oktainc.atlassian.net/browse/OKTA-XXXXXX) <!-- Required for Okta-generated PRs -->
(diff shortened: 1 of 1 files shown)


rchild-okta:
So it is a good question whether to leave the jira link in there or not. On the one hand, it is a very good reminder (I do like that in other templates). On the other hand, it defeats one of the main purposes of the PR, which is to make this feel like a process that external contributors are encouraged to submit - a jira number is something that they won't have.

jmelberg-okta:
@rchild-okta: Good point. Most OS repos I've seen that include a "Resolves" section, reference an issue in the Issues tab.

We could consider replacing the JIRA link with a GH Issues bullet, but I think that is a much larger discussion around changing our current workflow. We *could* have the first bullet be a GH Issue link, and the second (indicated by comments) reference a JIRA.

So this is what the contributer would see:

```markdown
## Description:

### Resolves:
* [Issue # X](#X)
<!--- UNCOMMENT this line to reference a JIRA
--> 

```

PR #1031, opened 2017-07-10
Improve d.o.c tests

vijetmahabaleshwar-okta:
## Description:
1. Create a test page to be used for selenium tests
2. Remove hard coded text checks in selenium tests

### Resolves:
* [Issue #X](#X)

* [OKTA-132585](https://oktainc.atlassian.net/browse/OKTA-132585)

### DIFF
--- a/_source/test_page/index.md
+++ b/_source/test_page/index.md
@@ -1,0 +1,40 @@
+---
+layout: docs_page
+weight: 1
+title: Test Page
+---
+
+# Overview
+
+This is a Test Page
+
+## First Section
+
+1. Step 1
+1. Step 2
+1. Step 3
+
+## Second Section
+
+Postman Test
+[![Run in Postman](https://run.pstmn.io/button.svg)](https://app.getpostman.com/run-collection/1755573c5cf5fbf7968b){:target="_blank"}
+
+> Note: This is a Test Note.
+
+Jump to Third Section - [Third Section](#third-section).
+
+## Third Section
+
+This is the third section
+
+## Last Section
+
+End of Main Sections
+
+### Sub Section 1
+
+This is a sub section
+
+### Sub Section 2
+
+This is the final section
--- a/tests/selenium/framework/page-objects/BasePage.js
+++ b/tests/selenium/framework/page-objects/BasePage.js
@@ -52,22 +52,7 @@ class BasePage {
       return elementList.length == expectedTextArray.length;
     })
   }
-
-  elementHasLinks(element, expectedTextArray) {
-    if (!Array.isArray(expectedTextArray)) {
-      expectedTextArray = [expectedTextArray];
-    }
-
-    return element.getText().then((text) => {
-      for(let i = 0; i < expectedTextArray.length; i++) {
-        if (text.indexOf(expectedTextArray[i]) < 0) {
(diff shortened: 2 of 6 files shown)


_source/test_page/index.md
```diff
@@ -0,0 +1,60 @@
...
+
+Jump to Third Section - [Third Section](#third-section).
+
+## Third Section
+
+**Collections Test**
+
+| Collections                               | Click to Run                                                                                                                            |
+|:------------------------------------------|:----------------------------------------------------------------------------------------------------------------------------------------|
+| Authentication                            |   [![Run in Postman](https://run.pstmn.io/button.svg)](https://app.getpostman.com/run-collection/f9684487e584101f25a3){:target="_blank"} |
```
rchild-okta: Probably don't need all this copied over right? If we want a longer page so we can see the behavior when scrolling, I'd suggest just using something like [lorem ipsum](http://www.lipsum.com/).

tests/selenium/framework/page-objects/BasePage.js
```diff
@@ -52,20 +52,12 @@ class BasePage {
...
+  elementHasLinkCount(element, minimumExpectedCount) {
     return element.getText().then((text) => {
-      for(let i = 0; i < expectedTextArray.length; i++) {
-        if (text.indexOf(expectedTextArray[i]) < 0) {
-          return false;
-        }
-      }
-      return true;
-    });
+      const sections = text.split('\n');
```
rchild-okta: I *think* this will probably work, but is dependent on how the code is generated. Might be better to try to get a count of the section elements (where you could add a class or some other identifier if it doesn't already exist).

tests/selenium/framework/page-objects/SideBarPage.js
```diff
@@ -47,16 +29,27 @@ class SideBarPage extends BasePage {
...
-  doesReferenceHaveLinks(links) {
-    return this.elementHasLinks(this.$referenceNav, links);
+  doesAuthenticationReferenceHaveSubLinks() {
+    return this.$$sideBarReferences.count().then(count => count > 0);
   }
-  
-  doesStandardsHaveLinks(links) {
-    return this.elementHasLinks(this.$standardsNav, links);
+
+  doesAPIReferenceHaveSubLinks() {
```
rchild-okta: nit on naming - seems easier to just say "apiReferenceHasLInks" (same applies to other ones in this file)

_source/test_page/index.md
vijetmahabaleshwar-okta: The only reason I kept them is if we ever wanted to add tests for postman collections in the future, we could do that.
Will remove it for now.

tests/selenium/framework/page-objects/SideBarPage.js
vijetmahabaleshwar-okta: Done

tests/selenium/framework/page-objects/BasePage.js
vijetmahabaleshwar-okta: Updated to check for count of `li` elements under each section. If at least 1 `li` element, then we can be sure that links/sub-links are present.

PR #1039, opened 2017-07-11
Social auth timeout doc

mystiberry-okta:
## Description:
- Doc the limit on social auth flows (social auth requests)

### Resolves:

* [OKTA-122142](https://oktainc.atlassian.net/browse/OKTA-122142)

### DIFF
--- a/_source/_docs/api/resources/social_authentication.md
+++ b/_source/_docs/api/resources/social_authentication.md
@@ -28,7 +28,7 @@ Configuring social login with Okta allows you to use the following features:
 
 ### The Social Login Process
 
-The social login process starts at the authorization endpoint, then goes out to the provider and back:
+The social login process starts at the [authentication endpoint](/docs/api/resources/oidc.html#authentication-request), then goes out to the provider and back:
 
 1. The user who wishes to authenticate clicks a “Log in with x” link.
 2. The user authenticates and is asked by the Provider to accept the permissions required by your app.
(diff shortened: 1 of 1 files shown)


_source/_docs/api/resources/social_authentication.md
```diff
@@ -55,9 +55,11 @@ ok -> ua: 302 to redirect_uri
 @enduml
 -->
 
+Social authentication requests can take up to fifteen minutes to complete before they are cancelled.
```
danielzeller-okta: Would we prefer "may take up to..." here rather than "can"? The way this reads to me is that we are expecting that it might take that long, rather than it is allowed to take that long.

_source/_docs/api/resources/social_authentication.md
```diff
@@ -28,7 +28,7 @@ Configuring social login with Okta allows you to use the following features:
 
 ### The Social Login Process
 
-The social login process starts at the auth point, then goes out to the provider and back:
+The social login process starts at the authorization endpoint, then goes out to the provider and back:
```
danielzeller-okta: should we link here?

_source/_docs/api/resources/social_authentication.md
mystiberry-okta: yes, will fix.

_source/_docs/api/resources/social_authentication.md
mystiberry-okta: Added the link, too, because I had to read the illustration below here to be sure with auth point we were talking about. changed it to authentication endpoint, since that's what the doc calls it, even though it's /authorize (that's an historical thing, right?).

danielzeller-okta:
LGTM 🐚

PR #1048, opened 2017-07-13
endpoint for social, not saml

mystiberry-okta:
## Description:
- idps endpoint /api/v1/users/:uid/idps is only for social idps, not saml.
### Resolves:

* [OKTA-132878](https://oktainc.atlassian.net/browse/OKTA-132878)


mystiberry-okta:
@royalchan-okta Please review, thanks!

PR #1056, opened 2017-07-17
Remove generated source & push dist to master

bdemers:
Remove generated code from 'source' branch publish generated code to master

NOTE: this 'idea' branch uses 'source-bdemers' and 'source-master' and does NOT have a GH key configured.

This branch still needs a readme update. but here is the gist of it:
a.) Only generated code gets put into 'master' 
b.) Only managed code gets put into 'source'
c.) Travis builds 'source' and deploys to 'master' on merge (after same code review as we have today)

The big difference is that we will not have any gnarly generated code diffs to review (or confuse the review)

Also the push to 'master' uses Travis publish deploy provider, which is what @bretterer is using for okta/okta-sdk-php

Also, if we stop using gh-pages this process work will still be needed (removing all of the generated cruft from the repo)

Concerns: 
* Conflicts with @rdegges work with Hugo
* Communication across team to use the 'source' branch instead of 'master' (change the default branch / ref will help)


bdemers:
*NOTE:* 6f07c877042e2ca03e128193f5c64ce366c0e463 just deletes the generated code and 748199e22aac35dbfb815ecffcdea12dbba32c3b is the actual changes (minimal) to make this work

ALSO, I'm not attempting to keep this branch up to date.  If we want to move forward with this we can pick a quite time (with few changes merging) and update it.  As is this should be enough for discussion.

jakubvul-okta:
@bdemers I think it's because the old command: `npm run clean && JEKYLL_ENV=production bundle exec jekyll serve --incremental` has been replaced with `npm run clean && && JEKYLL_ENV=production bundle exec jekyll serve --incremental`

Not sure why that second `&&` is in there, and the diff is too massive for me to find the change that caused it.

rchild-okta:
@bdemers I like the intent of the commit, and also agree that it's not great that we have the generated files checked in.

The underlying issue is that we're using github pages to serve the requests (this is a problem for other things as well), and serving it off of master. There's other work going on now from eng-prod to get us to S3, which is the best time to do this - the s3 deploy only depends on a build folder from the published npm artifact.

bdemers:
@rchild-okta I understand the s3 thing, but removing the generated code would need to happen for that anyway. And it is causing issues with reviews,  I think it would be easier to do this now, so we get the benefit immediately.  I've defiantly lost more cycles searching around this build tree and searching through reviews then this PR took to create.

Though I am slightly concerned with the fact that this would require a branch change (forcing everyone to work off of 'source' instead of 'master').  So, if we are moving to S3 before the end of the month I'm all for waiting, otherwise i'd like to push for this.

rchild-okta:
@bdemers there are three problems to resolve before thinking about this:

- Have to go through the existing orphaned files and see if we need to do anything about it (move it to source, add redirects, or just drop). @mystiberry-okta might have an opinion on this.
- Change the process to use source instead of weekly/master. @mystiberry-okta might have more thoughts on this, but it hasn't been seamless. Also need to think about what this means for the weekly flow.
- Github keys being checked into source. The same discussion we've had before in other repos.

mystiberry-okta:
Reread this. The problem for feature work is b. Feature work actually needs a branch, a place to live that's not a PR and not source that will be used to generate updates for master.

bdemers:
Currently we have:
* Master -> live code
* weekly-branched-from-master - staged code merged from PR


Same workflow with this just replace everything you think of as 'master'
with 'source'

So you have:
* Source -> live code
* weekly-branched-from-source - staged code merged from PR

The only change here is replace all of our usage of 'master' with 'source'
and everything is exactly the same

rchild-okta:
@bdemers most of the work is planned to be finished by the end of this month (details [here](https://oktawiki.atlassian.net/wiki/display/EP/CD+with+Conductor+-+Design))

bdemers:
I have one commit with changes, and another removing the dead files which is done with a script (in order to rebase this change)

Full list of files deleted with script:
```text
3rd_party_notices/
404.html
assets/
blog/
cla/
code/
confirm/
contact/
discussion/
docs/
documentation/
error.html
favicon.ico
features/
feed.xml
feed/
google77fbcd2526bf2e7b.html
index.html
integrate_with_okta/
oidc_playground/
pricing-preview/
privacy/
product/
reference/
search/
signup/
sitemap.xml
sitemap/
standards/
terms/
test_page/
use_cases/
```

bdemers:
**NOTE:** just before merged the `bdemers-` from the 'source' and 'master' branches.
Also, this branch will NOT be merged to master, it will just be renamed to 'source' (as master will only contain the generated bits)

bdemers:
Basically up for review is 0290b9b1829c55b226348661861401f85736636e
and the list of files deleted above

mystiberry-okta:
So we haven't got rid of the awkwardness of force pushing source onto weekly after we push all weekly changes to master (where travis will magically build them for us)? Everything looks fine as long as the needed HTML-not-generated files are in master. Thanks!

bdemers:
Already _merged_ into the `source` branch

PR #1078, opened 2017-07-21
OKTA-133829 Add some status codes

richardmateosian-okta:
## Description:
Add status codes to some /clients and /apps response examples
### Resolves:
https://oktainc.atlassian.net/browse/OKTA-133829


mystiberry-okta:
@richardmateosian-okta  I'll review after you rebase to get rid of unwanted commits.

royalchan-okta:
- Can we get the format to match that of OAuth 2.0?
image

- We need it for all example responses in clients, not just the create
- We also need them in the auth server apis, as the creates were added for those too. There wasn't a separate JIRA for it but it will be in the same set of changes.

richardmateosian-okta:
Killing this branch and starting over

PR #1090, opened 2017-07-25
Oidc post image fix

leebrandt:
## Description:
The Redirect URI image in the post had `localhost:5000` for the redirect and it should be `localhost:5000/signin-oidc`. The text was right, just the image was out of date.


mraible:
@jmelberg-okta Can this be merged?

PR #1094, opened 2017-07-26
Okta 120035

richardmateosian-okta:
## Description:
OIDC Flows

### Resolves:
https://oktainc.atlassian.net/browse/OKTA-120035


richardmateosian-okta:
Not yet ready for review. PR is open for convenience.

richardmateosian-okta:
@federations-okta This draft does not have diagrams, but I'd prefer to have feedback on the content before I draw pictures of it. As always, all comments welcome.

mystiberry-okta:
@richardmateosian-okta I don't understand the goals/requirements for this doc well enough to review it except for grammar stuff. Is that your intent?

richardmateosian-okta:
@mystiberry-okta I assume you mean more than grammar when you say "grammar stuff." You always give me good comments on clarity, and you're much clearer about Okta conventions -- written and unwritten -- than I am.

cwu-okta:
The images are not in the style we use for developer.okta.com. Please match the existing style.

mystiberry-okta:
@cwu-okta, Jakub found a new way to do these diagrams that is very fast compared to using PowerPoint. They will match the destination (the auth guide).

richardmateosian-okta:
@cwu-okta That style for flow diagrams is also on D.O.C at https://developer.okta.com/docs/api/resources/social_authentication.html#the-social-login-process

larsjohansen-okta:
@richardmateosian-okta The flow diagrams are showing Okta as multiple entities depending on the endpoint you're using; that is an odd pattern that we should try to avoid. Instead we should indicate which endpoint that we use as description on the arrow itself, like shown in https://developer.okta.com/docs/api/resources/social_authentication.html#the-social-login-process.

There's also significant differences between the diagram on the social login doc and these ones. We shouldn't skip steps like shown here: 
image

Please following the style from https://developer.okta.com/docs/api/resources/social_authentication.html#the-social-login-process by using 302 and GET properly. It would also be great if we standardized on the name for certain entities like your user agent, okta and your application across all docs that's going to use these new flow charts.

richardmateosian-okta:
@larsjohansen-okta Thanks for your feedback. I think it's important to emphasize the difference between the /token and /authorize endpoints, because that's important to understanding the differences between the various flows. I also think it's important to avoid loading the arrows with details. That information is in the text, and the diagrams are to provide an overview and framework for understanding the specifics in the text.

jakubvul-okta:
Minor, but according to the Platform style guide we aren't capitalizing "access token" or "ID token". @mystiberry-okta I think I remember you saying you don't think we should capitalize "authorization server" either, though I think we agreed that we should capitalize it when we are talking about the Okta Authorization Server entity (as opposed to the generic OAuth 2.0 term).

mystiberry-okta:
Jakub is correct--I just found a bunch of ID Token and Access Token that I fixed in other docs, likely where you picked this shortly-lived convention up. Okta Authorization Server, Custom Authorization Server, but authorization server.

richardmateosian-okta:
I will be so happy to stop capitalizing those terms.  :-)

_source/_standards/flows.md
```diff
@@ -0,0 +1,219 @@
...
+## Authorization Flows
+
+The preceding sequence of steps describes the Authorization Code flow, the most common way of obtaining tokens. The framework provides other possible flows. The choice of flow depends on your execution environment and security needs.
+
+The OAuth 2.0 flows are:
+
+  * Authorization Code -- your app receives an authorization code that it can exchange for an Access Token.
+  * Implicit -- the user authorizes the Authorization Server to send an Access Token directly to your app.
+  * Resource Owner Password -- Your app submits the user's credentials to the Authorization Server to obtain an Access Token.
+  * Client Credentials -- Your app submits its own credentials to the Authorization Server to obtain an Access Token.
```
cwu-okta: I am not clear what the client credntials flow does or why it is useful with this description. It sounds very similar to an Auth code flow. Elaborate on the differences.

_source/_standards/flows.md
```diff
@@ -0,0 +1,219 @@
...
+
+The OpenID Connect flows use the OAuth 2.0 flows to enable your app to receive ID Tokens as well as Access Tokens and Refresh Tokens. In addition, OpenID Connect specifies a Hybrid flow. It combines aspects of the Authorization Code and Implicit flows.
+
+The OpenID Connect flows are
```
cwu-okta: I'm not sure what value the OIDC flows section adds. It is a repeat of the OAuth flows except the mention of hybrid. Why not shorten this to just the description for hybrid and indicate it builds on top of all other OAuth flows.

_source/_standards/flows.md
```diff
@@ -0,0 +1,219 @@
...
+| `client_credentials` | Any or none | `client_id` and `client_secret` | Access | See note |
+
+> For the `refresh_token` grant type, scopes in the Access Token and the new Refresh Token are those in the supplied Refresh Token minus any not listed in the `scope` parameter in the request. For the `client_credentials` grant type, scopes in the Access Token are those in the request plus those granted by policy.
+
+
+### Authorization Code Flow
+
+{% img graphics/AuthCodeFlow.png alt:"Authorization Code Flow" %}
+
+<!-- Source for image generated using http://www.plantuml.com/plantuml/uml/
```
larsjohansen-okta: Are we generating the diagrams using a third party app? I wonder if legal would have issues with this; have we clarified with them about this?

_source/_standards/flows.md
```diff
@@ -0,0 +1,219 @@
...
+
+Your app redirects the user's browser to the Authorization Server's [`/authorize` endpoint](/docs/api/resources/oauth2.html#obtain-an-authorization-grant-from-a-user) with the `response_type` set to `token` or `token id_token` in the URL. It receives the requested combination of Access Token and ID Token in the response. The scopes in the returned tokens match those in the `scope` parameter of the request. In order for your app to receive an ID Token, `scope` must include `openid`.
+
+The Authorization Server authenticates the user and obtains the necessary permissions from the user but does not directly authenticate your app.
+
+The user and others with access to the machine your app runs on may be able to capture the returned tokens.
+
+Carefully consider the security implications of using this flow.
+
+### Resource Owner Password Flow
```
larsjohansen-okta: The right term is "Resource Owner Password Credentials" flow, I think we should change that everywhere where it is mentioned.

_source/_standards/flows.md
```diff
@@ -0,0 +1,219 @@
...
+The OpenID Connect flows are
+
+  * Authorization Code -- the OAuth 2.0 Authorization Code flow.
+  * Implicit -- the OAuth 2.0 Implicit flow.
+  * Hybrid -- the OAuth 2.0 Implicit flow, but your app also receives an authorization code.
+  * Password -- not part of the OpenID Connect specification, but you can obtain ID Tokens using the OAuth 2.0 Resource Owner Password flow.
+
+## How to Choose
+
+As a rule of thumb, use Authorization Code flow whenever possible. Use Implicit flow with single-page apps (SPAs) and other cases where your app consists entirely of software downloaded into a browser and does not have its own backend website. Use Client Credentials flow for communication between software packages within a single protected environment.
```
cwu-okta: If code flow is the recommended flow to use then more reason needs to be added here to let the dev know under what circumstances it makes sense to use this flow.

In general this doesn't do a good job helping the reader understand what flow he should use under what circumstance. For example, what is 'communication between software packages within a single protected environment'? I think this is too abstract and needs to be made more concrete.

_source/_standards/flows.md
cwu-okta: We should not use content generated by outside sources unless we own the license. You should contact legal team if it is imperative you use an outside source to generate images.

_source/_standards/flows.md
```diff
@@ -0,0 +1,219 @@
...
+@enduml
+-->
+
+The implicit flow is designed for situations in which your app is not on a separate website but instead is written in a language like JavaScript that runs within the user's browser. Examples include single-page apps (SPAs).
+
+This flow presents the following security problems:
+
+* You cannot exclude the client from communications between the user and the Authorization server, so your app might be able to steal the user's username and password.
+
+* You cannot exclude the browser from communication between your app and the Authorization Server, so any means of authenticating your app to the Authorization Server might be stolen by the user or anyone with access to the user's browser.
```
cwu-okta: This means that you cannot trust the client to store a secret. This is the reason why we do not allow refresh tokens for implicit flow.

_source/_standards/flows.md
```diff
@@ -0,0 +1,219 @@
...
+ot -> app: Requested tokens
+app -> pr
+@enduml
+-->
+
+Your app makes a request directly (with no prior request to the `/authorize` endpoint) to the Authorization Server's [`/token` endpoint](/docs/api/resources/oauth2.html#request-a-token) with `grant_type` set to `password`. The request includes the `username` and `password` parameters. By setting `scope` appropriately your app can receive an Access Token and any combination of ID Token and Refresh Token.
+
+This connection is directly between your app and the Okta Authorization Server, so it's not likely to be intercepted. However, it requires your app to know the user's username and password.
+
+Using this flow is not considered a good practice.
```
cwu-okta: Why is this not a good practice? Please provide more details if it isn't a good practice. Sometimes it might be necessary to support certain use cases.

_source/_standards/flows.md
```diff
@@ -0,0 +1,219 @@
...
+participant "Okta /token" as ot
+participant "Resource" as pr
+
+app -> ot: grant_type=client_credentials
+ot -> app: Access Token
+app -> pr
+@enduml
+-->
+
+If there is no end user in the picture, your app can use its own credentials (already registered with the Authorization Server). For example, if your app connects to an app that provides services to other apps, that app might accept tokens you obtain using the Client Credentials flow. This flow is typically for API-to-API communication between apps that are part of the same trusted environment. Using it may require setting up additional agreements outside the protocol.
```
cwu-okta: What does 'own credentials' mean? Don't all clients have userid/password combinations? In general this reads very abstractly and "additional agreements outside the protocol" is very hard to understand.

_source/_standards/flows.md
```diff
@@ -0,0 +1,219 @@
...
+
+Your app makes a request directly (with no prior request to the `/authorize` endpoint) to the Authorization Server's [`/token` endpoint](/docs/api/resources/oauth2.html#request-a-token) to obtain the desired tokens. The request has `grant_type` set to `client_credentials` and includes `client_id`, `client_secret`, and `scope`. This flow returns only an Access Token, which contains the scopes specified in the request.
+
+
+### Hybrid Flow
+
+This is the same as the [OAuth 2.0 Implicit flow](#implicit-flow), except that your app redirects the user's browser to the Authorization Server's [`/authorize` endpoint](/docs/api/resources/oauth2.html#obtain-an-authorization-grant-from-a-user) with the `response_type` set to `code id_token`, `code token` or `code id_token token` in the URL. It receives the requested combination of Access Token and ID Token in the response. It also receives an authorization code. The scopes in the authorization code and the returned tokens match those in the `scope` parameter of the request. In order for your app to receive an ID Token, `scope` must include `openid`.
```
cwu-okta: This is also a convenience flow (for efficiency) since you get two tokens with one request, eliminating the need to make an additional HTTP request to get both tokens.

_source/_standards/flows.md
```diff
@@ -0,0 +1,219 @@
...
+| --- | --- | --- | --- | --- |
+| `authorization_code` | Any or none + {`offline_scope` or `openid`} | `redirect_uri` | Access + {Refresh or ID} | From `/authorize` request |
+| `refresh_token` | Any + `offline_access` | `refresh_token` | Access + {Refresh} | See note |
+| `password` | Any or none + {`offline_scope` or `openid`} |   `username` and `password` | Access + {Refresh or ID} | From `/authorize` request |
+| `client_credentials` | Any or none | `client_id` and `client_secret` | Access | See note |
+
+> For the `refresh_token` grant type, scopes in the Access Token and the new Refresh Token are those in the supplied Refresh Token minus any not listed in the `scope` parameter in the request. For the `client_credentials` grant type, scopes in the Access Token are those in the request plus those granted by policy.
+
+
+### Authorization Code Flow
```
cwu-okta: After reading the various flows it still isn't super clear what the pros and cons are of each flow to help me decide (as a developer) which is the right one to use.

_source/_standards/flows.md
```diff
@@ -0,0 +1,219 @@
...
+
+
+### Hybrid Flow
+
+This is the same as the [OAuth 2.0 Implicit flow](#implicit-flow), except that your app redirects the user's browser to the Authorization Server's [`/authorize` endpoint](/docs/api/resources/oauth2.html#obtain-an-authorization-grant-from-a-user) with the `response_type` set to `code id_token`, `code token` or `code id_token token` in the URL. It receives the requested combination of Access Token and ID Token in the response. It also receives an authorization code. The scopes in the authorization code and the returned tokens match those in the `scope` parameter of the request. In order for your app to receive an ID Token, `scope` must include `openid`.
+
+
+### Refresh Tokens
+
+Refresh tokens eliminate the need for reauthorization. Access Tokens come with an expiration time and can be reused until they expire. When you acquire an Access Token, you can also acquire a Refresh Token, which has an expiration time and the same scopes as the Access Token. You can use the Refresh Token to obtain a fresh Access Token (with a later expiration time). You can also use a Refresh Token to obtain an Access Token with more limited scopes than the original Access Token.
```
cwu-okta: Both AT and ID tokens can be re-used until they expire. Calling out AT specifically can lead one to believe ID tokens have a different behavior.

```You can use the Refresh Token to obtain a fresh Access Token (with a later expiration time).``` This isn't super clear. It feels like all you need to say you can obtain another AT with a RT. Isn't it implied that if the AT is expired redeeming the RT for an AT will not contain the same expired date?

xrmxrm:
Thanks for all the detailed feedback. I think it will move the process forward.

mystiberry-okta:
Will restart in one week.

PR #902, opened 2017-05-26
Add Discourse comment support

mraible:
## Description:

Add comment support via Discourse using the jekyll-discourse-comments plugin.

https://github.com/oblakeerickson/jekyll_discourse_comments

To test, you will need to add the following entry to your `hosts` file:

```
127.0.0.1	developer.okta.com
```

Then navigate to <http://developer.okta.com:4000/blog> after running `npm start`.

This is what a post with no comments will look like:

blog-no-comments

After adding a comment:

blog-with-comments

When you click on "Start Discussion", you'll be taken to the topic on Discourse:

blog-comments-discourse

**NOTE:** This currently posts the initial blog post under my username. We could create a "devrel" account that posts the blogs as topics. I tried inviting devrel@okta.com, but I believe this is an internal-only alias.

To modify the settings for comments, login to Discourse as an admin and go to Admin > Customize > Embedding. Below is a screenshot of current settings.

discourse-embed-settings

## Type of Pull Request:

- [ ] Content (Documentation updates or typo-fixes)
- [ ] Blog Post
- [x] Functional (CSS changes, JS updates, or layout modifications)

## How Has This Been Tested?
- [x] I have built this locally and verified that it does not break existing functionality.
- [x] For functional changes, I have tested against Chrome, Firefox, and Safari, and mobile.
- [ ] For functional changes, I have added tests.

## Primary Reviewers:

DexEx Team

PR #1095, opened 2017-07-26
:seedling: Adds Discourse Comments

jmelberg-okta:
## Description:
Carried over from #902 -- @mraible:

Add comment support via Discourse using the jekyll-discourse-comments plugin.

https://github.com/oblakeerickson/jekyll_discourse_comments

To test, you will need to add the following entry to your `hosts` file:

```
127.0.0.1	developer.okta.com
```

Then navigate to <http://developer.okta.com:4000/blog> after running `npm start`.

This is what a post with no comments will look like:

blog-no-comments

After adding a comment:

blog-with-comments

When you click on "Start Discussion", you'll be taken to the topic on Discourse:

blog-comments-discourse

**NOTE:** This currently posts the initial blog post under my username. We could create a "devrel" account that posts the blogs as topics. I tried inviting devrel@okta.com, but I believe this is an internal-only alias.

To modify the settings for comments, login to Discourse as an admin and go to Admin > Customize > Embedding. Below is a screenshot of current settings.

discourse-embed-settings

## Type of Pull Request:

## How Has This Been Tested?
- [x] For functional changes, I have tested against Chrome, Firefox, and Safari, and mobile.


jmelberg-okta:
@mraible:
Overall, I think we are in good shape here.

Before we get this merged in, there are a couple of use cases I'd like us to test for.
  - Rendering a Post with `0` comments
  - Rendering a Post with `1` comment
  - Rendering a Post with `> 10` comments
    - Do we know if the comments are paged after a certain amount? Are they minified? I'd like us to have answers for these questions.
  - Does the comment counter work properly?

I think we can add to our selenium tests here: [blog-page-spec.js](https://github.com/okta/okta.github.io/blob/master/tests/selenium/spec/blog-page-spec.js).

mraible:
The only problem with writing integration tests for this is that these comments will show up on the production Discourse instance, under real blog entries, which means they'll show up on the site once this is live. We could script creating a blog entry, adding comments, then deleting everything, but that would require creating a PR and merging as part of our test process. Yuck.

PR #1098, opened 2017-07-27
no app-level MFA via api

mystiberry-okta:
## Description:
- Mention that app-level MFA via api is not supported.

### Resolves:

[OKTA-131972](https://oktainc.atlassian.net/browse/OKTA-131972)

### DIFF
--- a/_source/_docs/api/resources/authn.md
+++ b/_source/_docs/api/resources/authn.md
@@ -1403,9 +1403,9 @@ curl -v -X POST \
 
 ## Multifactor Authentication Operations
 
-You can enroll, activate, and verify factors using the `/api/v1/authn/factors` endpoint.
+You can enroll, activate, manage, and verify factors inside the authentication context with `/api/v1/authn/factors`.
 
-> You can also enroll, manage, and verify factors outside the authentication context with [`/api/v1/users/:uid/factors/`](factors.html#factor-verification-operations).
+> You can enroll, manage, and verify factors outside the authentication context with [`/api/v1/users/:uid/factors/`](factors.html#factor-verification-operations).
 
 ### Enroll Factor
 {:.api .api-operation}
--- a/_source/_docs/api/resources/policy.md
+++ b/_source/_docs/api/resources/policy.md
@@ -504,7 +504,7 @@ During policy evaluation each policy of the appropriate type is considered in tu
 Policies that have no rules are not considered during evaluation, and will never be applied.
 
 ### Policy Types
-Different policy types control settings for different operations.  All policy types share a common framework, message structure and API, but have different policy settings and rule data.  The data structures specific to each policy type are discussed in the various sections below.
+Different policy types control settings for different operations.  All policy types share a common framework, message structure and API, but have different policy settings and rule data. The data structures specific to each policy type are discussed in the various sections below.
 
 <a href="#OktaSignOnPolicy">Okta Sign On Policy</a>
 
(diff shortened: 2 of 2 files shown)


mystiberry-okta:
I need an MFA expert to verify this. @lilyhsiao-okta Whom should I ask? Also asking in hipchat.

_source/_docs/api/resources/authn.md
```diff
@@ -1403,9 +1403,12 @@ curl -v -X POST \
...
 ## Multifactor Authentication Operations
 
-You can enroll, activate, and verify factors using the `/api/v1/authn/factors` endpoint.
+You can enroll, activate, and verify factors in two ways:
 
-> You can also enroll, manage, and verify factors outside the authentication context with [`/api/v1/users/:uid/factors/`](factors.html#factor-verification-operations).
+* Use the `/api/v1/authn/factors` endpoint.
+* Outside the authentication context, use [`/api/v1/users/:uid/factors/`](factors.html#factor-verification-operations).
+
+>Note: Configuring application-level MFA via the API is not supported.
```
tthompson-okta: Hmm, this seems like the wrong place to be talking about configuring App Sign-on Policy. The reason it's confusing is that this API *does* handle MFA whether its triggered via *Okta* sign-on OR *App* sign-on policy.

I would put this bit of text in the Policy API doc.

mystiberry-okta: Wait for email chain to resolve.

mystiberry-okta:
@tthompson-okta I think this fixes it.

mystiberry-okta:
@tthompson-okta Is this shippable? Thanks!

_source/_docs/api/resources/policy.md
```diff
@@ -892,9 +892,10 @@ The following conditions may be applied to the rules associated with Okta Sign O
 ## Multifactor (MFA) Policy
 {: #OktaMFAPolicy }
 
-> The MFA Policy API is a {% api_lifecycle beta %} [release](/docs/api/getting_started/releases-at-okta.html).
+> The MFA Policy API is a {% api_lifecycle beta %} release.
 
 Multifactor (MFA) Policy controls which MFA methods are available for a user, as well as when a user may enroll in a particular factor.
+To configure app-level MFA settings, use the [Factors API](/docs/api/resources/factors.html).
```
tthompson-okta: Actually, we don't support configuring app-level MFA settings (i.e. App Sign-On Policy) at all; once we move App Sign-On Policy over to the standard Policy Framework, then it will be configurable via this API (the Policy API).

The Factors API doesn't manage policy, it just manages individual factor enrollments for users.

So the thing I think we should point out here in this doc is that while you can configure Okta SignOn Policy and MFA Enrollment Policy (by the way that's what this section is about -- we should add the `Enrollment` part to make this clear), you cannnot currently configure App Sign-On Policy via this API. I think this should go somewhere around the Okta Sign-On Policy content, rather than here with the MFA Enrollment Policy.

_source/_docs/api/resources/authn.md
```diff
@@ -1403,9 +1403,10 @@ curl -v -X POST \
 
 ## Multifactor Authentication Operations
 
-You can enroll, activate, and verify factors using the `/api/v1/authn/factors` endpoint.
+You can enroll, manage, and verify factors outside the authentication context with [`/api/v1/users/:uid/factors/`](factors.html#factor-verification-operations).
 
-> You can also enroll, manage, and verify factors outside the authentication context with [`/api/v1/users/:uid/factors/`](factors.html#factor-verification-operations).
```
tthompson-okta: I think this part actually reads better as it was before. I'd leave this section as-is and just add the notes to Policy API doc (per the other comments).

_source/_docs/api/resources/policy.md
mystiberry-okta: Ack on suggestions. Just want to confirm that you want the title "Multifactor (MFA) Policy" changed to "Multifactor (MFA) Enrollment Policy" I'm guessing they left that out since consent may be added. We can add "and Consent" to the title when that happens.

_source/_docs/api/resources/authn.md
mystiberry-okta: I would like to leave the addition of "outside the authentication context" because it's a pair with the following sentence.  Also, we aren't solving the original problem, which was the customer was here when they thought app-level MFA should be available. They missed the note that's already in the Okta Sign-on Policy section (though that note could be clearer)... Should I add the note requested in the bug here? "Note: Configuring app-level MFA via the API is not supported"?

PR #1100, opened 2017-07-27
Ran npm run build-prod to bring master current

dogeared:
## Description:
- Ran npm run build-prod to bring master current


dogeared:
not needed

Issue #1110, opened 2017-07-29
Fix feed.xml


dogeared:
Currently, the `feed.xml` template embeds the entire content of a post in the `<description>` section. This causes the resultant `feed.xml` to be very large.

Instead, it should reference the post's description from the front matter.

jmelberg-okta:
Fixed with #1221

PR #1111, opened 2017-07-29
Updated feed to include description, instead of the entire post.

dogeared:
## Description:
- fixes feed.xml to be of a normal size so that it can be used on feedburner and properly represents a post's description.

### Resolves:
* [1110](https://github.com/okta/okta.github.io/issues/1110)

### DIFF
--- a/_source/feed.xml
+++ b/_source/feed.xml
@@ -11,7 +11,7 @@ layout: null
     {% for post in site.posts %}
       <item>
         <title>{{ post.title | xml_escape }}</title>
-        <description>{{ post.content | xml_escape }}</description>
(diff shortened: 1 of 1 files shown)


dogeared:
yes, we haven't traditionally included a description in the front matter for our blog posts. But, we could start doing so. And, it wouldn't break feed burner. Right now, having the entire post in the description field makes the resultant feed.xml too large for feedburner.

jmelberg-okta:
@dogeared: Shouldn't we just update all the description fields? Seems like this is something we would want to have before we ship.

jmelberg-okta:
Closing due to: #1221

PR #1113, opened 2017-07-31
:fire: Updates URLs

jmelberg-okta:
## Description:
- Updates the URLs used in sample code and callouts to be consistent with our [Design Principles](https://developer.okta.com/docs/api/getting_started/design_principles.html).
- Starts `consistency-check` script to find all occurrences in `*.md` files.

#### NEW Example Domain:
`https://{yourOktaDomain}.com`

### Resolves:

* [OKTA-135771](https://oktainc.atlassian.net/browse/OKTA-135771)


jmelberg-okta:
@jakubvul @mystiberry-okta:

As I was parsing through D.O.C, I realized the domain we use for sample HTTP calls, API callouts, etc is different across pages. I'd like to include a `lint` check to ensure we all use the same example before changes are published.

As a side note our [Design Principles](https://developer.okta.com/docs/api/getting_started/design_principles.html) state we will use `https://your-domain.okta.com` in our docs. This is also consistent in our release notes.

I'm fond of: `https://{yourOrg}.oktapreview.com` or `https://example.oktapreview.com`, but know this is a much larger discussion.

jmelberg-okta:
@rchild-okta, @omgitstom, @johngronberg-okta: FYI ^

PR #1116, opened 2017-07-31
Update policy api doc with new features

mpeng-okta:
## Description:
- Update Policy API with weak password dictionary and optional recovery question changes

### Resolves:

* [OKTA-135946](https://oktainc.atlassian.net/browse/OKTA-135946)

### Password Complexity (Weak Password Dictionary added)
image

image

### Question Recovery Factor (Optional Status)
 image

I ran out of possible subhheadings, so there's insufficient nesting available. (maximum 6 hashes) For that reason, Weak Password Dictionary Object is in the document structure, but Common Password Lookup Object isn't, like so:
image

Let me know if you have any suggestions for the naming/phrasing/structure of this doc.

### DIFF
--- a/_source/_docs/api/resources/policy.md
+++ b/_source/_docs/api/resources/policy.md
@@ -1098,7 +1098,27 @@ minLowerCase | Indicates if a password must contain at least one lower case lett
 minUpperCase | Indicates if a password must contain at least one upper case letter: 0 indicates no, 1 indicates yes | integer | No | 1
 minNumber | Indicates if a password must contain at least one number: 0 indicates no, 1 indicates yes | integer | No | 1
 minSymbol | Indicates if a password must contain at least one symbol (e.g., !@#$%^&*): 0 indicates no, 1 indicates yes | integer | No | 1
-excludeUserName | Indicates if the user name must be excluded from the password | boolean | no | true
+excludeUserName | Indicates if the user name must be excluded from the password | boolean | No | true
+dictionary {%api_lifecycle beta %} | Weak password dictionary lookup settings | <a href="#WeakPasswordDictionaryObject">Weak Password Dictionary Object</a> | No | N/A
+
+###### Weak Password Dictionary Object
+{: #WeakPasswordDictionaryObject }
+
+> Weak password dictionary lookup is an {% api_lifecycle beta %} feature. Contact Support to enable it.
+
+Specifies settings for dictionaries of weak passwords against which lookups of users' passwords may be done, in order to help verify password strength. Designed to be extensible with multiple possible dictionary types against which to do lookups.
+
+Property | Description | Data Type | Required
+| --- | --- | --- | ---
+common | Lookup settings for commonly used passwords  | <a href="#CommonPasswordLookupObject">Common Password Lookup Object</a> | No
+
+Common Password Lookup Object
+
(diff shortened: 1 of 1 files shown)


_source/_docs/api/resources/policy.md
```diff
@@ -1098,7 +1098,27 @@ minLowerCase | Indicates if a password must contain at least one lower case lett
...
-excludeUserName | Indicates if the user name must be excluded from the password | boolean | no | true
+excludeUserName | Indicates if the user name must be excluded from the password | boolean | No | true
+dictionary {%api_lifecycle beta %} | Weak password dictionary lookup settings | <a href="#WeakPasswordDictionaryObject">Weak Password Dictionary Object</a> | No | N/A
+
+###### Weak Password Dictionary Object
+{: #WeakPasswordDictionaryObject }
+
+> Weak password dictionary lookup is an {% api_lifecycle beta %} feature. Contact Support to enable it.
+
+Specifies settings for dictionaries of weak passwords against which lookups of users' passwords may be done, in order to help verify password strength. Designed to be extensible with multiple possible dictionary types against which to do lookups.
```
gabewerman-okta: @mystiberry-okta how do you feel about this first sentence?

jakubvul-okta: It really should be `username`. We don't refer to it as a "user name" anywhere, which also means that the property should be `excludeUsername`, but I don't think that could be changed at this point.

jakubvul-okta: Also, @mystiberry-okta you can correct me here, but I don't think we put Beta tags inside property tables.

mystiberry-okta: It used to break the table but it doesn't now (?). I'm happier with the tag.

mystiberry-okta: @gabewerman-okta:
1. don't capitalize Lookup. 
2. You don't need the Contact Support sentence, the label links to the doc that explains all that. 
3. Is there any other kind of lookup aside from dictionary lookups? If not, I'd rewrite like this:

Weak password lookup is an.....

Specifies how lookups for weak passwords are done. Designed...

mystiberry-okta: Also, Jakub is correct, we call it username everywhere.

mpeng-okta: @mystiberry-okta  Thanks! Yeah, it was `excludeUserName` before my change, and I didn't catch it on my pass through.

1. Which "lookup" would you like me not to capitalize? I have it capitalized in the description for the `common` property and in the name of the Common Password Lookup Object.
2. Ack, removed!
3. There isn't yeah, it's just a lookup in different dictionaries (and there will be multiple * Password Lookup Objects that let us specify through the API which dictionaries to use). I've change it accordingly. Since I've removed the Contact Support sentence, the "Weak password lookup..." sentence is gone, but I've changed the other sentence.

mystiberry-okta: @mpeng-okta oops, I just meant don't capitalize it in the two sentences above. If it's the first word in a sentence or the first word in a column like Description, caps are fine.

mpeng-okta: Gotcha, thanks!

PR #1132, opened 2017-08-04
:fire: Remove dead pages

jmelberg-okta:
## Description:
- Removes outdated paths and unused files.
- `_site` contained duplicated `.html` pages that haven't been updated in ~ 1 year

### Resolves:
* [OKTA-135960](https://oktainc.atlassian.net/browse/OKTA-135960)


docs/api/resources/zones.md
```diff
@@ -1,643 +0,0 @@
-# Zones API
```
mystiberry-okta: I know this looks bad, but it's an actual doc. We shouldn't delete it or it's HTML companion. It's going to get improved next week.

docs/api/resources/index
```diff
@@ -1,8 +0,0 @@
-<!DOCTYPE html>
```
mystiberry-okta: I'm concerned about removing these index files. I think they exist so that if someone types developer.okta.com/docs/api they'll get a good result. ??

_source/_docs/api/resources/resource-server-beta/required-config-changes-for-auth-server.md
```diff
@@ -1,332 +1,4 @@
 ---
 layout: docs_page
```
mystiberry-okta: I'll ask Keith if we can delete this file. It was a "patch" for about 10 customers.

docs/api/resources/zones.md
jmelberg-okta: This file is duplicated. See the updated one here: [_docs/api/resources/zones.md](https://github.com/okta/okta.github.io/blob/master/_source/_docs/api/resources/zones.md)

docs/api/resources/index
jmelberg-okta: That is handled by `index.html`. These files are not read by the browser.

mystiberry-okta: Ah, I didn't know that. Thank you! So just the zones file needs to stay.

docs/api/resources/zones.md
mystiberry-okta: Got it, thanks.

PR #1134, opened 2017-08-04
making EA tags consistent

mystiberry-okta:
## Description:
- A few files were using old style text instead of EA tags.

### Resolves:
OKTA-126953


mystiberry-okta:
@jmelberg-okta At your convenience, could you review these minor changes?

PR #1154, opened 2017-08-11
Check in gem dependencies

jmelberg-okta:
## Description:
- Fix typos throughout site
- Check in `Gemfile.lock` to keep our dependencies constant.


_source/_docs/api/resources/oauth2.md
```diff
@@ -885,8 +885,8 @@ Okta provides a pre-configured Custom Authorization Server with the name `defaul
 This default authorization server includes a basic access policy and rule, which you can edit to control access. 
 It allows you to specify `default` instead of the `authorizationServerId` in requests to it:
 
-* `https://{YourOktaOrg}}/api/v1/authorizationServers/default`  vs 
-* `https://{YourOktaOrg}}/api/v1/authorizationServers/:authorizationServerId` for other Customer Authorization Servers
+* `https://{yourOktaDomain}.com/api/v1/authorizationServers/default`  vs
```
bdemers: This gets a little tricky with custom domains as it could be `my-okta.example.org` but as long as we keep it consistent from what we have elsewhere that works for now

jmelberg-okta: This really only gets confusing when you can have a different top-level domain, but this should be clear enough to the reader that they would need to replace the entire domain.

FYI: Custom URLs are still a roadmap item, so throughout the rest of the site we will currently use the convention: `https://{yourOktaDomain}.com`. Not sure if this is correct, but at least it is consistent.

bdemers: +1 for consistently

PR #1158, opened 2017-08-14
Add java jwt validation page

bdemers:
Also added a new variable section for library versions in _config.yml


_source/_code/java/jwt-validation.md
```diff
@@ -0,0 +1,90 @@
...
+  </dependency>
+```
+
+# Setting up the Library
+
+The Okta JWT Verifier can created via a fluent `JwtHelper` class:
+
+```java
+JwtVerifier jwtVerifier = new JwtHelper()
+    .setIssuerUrl("https://{oktaOrgDomain}.com.oktapreview.com/oauth2/default")
```
jmelberg-okta: That is a fun URL!
-> `https://{yourOktaDomain}.com/oauth2/defalut`

_source/_code/java/jwt-validation.md
```diff
@@ -0,0 +1,90 @@
...
+layout: docs_page
+title: JWT Validation Guide
+excerpt: How to validate Okta JWTs with Java.
+support_email: developers@okta.com
+---
+
+# Overview
+
+As a result of a successful authentication by [obtaining an authorization grant from a user](https://developer.okta.com/docs/api/resources/oauth2.html#obtain-an-authorization-grant-from-a-user) using the Okta API, you 
+using the Okta API, you will be provided with a signed JWT (`id_token` and/or `access_token`). A common use case for
```
jmelberg-okta: I'm curious why this paragraph is broken up into different lines. Can't this just be one?

_source/_code/java/jwt-validation.md
```diff
@@ -0,0 +1,90 @@
...
+The Okta JWT Verifier can created via a fluent `JwtHelper` class:
+
+```java
+JwtVerifier jwtVerifier = new JwtHelper()
+    .setIssuerUrl("https://{oktaOrgDomain}.com.oktapreview.com/oauth2/default")
+    .setClientOrAudience("api://default")
+    .build();
+```
+
+This helper class configures a JWT parser with the details found via the [OAuth discovery endpoint](https://openid.net/specs/openid-connect-discovery-1_0.html).  The public keys used to validate the JWTs will also be retrieved
```
jmelberg-okta: OAuth discovery -> OpenID Connect discovery

_source/_code/java/jwt-validation.md
```diff
@@ -0,0 +1,90 @@
...
+`decode` method (where `jwtString` is your access token in string format).
+
+```java
+Jwt jwt = jwtVerifier.decodeAccessToken(jwtString);
+```
+
+This will validate your JWT for the following:
+
+- Token expiration date
+- Valid token Not Before date
```
jmelberg-okta: Not Before -> not before

_source/_code/java/jwt-validation.md
```diff
@@ -0,0 +1,90 @@
...
+```
+
+# Setting up the Library
+
+The Okta JWT Verifier can created via a fluent `JwtHelper` class:
+
+```java
+JwtVerifier jwtVerifier = new JwtHelper()
+    .setIssuerUrl("https://{oktaOrgDomain}.com.oktapreview.com/oauth2/default")
+    .setClientOrAudience("api://default")
```
jmelberg-okta: This is a bit confusing. An `idToken` and `accessToken` have these elements. I'm assuming this is your way of indicating the difference between an `accessToken` and `idToken`?

_config.yml
```diff
@@ -92,6 +92,10 @@ defaults:
     values:
       layout: documentation
 
+# Okta Library versions
+versions:
```
jmelberg-okta: Looks like this breaks some of our tests.

_source/_code/java/jwt-validation.md
bdemers: ha, I meant to go back and look up the correct one, but event then this one looks pretty gnarly

_source/_code/java/jwt-validation.md
bdemers: For one type of token we use audience (`aud`) for the other we use clientId (`cid`).  You would call `decodeIdToken` or `decodeAccessToken` and the correct claim will be used (but the value is the same on either one).  If you need the values to be you could create two instances of a `JwtVerifier`.

_source/_code/java/jwt-validation.md
bdemers: Yes, but it is just manual word wrapped to fit on 120 char lines. This makes no difference on on the rendering of this file.  It does point out the line `using the Okta API, you 
+using the Okta API`
@bretterer i'll push a fix for the PHP page as well

_source/_code/java/jwt-validation.md
bdemers: Did the OAuth version of this never leave draft ?  I wanted to _try_ not confuse the two, but I this was the only _non_ draft spec I found.

_config.yml
bdemers: No, looks like the `/code/java` page test is just a little on the week side. I'll fix it

bdemers:
Lint fix on it's way. This was caused by the referencing of gh-pages sub projects

_source/_code/java/jwt-validation.md
jmelberg-okta: I see the logic here. Can we add a comment around this to clear it up?

tests/selenium/spec/code-layout-spec.js
```diff
@@ -1,6 +1,6 @@
 const CodePage = require('../framework/page-objects/CodePage');
 
-describe('code page spec', () => {
+describe('code page spec (java)', () => {
```
jmelberg-okta: Good find!

bdemers:
@jmelberg-okta RE: `setClientOrAudience()` if this is not clear by the method name alone, maybe we should change the method name. thoughts?

jmelberg-okta:
@bdemers: I think that is the right solution.

The purpose of the `aud` claim is to [identify who the JWT is intended for](https://tools.ietf.org/html/rfc7519#section-4.1.3). For an `idToken`, this can be single or multiple `clientId`s. For an `accessToken`, it is a list of all possible resource servers who can accept this token.

Could this be simplified to:

```java
```java
JwtVerifier jwtVerifier = new JwtHelper()
    .setIssuerUrl("https://{oktaOrgDomain}.com.oktapreview.com/oauth2/default")
   .for("api://default")
// .for("AbC1234")
// .for({"api://default", "myapi://default", "etc://default" })

```

bdemers:
I like the `.for()`!, I still think it would be singular though.  The token may have multiple clientId's but the application would validate against a single one (i.e. make sure my `id` is in the list).
I have a slight reservation though, if we start adding more helper setters it may be less obvious.  For example, the php equivalent has a `setDiscovery()` function which lets you switch between the OAuth and ODIC discovery endpoints (or a custom solution)

@bretterer thoughts

jmelberg-okta:
@bdemers I'm cool with it being singular for now.

So is this the intended use?
```java
// Validate idToken
JwtVerifier jwtVerifier = new JwtHelper()
    .setIssuerUrl("https://{oktaOrgDomain}.com.oktapreview.com/oauth2/default")
    .for("AbC1234")

Jwt jwt = jwtVerifier.decodeIdToken(jwtString);
// Validates the iss == issuer
// Validates aud == clientId
// Etc..

// Validate accessToken
JwtVerifier jwtVerifier = new JwtHelper()
    .setIssuerUrl("https://{oktaOrgDomain}.com.oktapreview.com/oauth2/default")
    .for("api://default")

Jwt jwt = jwtVerifier.decodeIdToken(jwtString);
// Validates the iss == issuer
// Validates cid == clientId
// Validates aud == my server
// Etc..
```

bdemers:
@jmelberg-okta @bretterer related:
https://github.com/okta/okta-jwt-verifier-java/pull/5/files#diff-04c6e90faac2675aa89e2176d2eec7d8

bdemers:
@jmelberg-okta can you take another look at this one ?

PR #1161, opened 2017-08-15
Fully document Zones API

mpeng-okta:
## Description:
- Fully documents the Zones API
### Resolves:
* [OKTA-137181](https://oktainc.atlassian.net/browse/OKTA-137181)

## Screenshots:
image


mystiberry-okta:
Hi @mpeng-okta , if you click on Details next to either Travis CI build failed message, scroll to the bottom, you'll see the broken link you need to fix.

mpeng-okta:
Yeah, sorry about that! I've noticed a bunch of issues so I've been fixing them locally. Thanks!

mystiberry-okta:
@mpeng-okta This looks great! I'm just waiting for your technical reviewer, then we can merge this, yes?

mpeng-okta:
Yup! :)

_source/_docs/api/resources/zones.md
```diff
@@ -7,18 +7,26 @@ title: Zones
 
 {% api_lifecycle beta %} This API is [a Beta release](/docs/api/getting_started/releases-at-okta.html).
```
tthompson-okta: We should remove the beta tags right?

_source/_docs/api/resources/zones.md
```diff
@@ -641,17 +898,4 @@ curl "https://yourorg.okta.com/api/v1/org/zones/?limit=-1&q=First"
     }
   }
 }
-~~~
-
```
nandagopalseshagiri-okta: Why is this removed?

Gemfile.lock
```diff
@@ -1,236 +0,0 @@
...
-  ruby
-
-DEPENDENCIES
-  github-pages (= 148)
-  html-proofer
-  jekyll-assets
-  tzinfo-data
-
-BUNDLED WITH
-   1.15.1
```
nandagopalseshagiri-okta: Why is this file being removed?

_source/_docs/api/resources/zones.md
```diff
@@ -7,18 +7,26 @@ title: Zones
 
 {% api_lifecycle beta %} This API is [a Beta release](/docs/api/getting_started/releases-at-okta.html).
 
-Zones are used to group IP Address ranges so that policy decisions can be made based on the client's IP location.
+The Okta Zones API provides operations to manage zones in your organization. Zones may be used to guide policy decisions.
```
nandagopalseshagiri-okta: A reason to remove the sentence the explains what zones are?

_source/_docs/api/resources/zones.md
```diff
@@ -104,17 +112,18 @@ curl -X POST
...
+curl -X POST
 -H "Accept: application/json"\
 -H "Content-Type: application/json" \
+-H "Authorization: SSWS ${api_token}" \
 -d '
 {
   "type": "IP",
   "id": null,
-  "name": "kgysngljmtiaeibcmpavhytgfevfbwjdpnmqnujmhdlgiiergvepikjmacdxmfntpfnarrlnisfpfyfbfehdvibuwgwgwhwhcxknenlrbdsebsgaernokhvaepdbdffdzbtn",
+  "name": "111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111",
```
tthompson-okta: Hmm, this is a really weird example to publish for an invalid request. I think we'd want to highlight something more specific to the zone data contract, like invalid CIDR or invalid IP range.

_source/_docs/api/resources/zones.md
mpeng-okta: I believe @mystiberry-okta will do this as part of the RN work.

_source/_docs/api/resources/zones.md
mpeng-okta: I changed the line to be more general, and moved the specific line regarding "policy decisions can be made based on the client's IP location" below when describing IP Zones specifically, in anticipation of potentially adding more types of zones.

_source/_docs/api/resources/zones.md
mpeng-okta: 👍

_source/_docs/api/resources/zones.md
mpeng-okta: This was moved up to the IP Zones Properties section and modified slightly

Gemfile.lock
mpeng-okta: This was necessary to pass travis--there was a breaking change introduced (that's only on the master branch, not on the weekly branch, against which doc updates must be based) where the repo no longer takes a Gemfile.lock file.

_source/_docs/api/resources/zones.md
```diff
@@ -104,287 +112,297 @@ curl -X POST
...
+
+##### Request Example
 {:.api .api-request .api-request-example}
 ~~~sh
-curl "https://yourorg.okta.com/api/v1/org/zones/"
+curl -X GET \
+-H "Accept: application/json" \
+-H "Content-Type: application/json" \
+-H "Authorization: SSWS ${api_token}" \
+"https://yourorg.okta.com/api/v1/zones"
```
tthompson-okta: Per Mysti's email from a few weeks ago, the convention across all examples is now `https://{yourOktaDomain}.com`. This prepares us for vanity URLs and other changes to domains that may occur, and gets rid of all the accidental "rain" examples.

For example: https://developer.okta.com/docs/api/resources/apps.html#request-example

mpeng-okta:
I believe she'Il add this to the ToC, yeah. Let me check in with her to make sure.

_source/_docs/api/resources/zones.md
mpeng-okta: 👍

mpeng-okta:
@tthompson-okta Mysti has been ill, but we had a chance to communicate, and she confirmed that she'll be flipping the lifecycle tag and adding this reference doc to the ToC Monday morning.

_source/_docs/api/resources/zones.md
```diff
@@ -434,18 +452,49 @@ curl "https://yourorg.okta.com/api/v1/org/zones/?limit=100&filter=%28id+eq+%22nz
...
+
+#### List Zones with a Filter
+{:.api .api-operation}
+
+Lists all zones that match the filter criteria
+
+This operation:
+
+ * Filters against the most up-to-date data. For example, if you create a zone or change an attribute and then issue a filter request, the changes are reflected in your results.
+ * Requires [URL encoding](http://en.wikipedia.org/wiki/Percent-encoding). For example, `filter=type eq "IP"` is encoded as `filter=type%20eq%20%22IP%22`.
```
tthompson-okta: Hmm, shouldn't we just link to https://developer.okta.com/docs/api/getting_started/design_principles.html#filtering ?

mpeng-okta: I got this from the Users API https://developer.okta.com/docs/api/resources/users.html#list-users-with-a-filter, I think it's specifically to discuss the URL percent encoding. I can go ahead and change the link if you think it would be more appropriate

_source/_docs/api/resources/zones.md
```diff
@@ -434,18 +452,49 @@ curl "https://yourorg.okta.com/api/v1/org/zones/?limit=100&filter=%28id+eq+%22nz
...
+
+> Hint: If filtering by `name`, it may be easier to use `q` instead of `filter`.
```
tthompson-okta: Ah, I didn't realize we implemented `q` on this API. We should not document this and ultimately remove it from code. We have standardized on SCIM filter syntax across all APIs; `q` remains in place for some legacy use cases like User API (where we search across multiple fields), but for Zones the only searchable field should be the name.

Here's some additional data for reference: https://oktawiki.atlassian.net/wiki/spaces/eng/pages/143425844/API+Search+Patterns+and+Scalability

_source/_docs/api/resources/zones.md
```diff
@@ -434,18 +452,49 @@ curl "https://yourorg.okta.com/api/v1/org/zones/?limit=100&filter=%28id+eq+%22nz
...
+~~~
+
+#### List Zones with a Filter
+{:.api .api-operation}
+
+Lists all zones that match the filter criteria
+
+This operation:
+
+ * Filters against the most up-to-date data. For example, if you create a zone or change an attribute and then issue a filter request, the changes are reflected in your results.
```
tthompson-okta: I don't think this explanation is necessary, as this is the SLA (and expectation) across all of our APIs .. what you PUT is what you GET, with immediate consistency.

_source/_docs/api/resources/zones.md
mpeng-okta: Ah, got it. I'll remove references to `q` in this doc then.

_source/_docs/api/resources/zones.md
```diff
@@ -434,18 +452,49 @@ curl "https://yourorg.okta.com/api/v1/org/zones/?limit=100&filter=%28id+eq+%22nz
...
+ * Requires [URL encoding](http://en.wikipedia.org/wiki/Percent-encoding). For example, `filter=type eq "IP"` is encoded as `filter=type%20eq%20%22IP%22`.
+
+> Hint: If filtering by `name`, it may be easier to use `q` instead of `filter`.
+
+See [Filtering](/docs/api/getting_started/design_principles.html#filtering) for more information about the expressions used in filtering.
```
tthompson-okta: This is good, but we need a table explaining what attributes can be filtered on (i'm assuming `name`) and what specific filter operators are supported (i.e. `sw`, `co`, `eq`, etc)

_source/_docs/api/resources/zones.md
mpeng-okta: Makes sense, I just got it from the Users API.

_source/_docs/api/resources/zones.md
mpeng-okta: That makes sense, i.e. you have something like https://developer.okta.com/docs/api/resources/users.html#list-users-with-a-filter in mind, right?

Would you happen to know if the filterable attributes are listed anywhere, by any chance?

_source/_docs/api/resources/zones.md
tthompson-okta: Ahh, this makes more sense now. The reason we explicitly call out the "filters against most up-to-date" part in the User API is because we do have a special search operations for Users only which goes against the eventual-consistent datastore (elastic search). I think in general this quoting guidance is just part of the SCIM syntax and applies everywhere; so it seems out of place to call it out here.

_source/_docs/api/resources/zones.md
mpeng-okta: Spoke offline, will add in future.

mpeng-okta:
Created this epic to track Post GA work, with 2 corresponding JIRAs to 1) remove `q` support from the backend (currently not documented and therefore not supposed to be supported), 2) add table of supported filter operations and filterable attributes to the doc

mystiberry-okta:
Confirming I'll change the tag & add to ToC in separate pr.

mpeng-okta:
Awesome, thanks Mysti!

PR #1174, opened 2017-08-17
Production build

jmelberg-okta:
## Description:
- Production build on PR #1173


Gemfile.lock
```diff
@@ -0,0 +1,249 @@
+GEM
```
rchild-okta: Does this need to be removed because it is not accompanied by the travis.yml change?

jmelberg-okta: Yep. Added that by accident. Good catch.

We are still rebuilding all of our assets, causing the 200+ file changes. But we can investigate that later.

PR #1180, opened 2017-08-17
Developer Redesign Post-launch Updates/Fixes

benbridge:
## Description:
This addresses items that were identified after launching the developer redesign on 8-16.

- fixes footer links (the link to the Integrate page was mentioned in https://github.com/okta/okta.github.io/pull/1172, which should not be needed any more when this gets merged)
- removes a target attribute
- fixes the mobile nav open/close icons and styles
- adjusts code sample font sizes in blog posts
- fixes overflow caused by header on small viewports
- ~~adds a query string to the shared master.css file being linked to from okta.com~~
- Changes inline code block bg color

I cherry picked `70fdd8fca2d34d1ffd3edb750aba81c3e96bcf26` from the [`blog_test_fix`](https://github.com/okta/okta.github.io/tree/blog_test_fix) branch to allow blog tests to pass, hopefully that doesn't introduce any problems.

Thanks.


rchild-okta:
@benbridge it looks like you removed the cachebusting param from the css link - does that mean that you're thinking of doing another approach to deal with the caching issues?

Also, as a process note for the future - I think it will be easier if we save the build-prod for after the review is approved - that way it's easier to look at the commits, and when you get the ship you can push one more commit after merging in latest master and running the prod build.

benbridge:
@rchild-okta The cache busting we added is not ideal, since it would happen when the Jekyll site is built, not when the theme in Drupal is updated and it would not invalidate all of the caches anyway. That said, after changes are made on the Drupal side, multiple caches need to be cleared. Right now, this is a manual process, but we're hoping to automate it at some point (after we have the access we need).

Apologies for adding the prod build, I'll wait for approval next time.

Would you recommend removing the prod build commit from the feature branch and pushing back up here? I'd be happy to rebase and clean up the branch.

Thanks for the feedback.

rchild-okta:
@benbridge yes please, rebase & push w/out the prod branch. 

Am still wondering about the caching - so there's the open question of invalidating the cloudfront cache, but I am also concerned about the users' browser caches - would still need some cachebusting param. I don't think we need to do it for this PR since there are some other changes here that we need to get in, but the strategy we use in the okta.github.io repo is to cachebust based on the file sha (although pulling that in here in the separate repo is an issue with that approach).

benbridge:
@rchild-okta The link check test fails with the query string and we can't use the asset pipeline to hash the filename since it's a remote file. It looks like updating the link checker script `scripts/find-missing-slashes.js` to allow the query string on the external css link and adding the query string back would be the path of least resistance. 

Do you think that would work?

rchild-okta:
@benbridge yep that would work

benbridge:
@rchild-okta Sounds good, that will come in separate pr. Appreciate the feedback.

PR #1184, opened 2017-08-18
:seedling: Adds Angular Implicit quickstart

jmelberg-okta:
## Description:
- Adds angular implicit quickstart

### Resolves:
* [OKTA-137727](https://oktainc.atlassian.net/browse/OKTA-137727)


quickstarts/widget/default-example.html
```diff
@@ -1,10 +0,0 @@
-<!DOCTYPE html>
```
lboyette-okta: so, we're keeping the quickstart directory, without the `s`?

jmelberg-okta: Yep. By accident we created quickstart(s). I'm forwarding all the traffic to the correct path `quickstart`.

See: https://developer.okta.com/quickstart/#/widget/nodejs/generic

PR #1189, opened 2017-08-18
Redirect javadoc files in this project to new location

bdemers:


jmelberg-okta:
We should probably just set a redirect on just the root page: `_docs/sdk/core/java_api_sdk/`. I'm not sure if we link directly to any of the classes externally.

bdemers:
I doubt we link out, but there are likely existing links else where, so we should keep them (it is easier to switch them all now, and then prune them out later if needed) Ideally with a 301 redirect rule instead of html files (but we can do that that once we deploy to an s3 bucket)

bdemers:
From what I can tell, because the meta redirect sets the canonical link, we should be OK for SEO. Though meta refreshes are NOT idea (i'm guessing this is better then the the multiple duplicates we have now)

```html
<!DOCTYPE html>
<html lang="en-US">
  <meta charset="utf-8">
  <title>Redirecting…</title>
  <link rel="canonical" href="https://developer.okta.com/okta-sdk-java/0.0.4/apidocs/index.html">
  <meta http-equiv="refresh" content="0; url=https://developer.okta.com/okta-sdk-java/0.0.4/apidocs/index.html">
  <h1>Redirecting…</h1>
  <a href="https://developer.okta.com/okta-sdk-java/0.0.4/apidocs/index.html">Click here if you are not redirected.</a>
  <script>location="https://developer.okta.com/okta-sdk-java/0.0.4/apidocs/index.html"</script>
</html>

```

jmelberg-okta:
Gotcha. Is there a reason we do not use relative linking as our redirects?

bdemers:
Our lint/link checking doesn't work for files not hosted in this repository, the new location of the javadocs is okta/okta-sdk-java.

bdemers:
@jmelberg-okta After reviewing another commit that used a Jekyll redirect. It looks like the generated files use the absolute path even if we put the relative paths in the `.md` files. I'll update!

Re: ignoring, scanning them is better if at all possible as it we are now redirecting we need to know if something has changed (in how the file in generated or something else)

bdemers:
And now with relative links, coming to CI near you!

PR #1212, opened 2017-08-25
Spa changes

lboyette-okta:


robertdamphousse-okta:
I'm guessing this is an old version, we iterated this quite a bit in recent weeks.

PR #1221, opened 2017-08-25
Limit number of posts contained in RSS feed to prevent FeedBurner error

pauljmartinez:
## Description:

Limit number of posts contained in RSS feed to prevent FeedBurner error

--

### Resolves:
Issue submitted to Okta Marketing queue filed by maziel.martinez@okta.com
FeedBurner will not process any feeds larger than 1MB
https://app.asana.com/0/370676378542055/398673129913955

### DIFF
--- a/_source/feed.xml
+++ b/_source/feed.xml
@@ -8,7 +8,7 @@ layout: null
     <description>{{ site.description | xml_escape }}</description>
     <link>{{ site.url }}</link>
     <atom:link href="{{ site.url }}/feed.xml" rel="self" type="application/rss+xml" />
-    {% for post in site.posts %}
+    {% for post in site.posts limit:15 %}
       <item>
         <title>{{ post.title | xml_escape }}</title>
         <description>{{ post.content | xml_escape }}</description>


pauljmartinez:
@rchild-okta FeedBurner has not updated the feed since June. Maziel asked if we could troubleshoot. FeedBurner won't process any feeds larger than 1MB, so I added a limit to the number of posts in the feed to stay under the cap.
http://feeds.feedburner.com/OktaDeveloperBlog/

jmelberg-okta:
@pauljmartinez We are also passing in the entire content of the post: [feed.xml#L14](https://github.com/okta/okta.github.io/blob/6da8a1a6d7d4877f43b92330616cb11b92a949e8/_source/feed.xml#L14). It probably makes sense to only include the description instead, as noted in #1111.

pauljmartinez:
Hi @jmelberg-okta I'm not aware of a variable called `post.description` but I think `post.excerpt` might give us the same result.

My PR was attempting to not change the feed, i.e. users may be accustomed to reading the entire post in their feed reader of choice, and by switching to `post.excerpt` we'd be requiring them to visit the site to consume that content. Personally, I'm don't have an opinion on which way to go, but I wanted to make sure we didn't overlook this.

Could you please let me know which method would be preferred for shortening the feed? We don't need both.

1. Use page.excerpt
2. Limit number of posts in feed

mraible:
As a reader, I prefer the whole post in RSS. +1 for limiting the number of posts.

jmelberg-okta:
@pauljmartinez & @mraible:

Sounds good to me!

bretterer:
This should have no limited the number of posts, but rather pull an exerpt from the post instead of the entire body, as stated in #1110

PR #1234, opened 2017-09-01
:bug: Widget quickstart fix

jmelberg-okta:
## Description:
- Fix bug in the Sign-In Widget quickstart

### DIFF
--- a/_source/quickstart/widget/default-example.md
+++ b/_source/quickstart/widget/default-example.md
@@ -125,7 +125,7 @@ function callMessagesApi() {
     $.ajax({
         url: 'http://localhost:{serverPort}/api/messages',
         headers: {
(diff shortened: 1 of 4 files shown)


_source/quickstart/widget/default-example.md
```diff
@@ -115,7 +115,7 @@ Please continue down to the next section, Server Setup, to learn about access to
 
 ```javascript
 function callMessagesApi() {
-    const accessToken = oktaSignIn.tokenManager.get("accessToken");
+    const accessToken = oktaSignIn.tokenManager.get("accessToken").accessToken;
```
lboyette-okta: what if there is no accessToken?

We should just add `accessToken.accessToken` in the headers below even though it looks funny.

PR #1235, opened 2017-09-01
password policy GA

mystiberry-okta:
## Description:
- Password Policy (GroupPasswordPolicy) is GA

### Resolves:
[OKTA-139748](https://oktainc.atlassian.net/browse/OKTA-139748)

### DIFF
--- a/_source/_docs/api/resources/policy.md
+++ b/_source/_docs/api/resources/policy.md
@@ -6,8 +6,6 @@ redirect_from: "/docs/getting_started/policy.html"
 
 # Policy API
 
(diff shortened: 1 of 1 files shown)


mystiberry-okta:
@mpeng-okta and @lilyhsiao-okta , the policy API is still marked as EA. So it's unclear to me how the Password Policy can be GA. Don't they have to have the Policy API enabled by support before being able to issue any API requests against the password policy object, even if it is GA?

mpeng-okta:
Good point, @mystiberry-okta! Checking with @edjohns-okta and @srinivasanagandla-okta to be sure, but we should be able to GA the general Policy API as well, since there is effectively no difference for us whether a feature is EA or GA in the policy API, I believe (as opposed to the Lifecycle status in the code).

PR #1236, opened 2017-09-01
Key rollover okta 131959

bellatumaneng-okta:
## Description:
- Exposes auto_key_rollover property for OIDC client in /apps API
- Updates OAuth Credential Object and examples

### Resolves:
* [Issue #131959](#131959)
* [OKTA-131959](https://oktainc.atlassian.net/browse/OKTA-131959)

### DIFF
--- a/_source/_docs/api/resources/apps.md
+++ b/_source/_docs/api/resources/apps.md
@@ -979,7 +979,7 @@ Adds an OAuth 2.0 client application. This application is only available to the 
 | client_id                  | Unique identifier for the client application                   | String                                                                      | TRUE     | TRUE   | TRUE       |
 | client_secret              | OAuth 2.0 client secret string (used for confidential clients) | String                                                                      | TRUE     | FALSE  | TRUE       |
 | token_endpoint_auth_method | Requested authentication method for the token endpoint         | `none`, `client_secret_post`, `client_secret_basic`, or `client_secret_jwt` | FALSE    | FALSE  | FALSE      |
-| auto_key_rotation          | Optional: requested key rotation mode for an OIDC client       | Boolean                                                                     | TRUE     | FALSE  | FALSE      |
+| auto_key_rotation          | Requested key rotation mode                                    | Boolean                                                                     | TRUE     | FALSE  | FALSE      |
 
 ##### Settings
 {:.api .api-request .api-request-params}
@@ -1099,7 +1099,7 @@ curl -v -X POST \
     "oauthClient": {
       "client_id": "0oa1hm4POxgJM6CPu0g4",
       "client_secret": "5jVbn2W72FOAWeQCg7-s_PA0aLqHWjHvUCt2xk-z",
-      "token_endpoint_auth_method": "client_secret_post"
+      "token_endpoint_auth_method": "client_secret_post",
       "auto_key_rotation": true
     }
   },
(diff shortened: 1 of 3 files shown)


_source/_docs/api/resources/apps.md
```diff
@@ -1098,6 +1100,7 @@ curl -v -X POST \
       "client_id": "0oa1hm4POxgJM6CPu0g4",
       "client_secret": "5jVbn2W72FOAWeQCg7-s_PA0aLqHWjHvUCt2xk-z",
       "token_endpoint_auth_method": "client_secret_post"
+      "auto_key_rotation": true
```
mystiberry-okta: Wondering if true is supposed to have quotations around it? github thinks something about this line is incorrect (github isn't always right!)

_source/_docs/api/resources/apps.md
```diff
@@ -4759,12 +4763,15 @@ Determines how to authenticate the OAuth 2.0 client
 
 * The `client_id` must consist of alphanumeric characters or the following special characters `$-_.+!*'(),`. It must contain between 6 and 100 characters, inclusive, and must not be the reserved word `ALL_CLIENTS`. The `client_secret` must consist of printable characters, which are defined in [the OAuth 2.0 Spec](https://tools.ietf.org/html/rfc6749#appendix-A), and must contain between 14 and 100 characters, inclusive.
 
+* If `auto_key_rotation` is not specified, the client will automatically opt in for Okta's [key rotation](/docs/api/resources/oidc.html#validating-id-tokens). This property may be updated via the the API and via the Okta admin UI.
```
mystiberry-okta: we try to avoid future tense unless it's necessary, so I recommend "the client automatically opts in..."
Also, might be more natural to say "via the API or via the Okta Admin user interface" since we mean 'either one'. Note caps for Admin, we're just copying the admin doc in this convention. We should always spell out "user interface".

_source/_docs/api/resources/apps.md
```diff
@@ -4751,6 +4754,7 @@ Determines how to authenticate the OAuth 2.0 client
 | client_id                  | Unique identifier for the OAuth 2.0 client application                           | String   | TRUE     |
 | client_secret              | OAuth 2.0 client secret string                                                   | String   | TRUE     |
 | token_endpoint_auth_method | Requested authentication method for the token endpoint                           | String   | FALSE    |
+| auto_key_rotation          | Optional: requested key rotation mode for an OIDC client                         | Boolean  | TRUE     |
```
mystiberry-okta: Isn't this for an OAuth 2.0 client? (I could be confused).

_source/_docs/api/resources/apps.md
```diff
@@ -979,6 +979,7 @@ Adds an OAuth 2.0 client application. This application is only available to the
 | client_id                  | Unique identifier for the client application                   | String                                                                      | TRUE     | TRUE   | TRUE       |
 | client_secret              | OAuth 2.0 client secret string (used for confidential clients) | String                                                                      | TRUE     | FALSE  | TRUE       |
 | token_endpoint_auth_method | Requested authentication method for the token endpoint         | `none`, `client_secret_post`, `client_secret_basic`, or `client_secret_jwt` | FALSE    | FALSE  | FALSE      |
+| auto_key_rotation          | Optional: requested key rotation mode for an OIDC client       | Boolean                                                                     | TRUE     | FALSE  | FALSE      |
```
mystiberry-okta: OIDC or OAuth 2.0 or both?

mystiberry-okta: Also, if it's nullable, that means it's optional, right? Not sure we want to call out Optional in the description since we don't do it for other optional properties (rather we rely on nullable value). If I'm wrong about that, let me know!

_source/_docs/api/resources/apps.md
bellatumaneng-okta: ah, there's supposed to be a comma after `"client_secret_post"` so `true` isn't registering as a boolean...

_source/_docs/api/resources/apps.md
bellatumaneng-okta: Yes - the way they're implemented, they're the same in terms of app instances, so I didn't actually have to specify OIDC here.

mystiberry-okta:
@bellatumaneng-okta I restarted that top Travis test, the build error looked like a transient problem. If it doesn't resolve as passed, let me know and I'll get the DevEx team to follow up

PR #1255, opened 2017-09-08
Fix incorrect postman example of SMS factor enrollment

robertjd:
## Description:

Fix incorrect postman example of SMS factor enrollment

Resolves: OKTA-140818


mystiberry-okta:
Just FYI these changes won't be live until I republish them to a collection. I'll create a JIRA and post it in description for both the change and the publication.

mystiberry-okta:
There's nothing to build for this change. I'll update the links to this collection in a separate PR.

PR #1259, opened 2017-09-11
97529 he social links

zdevil:
## Description:
- Move author social icons below author
- Add Addthis (Sharing) to blog post
- Change RSS link from OktaBlog to OktaDeveloperBlog

The documentation says to build for production if the changes are related to a blog post. 
Does it include changes that are _not_ a new blog post? Should we also squash the commits?

Happy to make any necessary changes.

Thanks
HE


mystiberry-okta:
@zdevil You do have to build the production files, or base on `weekly` (and the changes will show up in the next release, which is usually a Wednesday).  I'm going to defer to @rchild-okta for a review since these are functional changes. A squash happens at merge time, so you don't need to do that (but are welcome to).

PR #1261, opened 2017-09-12
Change global JS links location

jmelberg-okta:
## Description:
Previously, the documentation (index) and quickstart pages had global scripts running on every page. To enhance performance:
- Move uncommon `.js` scripts to load only on desired page.

### DIFF
--- a/_source/_assets/js/quickstart.js
+++ b/_source/_assets/js/quickstart.js
@@ -313,8 +313,7 @@
     $('#server_setup_link').removeClass('active');
   };
 
-  if (window.location.pathname.match('^/quickstart')) {
-    main();
-  }
+  // Load the quickstart partials
+  main();
 
 })(jQuery);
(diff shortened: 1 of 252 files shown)


_source/_assets/js/quickstart.js
```diff
@@ -68,10 +68,6 @@
             name: 'spring',
             label: 'Spring',
             active: true
-          },
-          {
-            name: 'generic',
```
lboyette-okta: if this is removed, we need to change the links here: https://raw.githubusercontent.com/okta/okta.github.io/master/_source/_code/java/index.md

jmelberg-okta: Removing this also removed the implicit flow example. Will add in another PR.

_source/_layouts/docs_index.html
```diff
@@ -1,6 +1,7 @@
 ---
 layout: master
 show_bottom_cta: true
+js: docsIndex
```
lboyette-okta: didn't know you could do this. neat

PR #1262, opened 2017-09-12
Update widget implicit quickstart

robertjd:
## Description:

* The previous example would return a session token, not an access/id token pair.  Example is now updated to do this.
* Example is also updated to redirect through Okta to get the session token, so that Safari can get the SSO cookie stored
* Cleaned up example code to use 2 spaces, and not use const for cross-browser concerns

### Resolves:
* [OKTA-140258](https://oktainc.atlassian.net/browse/OKTA-140258)

### DIFF
--- a/_source/quickstart/widget/default-example.md
+++ b/_source/quickstart/widget/default-example.md
@@ -3,21 +3,23 @@ layout: quickstart_partial
 libraryName: Sign-In Widget
 ---
 
-This guide will walk you through integrating the [Okta Sign-In Widget](https://github.com/okta/okta-signin-widget) into an app by performing these steps:
+This guide will walk you through integrating the [Okta Sign-In Widget](https://github.com/okta/okta-signin-widget) into a front-end application by performing these steps:
 
 1. Add an OpenID Connect Client in Okta
 2. Add Sign-In Widget assets to your project
 3. Implement Okta Sign-In
 4. Using the Access Token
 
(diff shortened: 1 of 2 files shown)


_source/quickstart/widget/default-example.md
```diff
@@ -70,66 +74,95 @@ First, create a `<div>` inside of your HTML file:
...
+      issuer: "https://{yourOktaDomain}/oauth2/default",
+      responseType: ['token', 'id_token'],
+      display: 'page'
+    }
+  });
+
+  if (oktaSignIn.token.hasTokensInUrl()) {
+    oktaSignIn.token.parseTokensFromUrl(
+      function success(res) {
+        oktaSignIn.tokenManager.add('accessToken', res[0]);
```
jmelberg-okta: We should keep the comment here on the order of the tokens.

_source/quickstart/widget/default-example.md
```diff
@@ -70,66 +74,95 @@ First, create a `<div>` inside of your HTML file:
...
-        return;
+  var accessToken = oktaSignIn.tokenManager.get("accessToken");
+
+  if (!accessToken) {
+    return;
+  }
+
+  // Make the request using jQuery
+  $.ajax({
+    url: 'http://localhost:3000/api/messages',
```
jmelberg-okta: `http://localhost:3000` -> `http://localhost:{serverPort}`

PR #1264, opened 2017-09-12
Add query param cachebusting to external css url

benbridge:
## Description:
- Add query string cache busting to external styles
Adding the query string to the external css file on okta.com caused the 'find-missing-slashes' script to mark the link as invalid and tests to fail, it is now excluded from the checks.

See https://github.com/okta/okta.github.io/pull/1180#issuecomment-324129815 for more context.

### Resolves
Link to external stylesheet has no cache busting since it does not use "jekyll-assets".


_source/_includes/head.html
```diff
@@ -38,7 +38,7 @@
  		<meta name="robots" content="noindex,nofollow">
  	{% endif %}
 
-  <link type="text/css" rel="stylesheet" href="{{ site.external_domain }}/sites/all/themes/developer/css/master.css" media="all" />
+  <link type="text/css" rel="stylesheet" href="{{ site.external_domain }}/sites/all/themes/developer/css/master.css?{{site.time | date: '%s%N'}}" media="all" />
```
rchild-okta: Am curious - will this actually help in this case? This is based off of changes that happen locally in this repo, but we need to have a different query param value whenever the external css changes

benbridge: You're right, the query param will only change when this site is rebuilt, so it's not an ideal fix. My original thinking was that this would be better than having no cache busting, especially when changes are made in the okta.com repo and this repo.

PR #1266, opened 2017-09-12
Updated Angular quickstart

jmelberg-okta:
## Description:
- Updates the Angular quickstart to use the new `@okta/okta-angular` SDK

### Resolves:
* [OKTA-141102](https://oktainc.atlassian.net/browse/OKTA-141102)

### DIFF
--- a/_source/quickstart/angular/default-example.md
+++ b/_source/quickstart/angular/default-example.md
@@ -6,8 +6,8 @@ exampleDescription: Angular Implicit
 This guide will walk you through integrating authentication into an Angular app with Okta by performing these steps:
 
 1. Add an OpenID Connect Client in Okta
-2. Create an Authentication Service
-3. Create Routes
+2. Install the Okta Angular SDK
+3. Attach Components to Routes
 4. Use the Access Token
 
 At the end of the Angular instructions you can choose your server type to learn more about post-authentication workflows, such as verifying tokens that your Angular application can send to your server.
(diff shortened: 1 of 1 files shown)


_source/quickstart/angular/default-example.md
```diff
@@ -134,23 +92,38 @@ export class AppComponent {
...
+    component: OktaCallbackComponent
+  },
+  ...
+]
+
+@NgModule({
+  ...
+  declarations: [
+    ...
+    ProtectedComponent
```
jmelberg-okta: OktaCallbackComponent

_source/quickstart/angular/default-example.md
```diff
@@ -134,23 +92,38 @@ export class AppComponent {
...
+  ]
+})
```
robertjd: Not sure what this is?  I doesn't seem like we need to declare anything.

_source/quickstart/angular/default-example.md
```diff
@@ -134,23 +92,38 @@ export class AppComponent {
 ```
 
 ### Create the Callback Handler
-In order to handle the redirect back from Okta, you need to capture the token values from the URL. You'll use `/implicit/callback` as the callback URL, and again you'll use our `OktaAuthService` to delegate the callback details to the [Okta Auth JS](/code/javascript/okta_auth_sdk) library.
-
-Create a new component `callback.component.ts`, and have it handle the `/implicit/callback` route:
+In order to handle the redirect back from Okta, you need to capture the token values from the URL. You'll use `/implicit/callback` as the callback URL, and specify the default `OktaCallbackComponent` and declare it in your `NgModule`.
```
robertjd: We should mention "In these examples we use the [Angular Router](https://angular.io/guide/router)", the link being important so that they can go read about how to set up routing.  The default angular example app does not have routing in it.

_source/quickstart/angular/default-example.md
```diff
@@ -32,83 +32,41 @@ After you have created the application there are two more values you will need t
...
+  issuer: 'https://{yourOktaDomain}.com/oauth2/default',
+  redirectUri: 'http://localhost:{port}/implicit/callback',
+  clientId: '{clientId}'
 }
 ```
 
-## Create Routes
+## Attach Components to Routes
 
 You'll need to provide these routes in your sample application, so that we can sign the user in and handle the callback from Okta:
```
robertjd: In this section, the bullet point about the implicit/callback route should mention "We'll set this up below", to clue me into the fact that you're going to tell me how to do this (as currently written, I feel like I'm supposed to go do this now).

_source/quickstart/angular/default-example.md
jmelberg-okta: Added this to the intro of the **Attach Components** to Routes section

PR #1290, opened 2017-09-22
Replace {yourOktaDomain} with developer account

jmelberg-okta:
## Description:
-  Replaces all `https://{yourOktaDomain}.com`  occurrences throughout the site with the the most recently used ***.oktapreview** <https://login.okta.com> org session.

> Note: Requires pending feature add.

See: #1260


_source/_assets/js/myOkta.js
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    function findPreviewOrg(account) {
+      // Extract the first 'oktapreview' domain
+      if (account.origin.indexOf('.oktapreview.com') !== -1){
+        return account.origin;
+      }
+    }
+
+    var previewOrg = event.data.find(findPreviewOrg);
```
jeromelecomte-okta: we need polyfilling since IE doesn't necessarily support Array.find. (see https://github.com/okta/my-okta/blob/master/lib/environment.js#L26, this might not be the best way, but it proves the point we had to do it in My Okta to avoid uncaught errors in IE)

_source/_assets/js/myOkta.js
```diff
@@ -0,0 +1,33 @@
...
+
+    $('.okta-preview-domain').text(previewOrg || 'https://{yourOktaDomain}.com');
+
+      // if (account.origin.indexOf('.oktapreview.com') !== -1){
+      //   document.body.innerHTML = document.body.innerHTML.replace(/https:\/\/{yourOktaDomain}.com/g, account.origin);
+      //   return;
+      // }
```
jeromelecomte-okta: please remove commented out code

_source/_layouts/base.html
```diff
@@ -14,5 +14,7 @@
   <!-- START Post Footer -->
   {% include post-footer.html %}
   <!-- END Post Footer -->
+  <!-- Hidden iframe -->
+  <iframe id="myOkta" src="https://login.okta.io:8080" style="display:none;"></iframe>
```
jeromelecomte-okta: will this work in prod?

_source/_assets/js/myOkta.js
jmelberg-okta: I pushed a commit here while it was still very much in progress :) Good catch!

_source/_assets/js/myOkta.js
```diff
@@ -0,0 +1,33 @@
+(function($) {    
+  $(window).load(function() {
```
rchild-okta: More common to just use [$.ready](http://api.jquery.com/ready/), which can be invoked like:

```javascript
// Note, there are different ways to call $.ready, but this is the syntax they recommend
$(function () {
  // some stuff you want to run after the dom is ready
})
```

rchild-okta: And I think in this case, we can just load the iframe in this JS rather than adding it to the html. Similar to what we're doing in the [AccountChooser](https://github.com/okta/okta-core/blob/master/clients/loginpage/src/AccountChooser.js) code.

After adding it to the page, we should wait for it to [load](http://api.jquery.com/load-event/) - don't want to send the event too early.

And also, we should use https://login.okta.com/discovery/iframe.html as the source - this does not load all the extra stuff we don't need.

_source/_assets/js/myOkta.js
```diff
@@ -0,0 +1,33 @@
...
+    // and make a postMessage request
+    var iframe = document.getElementById('myOkta');
+    if (!iframe) { return; }
+    window.addEventListener('message', receiveMessage, false);
+    iframe.contentWindow.postMessage({messageType: 'get_accounts_json'}, 'https://login.okta.com');
+  });
+
+  function receiveMessage(event) {
+    // Verify the event origin is trusted
+    if (event.origin !== 'https://login.okta.com' && !event.data) {
```
rchild-okta: ||, not &&

_source/_assets/js/myOkta.js
```diff
@@ -0,0 +1,33 @@
...
+      return;
+    }
+
+    function findPreviewOrg(accounts) {
+      // Extract the first 'oktapreview' domain
+      for (i=0; i< accounts.length; i++) {
```
rchild-okta: var i = 0; i < accounts.length <-- add the var, and spacing

_source/_assets/js/myOkta.js
```diff
@@ -0,0 +1,38 @@
+(function($) {    
+  $(function() {
+    // Once the page is loaded, find the hidden iframe element
```
rchild-okta: Nit: can remove this comment since it doesn't apply anymore

_source/_assets/js/myOkta.js
```diff
@@ -0,0 +1,38 @@
...
+    // and make a postMessage request
+
+    var iframe = $('<iframe id="myOktaIFrame" src="https://login.okta.com/discovery/iframe.html" style="display:none"></iframe>').get(0);
+
+    $('body').append(iframe);
+
+    $('#myOktaIFrame').load(function() {
```
rchild-okta: Don't need to search for it if we don't do the get(0) up front - can just use the reference you already have (`iframe`). Probably easier to do something like:

```js
var iframe = $('stuff'); // no .get(0)
$('body').append(iframe);
iframe.load(function () {
  iframe.get(0).contentWindow // etc
});
```

PR #1294, opened 2017-09-25
Weekly now source

bdemers:
Merge's source into `weekly`. 

Going forward (because there is not generated content in this branch anymore), we can just `git merge source`, and create a PR.  (no more force pushing should be needed)


mystiberry-okta:
🚢

PR #1301, opened 2017-09-26
Add query param cachebusting to external css url #1264

bdemers:
Originally was: #1264


bdemers:
@benbridge Is this still needed ?

bretterer:
Vote to close at this point

PR #1311, opened 2017-09-27
Update react quickstart

jmelberg-okta:
## Description:
- Updates the React quickstart to use the new React SDK

### DIFF
--- a/_source/quickstart/react/default-example.md
+++ b/_source/quickstart/react/default-example.md
@@ -6,8 +6,8 @@ exampleDescription: React Implicit
 This guide will walk you through integrating authentication into a React app with Okta by performing these steps:
 
 1. Add an OpenID Connect Client in Okta
-2. Create an Authentication Utility
-3. Create Routes
+2. Install the Okta React SDK
+3. Attach Components to the Secure Router
 4. Use the Access Token
 
 At the end of the React instructions you can choose your server type to learn more about post-authentication workflows, such as verifying tokens that your React application can send to your server.
@@ -19,7 +19,7 @@ At the end of the React instructions you can choose your server type to learn mo
 ## Add an OpenID Connect Client in Okta
 In Okta, applications are OpenID Connect clients that can use Okta Authorization servers to authenticate users.  Your Okta Org already has a default authorization server, so you just need to create an OIDC client that will use it.
 * Log into the Okta Developer Dashboard, click **Applications** then **Add Application**.
-* Choose **Single Page App (SPA)** as the platform, then submit the form the default values, which should look like this:
+* Choose **Single Page App (SPA)** as the platform, then populate your new OpenID Connect application with values suitable for your app. If you are running this locally and using the defaults from the [React Quickstart](https://facebook.github.io/react/docs/installation.html#creating-a-new-application), your `port` will be `3000`:
 
 | Setting             | Value                                        |
 | ------------------- | -------------------------------------------- |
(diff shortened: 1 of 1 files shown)


_source/quickstart/react/default-example.md
```diff
@@ -38,221 +38,150 @@ After you have created the application there are two more values you will need t
...
-  );
-}));
-```
 
-### Create the Callback Handler
-In order to handle the redirect back from Okta, you need to capture the token values from the URL. You'll use `/implicit/callback` as the callback URL, and again you'll use our `auth.js` to delegate the callback details to the [Okta Auth JS](/code/javascript/okta_auth_sdk) library.
+In the relevant location in your application, you will want to provide `Login` and `Logout` buttons for the user. You can show/hide the correct button by using the `auth.isAuthenticated()` method. For example:
 
-Create a new component `Callback.js`, and have it handle the `/implicit/callback` route:
+```javascript
```
lboyette-okta: if we say this `typescript` then we get jsx highlighting

_source/quickstart/react/default-example.md
```diff
@@ -38,221 +38,150 @@ After you have created the application there are two more values you will need t
...
-      if (token.idToken) {
-        this.oktaAuth.tokenManager.add('idToken', token);
-      } else if (token.accessToken) {
-        this.oktaAuth.tokenManager.add('accessToken', token);
-      }
-    }
-  }
+const config = {
+  issuer: 'https://{yourOktaDomain}.com/oauth2/default',
+  redirectUri: window.location.origing + '/implicit/callback',
```
jmelberg-okta: origin

_source/quickstart/react/default-example.md
```diff
@@ -38,221 +38,150 @@ After you have created the application there are two more values you will need t
...
 ```bash
-npm install @okta/okta-auth-js --save
+npm install @okta/okta-react --save
 ```
 
-You will need to create a class that encapsulates the interaction with the [Okta Auth JS](/code/javascript/okta_auth_sdk.html) library. This file will expose a `withAuth` method that makes it easy to create [Higher-Order Components](https://facebook.github.io/react/docs/higher-order-components.html) that include an `auth` property.
+### Configuration
 
-To create this file, you will need the values from the OIDC client that you created in the previous step.  You will also need to know your Okta Org URL, which you can see on the home page of the Okta Developer console.
+You will need the values from the OIDC client that you created in the previous step to instantiate the middleware. You will also need to know your Okta Org URL, which you can see on the home page of the Okta Developer console.
```
lboyette-okta: `You will also need to know your Okta Org URL, which you can see on the home page of the Okta Developer console` - you actually just need to know the issuer

_source/quickstart/react/default-example.md
```diff
@@ -38,221 +38,150 @@ After you have created the application there are two more values you will need t
...
+    return this.state.authenticated ?
+      <button onClick={this.props.auth.logout}>Logout</button> :
+      <button onClick={this.props.auth.login}>Login</button>;
   }
 });
 ```
 
+### Update your `App.js`
+
+Finally, instantiate the Secure Router by passing in your configuration into `Security`, and connect your application's paths:
```
lboyette-okta: hmm, nowhere do we mention a Secure Router, so maybe we should rethink the naming

_source/quickstart/react/default-example.md
```diff
@@ -38,221 +38,150 @@ After you have created the application there are two more values you will need t
...
+
+```javascript
+import React, { Component } from 'react';
+import { BrowserRouter as Router, Route } from 'react-router-dom';
+import { Security, SecureRoute, ImplicitCallback } from '@okta/okta-react';
+import Home from './Home';
+
+const config = {
+  issuer: 'https://{yourOktaDomain}.com/oauth2/default',
+  redirect_uri: window.location.origing + '/implicit/callback',
```
lboyette-okta: origin

PR #1333, opened 2017-10-04
fix requiredness description

mystiberry-okta:
## Description:
- Fix requiredenss of scope.

### Resolves:
[OKTA-OKTA-139564](https://oktainc.atlassian.net/browse/OKTA-139564)


_source/_docs/api/resources/oauth2.md
```diff
@@ -49,24 +49,24 @@ This is a starting point for OAuth 2.0 flows such as implicit and authorization
...
+| response_mode         | Specifies how the authorization response should be returned. 







[Valid values: ``fragment``, ``form_post``, ``query`` or ``okta_post_message``](#request-parameter-details). If ``id_token`` or ``token`` is specified as the *response_type*, then ``query`` isn&#8217;t allowed as a *response_mode*. Defaults to ``fragment`` in implicit and hybrid flows. Defaults to ``query`` in authorization code flow and cannot be set as ``okta_post_message``. | Query | String   | FALSE            | See Description. |
+| scope                 | Can be a combination of reserved scopes and custom scopes. The combination determines the claims that are returned in the Access Token and ID Token. The ``openid`` scope has to be specified to get back an ID Token. If not specified at all, the default scopes configured in the Custom Authorization Server are used.                                                                                                                                | Query | String   | See Description. |                  |
```
johngronberg-okta: We may want to be more clear here: if no scope is included on the request it will fail unless a default scope is specified on the auth server.

johngronberg-okta:
👍

mystiberry-okta:
@johngronberg-okta on this branch (source) you have to use the official Review control (green Review button in upper right corner).

_source/_docs/api/resources/oauth2.md
```diff
@@ -49,24 +49,24 @@ This is a starting point for OAuth 2.0 flows such as implicit and authorization
...
-| idp_scope             | A space delimited list of scopes to be provided to the Social Identity Provider when performing [Social Login](social_authentication.html). These scopes are used in addition to the scopes already configured on the Identity Provider.                                                                                                                                                                                                                 | Query | String   | FALSE    |                  |
+| Parameter             | Description                                                                                                                                                                                                                                                                                                                                                                                                                                               | Type  | DataType | Required         | Default          |
```
jmelberg-okta: Do we need the `Default` column here? Looks like it is mostly empty.

mystiberry-okta: yes--there are defaults for three of the values, and they are complicated, so we can't put the default in the column (that's why the pointers).

mystiberry-okta: I could take out the Type and DataType columns out and simply say in a sentence before the table that it's query and string...

mystiberry-okta: Following up with John, will do in a separate jira if he thinks it's okay.

jmelberg-okta: I think the `Type` and `DataType` sections are fine -- it just looks a bit odd seeing a white space in the `Default` column. Especially when we point them back to another column -> "See description". But I understand the need here.

PR #1349, opened 2017-10-10
Adds Android quickstart

jmelberg-okta:
## Description:
- Adds a quickstart example using the new [Okta AppAuth Android](https://github.com/okta/okta-sdk-appauth-android) SDK.

Android

### Resolves:

* [OKTA-140044](https://oktainc.atlassian.net/browse/OKTA-140044)


_source/quickstart/android/default-example.md
```diff
@@ -3,4 +3,191 @@ layout: quickstart_partial
...
+### Update the URI Scheme
+In order to redirect back to your application from a web browser, you must specify a unique URI to your app. To do this, you must define a gradle manifest placeholder in your app's `build.gradle`:
+
+```
+android.defaultConfig.manifestPlaceholders = [
+    "appAuthRedirectScheme": "com.okta.example"
+]
+```
+
+Make sure this is consistent with the redirect URI used in `okta_app_auth_config.json`. For example, if your **Login Redirect URI** is `com.okta.example:/callback`, the **AppAuth Redirect Scheme** will be `com.okta.example`.
```
bdemers: `com.okta.example:/callback` ?

jmelberg-okta: Yep. The **Redirect Scheme** omits the path.

_source/quickstart/android/default-example.md
```diff
@@ -3,4 +3,191 @@ layout: quickstart_partial
...
+            }
+        );
+    }
+}
+```
+
+Once the `OktaAppAuth` instance is initialized, you can start the authorization flow by simply calling `login` whenever you're ready:
+
+```java
+private void startAuth() {
```
bdemers: Where does this method go?  If it is just a wrapper around the below code we can probably just pull it out of the method.

_source/quickstart/android/default-example.md
```diff
@@ -3,4 +3,191 @@ layout: quickstart_partial
...
+
+## Handle the Login State
+In native applications, it is common for users to have a long-lived session. It is important for the app to manage the user's session by refreshing tokens when they expire, using the `refresh_token` or re-prompting the user to login. See [refreshing a token manually](https://github.com/okta/okta-sdk-appauth-android#refresh-a-token-manually) for more information.
+
+When starting up the application, check for the existance of an `access_token` to see if the user has an existing session:
+
+```java
+// LoginActivity.java
+
+protected void onCreate(Bundle savedInstanceState) {
```
bdemers: Why protected? should this be marked with an `@Override` ?

_source/quickstart/android/default-example.md
```diff
@@ -3,4 +3,191 @@ layout: quickstart_partial
...
+```
+
+## Using the Access Token
+
+Your Android application now has an access token in private Shared Preferences that was issued by your Okta Authorization server. You can use this token to authenticate requests for resources on your server or API. As a hypothetical example, let's say that you have an API that gives us messages for our user.  You could create a `callMessagesApi` function that makes an authenticated request to your server.
+
+Please continue down to the next section, Server Setup, to learn about access token validation on the server.  Here is what the request could look like for this hypothetical example:
+
+```java
+private void callMessagesApi() {
```
bdemers: probably drop the wrapping method, and just show me the code (unless there is a specific place that these `callMessageApi` method is called from

_source/quickstart/android/default-example.md
```diff
@@ -3,4 +3,191 @@ layout: quickstart_partial
...
+        }
+
+        @Override
+        public void onFailure(int i, Exception e) {
+            // Handle failure to make your authorized request or a response with a 4xx or
+            // 5xx HTTP status response code
+        }
+    });
+}
+```
```
bdemers: Closing section?

_source/quickstart/android/default-example.md
bdemers: This might just be some androidy thing that I don't fully understand :D

_source/quickstart/android/default-example.md
jmelberg-okta: We explicitly call out creating a `callMessagesApi` method above, so it makes sense to include the function declaration here.

jmelberg-okta:
@bdemers Added a closing note to both the **iOS** and **Android** quickstarts. Good catch!

Gemfile.lock
```diff
@@ -244,7 +244,7 @@ PLATFORMS
 DEPENDENCIES
   github-pages
   html-proofer
-  jekyll-assets
+  jekyll-assets (= 2.3.2)
```
bdemers: likely conflict from @jmelberg-okta last post

_source/quickstart/android/default-example.md
bdemers: I noticed, I just don't think it is clear how/where the `callMessagesApi` method should be called, I think it would be more obvious that it was up to the user's discretion if it wasn't wrapped in a method (it is just code to stick in your application somewhere)

That said, it looks like the iOS example is structured the same way, so I don't want to block on this. I'm a [0](https://www.apache.org/foundation/voting.html#expressing-votes-1-0-1-and-fractions) on that topic, but I'm +1 for everything else

PR #1356, opened 2017-10-11
Added missing html redirects back in

bdemers:
Generated mostly from d91bb31599764210fddccb3f3de26eaaafe4783d and the following script (actual script not important but included here for record)

```bash
#!/bin/bash
REDIRECTS=$(find . -type f | grep '\.html'| grep -v /javadoc/ |grep -v java_api_sdk | xargs grep -l 'http-equiv="refresh" content="0; url=' |sort |uniq | xargs grep 'http-equiv="refresh" content="0' | sed 's_<meta http-equiv="refresh" content="0; url=__' | sed 's_">$__' | sed 's_https://developer.okta.com__')

DEST_DIR="$HOME/dev/source/okta/blog/_source/_docs/"

while read -r line; do
    IFS=':  ' read -r -a ITEM <<< "$line"
    CURRENT_FILE="${DEST_DIR}/$(echo ${ITEM[0]} | sed 's_\.html_\.md_')"
    NEW_FILE="${ITEM[1]}"
    echo "$CURRENT_FILE"
    
    mkdir -p $(dirname $CURRENT_FILE)
    
cat > "$CURRENT_FILE" <<EOF
---
redirect_to: ${NEW_FILE}
---
EOF

done <<< "$REDIRECTS"
```

### DIFF
--- a/_source/_docs/api/rest/apps.md
+++ b/_source/_docs/api/rest/apps.md
@@ -1,0 +1,3 @@
+---
+redirect_to: /docs/api/resources/apps.html
+---
--- a/_source/_docs/api/rest/search-beta.md
+++ b/_source/_docs/api/rest/search-beta.md
@@ -1,0 +1,3 @@
+---
+redirect_to: /docs/api/resources/search-beta
+---
--- a/_source/_docs/api/rest/search-beta.md
+++ b/_source/_docs/api/rest/search-beta.md
@@ -1,3 +1,0 @@
----
-redirect_to: /docs/api/resources
----
--- a/_source/_docs/api/rest/search-beta.md
+++ b/_source/_docs/api/rest/search-beta.md
@@ -1,3 +1,3 @@
 ---
-redirect_to: /docs/api/resources/search-beta
+redirect_to: /docs/api/resources
 ---
--- a/_source/_docs/guides/scim_developer_program.md
+++ b/_source/_docs/guides/scim_developer_program.md
@@ -1,0 +1,3 @@
+---
+redirect_to: /standards/SCIM/scim_developer_program
+---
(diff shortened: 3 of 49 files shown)


_source/_docs/guides/scim_developer_program.md
```diff
@@ -1,3 +1,3 @@
 ---
-redirect_to: /standards/SCIM/scim_developer_program
+redirect_to: /standards/SCIM/index
```
jmelberg-okta: Can this just be: `standards/SCIM/`?

_source/_docs/api/rest/search-beta.md
```diff
@@ -1,3 +1,3 @@
 ---
-redirect_to: /docs/api/resources/search-beta
+redirect_to: /docs/api/resources
```
jmelberg-okta: I think this is safe to remove. Anything with `/rest` was been updated > 1 year ago to new routes.

_source/_docs/guides/scim_developer_program.md
bdemers: Yes, but we already had another existing redirect that used `/index`, so i wanted to keep them the same.

_source/_docs/api/rest/search-beta.md
bdemers: 👍

_source/_docs/api/rest/apps.md
```diff
@@ -0,0 +1,3 @@
+---
+redirect_to: /docs/api/resources/apps.html
```
jmelberg-okta: All the `/rest` paths can be removed. @mystiberry-okta and I talked about this a few months back.

PR #1357, opened 2017-10-11
Change date on docker-compose post so it comes first

mraible:
## Description:

Change date on docker-compose post so it comes first

Currently, it comes after my post that was published yesterday.

screen shot 2017-10-11 at 2 47 20 pm


bdemers:
ugg, dates in URLs... we should probably add a redirect

```text
---
redirect_to: /some/place/great
---
```

mraible:
Where do I add this, in the post itself, or create a new page (that matches the old page name) that redirects to this one?

bdemers:
Option B, New page that matches the old file name, that redirects to the new one

mraible:
Please review again @bdemers. I tried to test locally, but it redirects to https://developer.okta.com/blog/2017/10/11/developers-guide-to-docker-part-3 (which looks right), but is a 404 locally.

PR #1372, opened 2017-10-13
getting 'weekly' inline with 'source'

bdemers:
Looks like something happened where 'master' was pulled in and messed up the history.
This commit puts 'weekly' back on par with 'source' (without doing a force push)

I could not figure out when/where this happened  (though I'm guessing this was caused by a force push)


_source/_code/angular/okta_angular_auth_js.md
```diff
@@ -37,15 +37,15 @@ ng new okta-app
...
-The simplest way to add authentication into an Angular app is using the library [Okta Auth JS](/code/javascript/okta_auth_sdk). We can install it via `npm`:
+The simplest way to add authentication into an Angular app is using the library [Okta Auth JS](/code/javascript/okta_auth_sdk.html). We can install it via `npm`:
 
 ```bash
 npm install @okta/okta-auth-js --save
 ```
 
 ## Create an Authentication Service
 Users can sign in to your Angular application a number of different ways.
-The easiest, and most secure way is to use the **default login page**. This page renders the [Okta Sign-In Widget](/code/javascript/okta_sign-in_widget), equipped to handle User Lifecycle operations, MFA, and more.
```
jakubvul-okta: FWIW: This shouldn't make any difference.

jmelberg-okta: @jakubvul-okta It actually does. Our link checker verifies that the relative URLs contain either a `/` or end in `.html`.

jakubvul-okta: Ohhh, ok. I just meant from a URL resolution standpoint. But good to know!

PR #1375, opened 2017-10-13
Remove the Travis-CI 'pages' provider in favor of a script that does not force push

bdemers:
Result will look like commits on: [`master-test`](https://github.com/okta/okta.github.io/commits/master-test)

This implementation is pretty basic, it makes assumptions about how the `dist` directory is setup (which since this is CI it will always be the same way)

If we are happy with this change, we can make future improvements to remove the git-fu from the `build-prod` step, (which will save dev time)


scripts/gh-pages-publish.sh
```diff
@@ -0,0 +1,26 @@
...
+pushd "${GENERATED_SITE_LOCATION}"
+
+GIT_AUTHOR_EMAIL="deploy-docs@okta.com"
+GIT_AUTHOR_NAME="Auto Deploy Docs"
+
+
+GITHUB_SLUG=okta/okta.github.io
+CURRENT_BRANCH=$(git rev-parse --abbrev-ref HEAD)
+
+REMOTE_NAME=gh-pages
```
jmelberg-okta: What if this branch gets deleted? (Hopefully this never happens, but you know- it might). We *should* add in the branch creation if it doesn't exist.

bdemers: possibly, but if the branch is deleted the publish will fail (which is good) and then we could recover by creating the branch and then re-running the build.

This script should allow us to protect the 'master' branch so it _shouldn't_ get removed.

bdemers:
moved to 1391

PR #1377, opened 2017-10-13
Fail build if Node, npm, or Ruby versions are not correct

bdemers:
idea, to help make envs consistent

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,7 @@
     "dist"
   ],
   "scripts": {
+    "postinstall": "check-node-version --node $(cat .nvmrc) --npm 4.2 && ruby --version | grep -q 2.2.4 || (echo 'Invalid ruby version, expecting 2.2.4' && exit 1)",
     "clean": "node scripts/clean",
     "pregenerate": "npm run clean",
     "generate": "cross-env PID_FINDER=okta_github_io JEKYLL_ENV=production bundle exec jekyll",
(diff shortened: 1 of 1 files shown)


bdemers:
Any version range introduces instability in the build, I'm all for locking it down to start with. Get our environments stable, and then maybe move to supporting patch versions, for example our `.nvmrc` file is just `7.9`, that said, the node version is less important then the ruby version, so I'm mostly indifferent.

jmelberg-okta:
Agree. Lets lock down the `Ruby` version, as that causes most of our issues. If we want to lock down our version of `node`, lets set it to `8.x` and `npm` `5.x` so we use the `package.lock` file for dependency management.

bdemers:
@jakubvul @mystiberry-okta Have either of you had problems with diffs in the lock files recently ?  If not I think I'm just going to close this

jakubvul-okta:
Nope, they've been trouble-free.

bdemers:
OK, lets keep them all in there then.  Switching node version should be a different PR, but I did use `cat .nvmrc` instead of hard coding it (the npm version is relaxed from 4.2.0 to 4.2)

I'm all for updating node/npm but that needs to run through its own CI cycle first, i have no idea if that is going to cause any issues. 

**NOTE:** current version checks from from the related dot files:
.ruby-version -> ruby-2.2.4
.nvmrc ->7.9
(We use the npm that is distributed with node, but a different version could be found on a user's path, which is why we check both versions)

bdemers:
@jmelberg-okta thoughts ?

jmelberg-okta:
Now that we've locked down all of our dependencies, and we aren't running into conflicts like we were before -- we can probably just close this PR. I think forcing a specific environment was only necessary during the build-step.

bdemers:
Sounds good

PR #1382, opened 2017-10-13
rsync the original images to dist

bdemers:
Chicken & egg
In order to validate one of the html pages that contains an image (content pulled from another github repo)
We need to add the raw images files back in place.  This has to get merged/deployed BEFORE we can validate external links


package.json
```diff
@@ -10,7 +10,7 @@
   ],
   "scripts": {
     "clean": "node scripts/clean",
-    "build-prod": "npm run clean && git fetch --force origin master:master && git clone -b master . dist && cross-env JEKYLL_ENV=production bundle exec jekyll build --verbose",
+    "build-prod": "npm run clean && git fetch --force origin master:master && git clone -b master . dist && rsync -r _source/_assets/img/ ./dist/assets/img/ && cross-env JEKYLL_ENV=production bundle exec jekyll build --verbose",
```
jmelberg-okta: Can we split this into two scripts? You can use the `pre` script that will execute before the actual script.

```json
"prebuild-prod": "...",
"build-prod": "...",
```

bdemers: Great idea! but lets move that one into a followup for: #1375

Which I may or may not have ready by the time you see this, so ping me :)

bdemers:
`npm test` tramples the directory, so this will not fix the issue

PR #1390, opened 2017-10-18
Fix incorrect example

joelfranusic-okta:
_sigh_


joelfranusic-okta:
Also update links to binaries to point to the latest version.

PR #1391, opened 2017-10-18
Keep the old bits around when deploying via travis

bdemers:


bdemers:
Similar to #1375

package.json
```diff
@@ -10,11 +10,12 @@
   ],
   "scripts": {
     "clean": "node scripts/clean",
-    "build-prod": "npm run clean && git fetch --force origin master:master && git clone -b master . dist && cross-env JEKYLL_ENV=production bundle exec jekyll build --verbose",
+    "build-prod": "npm run clean && cross-env JEKYLL_ENV=production bundle exec jekyll build --verbose && npm run post-build-prod",
+    "post-build-prod": "bash scripts/post-build-prod.sh",
```
jmelberg-okta: You can simplify this by changing the name to `postbuild-prod`. Therefore it gets called directly after the `build-prod` command.

```
"build-prod": "npm run clean && cross-env JEKYLL_ENV=production bundle exec jekyll build --verbose",
"postbuild-prod": "bash scripts/post-build-prod.sh",
```

package.json
```diff
@@ -10,11 +10,12 @@
...
     "start": "npm run clean && cross-env JEKYLL_ENV=production bundle exec jekyll serve --incremental",
     "stop": "if-env CI=true && exit 0 || (pkill -f clean && kill $(lsof -t -i:4000; lsof -t -i:4444))",
     "test-only": "protractor tests/selenium/conf.js",
-    "pretest": "npm start & wait-on http://localhost:4000 && webdriver-manager update --gecko false",
+    "pretest": "npm run clean && cross-env JEKYLL_ENV=production bundle exec jekyll serve --no-watch & wait-on http://localhost:4000 && webdriver-manager update --gecko false",
```
jmelberg-okta: Some of these scripts are getting confusing.

Suggestion:

```
  "scripts": {
    "clean": "node scripts/clean",
    "pregenerate": "npm run clean",
    "generate": "cross-env JEKYLL_ENV=production bundle exec jekyll",
    "build-prod": "npm run generate build -- --verbose",
    "postbuild-prod": "bash scripts/post-build-prod.sh",
    "start": "npm run generate serve -- --incremental",
    "stop": "if-env CI=true && exit 0 || (pkill -f clean && kill $(lsof -t -i:4000; lsof -t -i:4444))",
    "test-only": "protractor tests/selenium/conf.js",
    "pretest": "npm run generate serve -- --no-watch & wait-on http://localhost:4000 && webdriver-manager update --gecko false",
    "test": "npm run test-only || (npm stop && exit 1)",
    "posttest": "npm stop",
    "import-external": "node scripts/import-external 2>&1",
    "find-missing-slashes": "node scripts/find-missing-slashes",
    "post-build-lint": "bash scripts/post-build-lint.sh"
  },
```

package.json
bdemers: Cool, is that NPM magic?

package.json
bdemers: +1

package.json
```diff
@@ -10,11 +10,14 @@
...
   "scripts": {
     "clean": "node scripts/clean",
-    "build-prod": "npm run clean && git fetch --force origin master:master && git clone -b master . dist && cross-env JEKYLL_ENV=production bundle exec jekyll build --verbose",
-    "start": "npm run clean && cross-env JEKYLL_ENV=production bundle exec jekyll serve --incremental",
+    "pregenerate": "npm run clean",
+    "generate": "cross-env JEKYLL_ENV=production bundle exec jekyll",
+    "build-prod": "npm run generate build -- --verbose",
+    "postbuild-prod": "bash scripts/post-build-prod.sh",
+    "start": "npm run generate serve -- --incremental",
     "stop": "if-env CI=true && exit 0 || (pkill -f clean && kill $(lsof -t -i:4000; lsof -t -i:4444))",
```
jmelberg-okta: Interestingly, this isn't killing both threads when we run `npm test`. Since we are already here, mind updating this line to:
```
"stop": "if-env CI=true && exit 0 || (kill $(lsof -i:4000; $lsof -i:4444) 2> /dev/null && pkill -f clean)"
```

package.json
```diff
@@ -7,21 +7,20 @@
...
     "prestart": "npm run clean",
     "start": "npm run generate serve -- --incremental",
-
+    "stop": "if-env CI=true && exit 0 || (pkill -f okta_github_io)",
     "test-only": "protractor tests/selenium/conf.js",
     "pretest": "npm run generate serve -- --no-watch & wait-on http://localhost:4000 && webdriver-manager update --gecko false",
     "test": "npm run test-only || (npm stop && exit 1)",
-    "posttest": "npm run stop"
+    "posttest": "npm run stop",
+    "import-external": "node scripts/import-external 2>&1",
```
jmelberg-okta: Where is this script?

package.json
```diff
@@ -7,21 +7,20 @@
...
-    "build-prod": "npm run clean && npm run sync-assets && npm run generate build --verbose",
-
+    "pregenerate": "npm run clean",
+    "generate": "cross-env PID_FINDER=okta_github_io JEKYLL_ENV=production bundle exec jekyll",
+    "build-prod": "npm run generate build -- --verbose",
+    "postbuild-prod": "bash scripts/post-build-prod.sh",
     "prestart": "npm run clean",
     "start": "npm run generate serve -- --incremental",
-
+    "stop": "if-env CI=true && exit 0 || (pkill -f okta_github_io)",
```
jmelberg-okta: I like this, but we need to make sure we kill the chrome driver too.

scripts/gh-pages-publish.sh
```diff
@@ -0,0 +1,31 @@
...
+    exit 1
+fi
+
+# commit
+git add .
+git commit -m "Deploy site for ${CURRENT_HASH}"
+
+# add the remote
+git remote add ${REMOTE_NAME} ${REMOTE_URL}
+#git push ${REMOTE_NAME} ${TARGET_BRANCH}
```
jmelberg-okta: Will this get enabled when we want to merge this in?

scripts/post-build-prod.sh
```diff
@@ -0,0 +1,35 @@
...
+# We are overlaying the new version of the site on top of the old (not deleting)
+
+# clone repo
+git clone -b ${TARGET_BRANCH} ${REMOTE_URL} "${CLONE_DIR}"
+cd ${CLONE_DIR}
+
+# copy the original version of the images in
+IMAGE_DIR="${GENERATED_SITE_LOCATION}/assets/img/"
+mkdir -p "${IMAGE_DIR}"
+rsync -r "${SOURCE_REPO_DIR}/_source/_assets/img/" "${IMAGE_DIR}"
```
jmelberg-okta: This is great for ensuring we copy over all of our assets. But if we want to remove assets, they will be their indefinitely -- correct?

package.json
bdemers: merge error, I'll remove

scripts/post-build-prod.sh
bdemers: We would just need to remove it directly from master, this could be done with a PR after this is merged

scripts/gh-pages-publish.sh
bdemers: doh!

scripts/post-build-prod.sh
jmelberg-okta: Wouldn't it make more sense to perform this action once to get our `source` and `master` branches in sync? Once we are sure all assets in `master` are also in `source`, any updates to the `source` branch directly effect `master`. Isn't this the intended behavior?

jmelberg-okta: Seems like that contradicts cache-busting. Ideally the assets that go into `master` are the assets declared in `source`. With this new approach, you require two PRs to remove an asset (one in `source` and one in `master`). We should **never** be submitting a PR directly to `master` though.

bdemers: We shouldn't need to remove files though, that would only be in the case of a security issue, everything else should be additions or meta-redirects

PR #1393, opened 2017-10-18
Fix sign up button hover state on docs pages

codyrobert:
## Description:
The sign up button hover style is being overridden by the primary nav item hover style. This occurs on docs page when the nav background switches to white after scrolling down the page.

This is fixed by making the primary nav item styles more specific.

### Resolves:
Request by Dana Palmie


codyrobert:
omsqxvljir

bdemers:
@codyrobert  can you target this to the `source` branch.

PR #1398, opened 2017-10-19
Contradiction about API rate limits

chrisbarry-okta:
In two different places, the document says for other endpoints, the limit is 10,000 and 1000. Please either change the values or provide more clarification on the difference.

## Description:

### Resolves:
* [Issue #X](#X)

### DIFF
--- a/_source/_docs/api/getting_started/design_principles.md
+++ b/_source/_docs/api/getting_started/design_principles.md
@@ -305,6 +305,7 @@ endpoint.
 ### Org-Wide Rate Limits
 
 API rate limits apply to the endpoints in an org. The rate applies either to all the endpoints with the same base URL or to an exact URL, as noted in the following table.
+<!-- It says here that for other endpoints, the limit is 10,000. This contradicts the value in the table below. -->
 For all endpoints not listed, the API rate limit is a combined 10,000 requests per minute.
 
 <table border="1" style="width: 100%;">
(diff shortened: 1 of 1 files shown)


mystiberry-okta:
@chrisbarry-okta Working on a related bug, will make sure we resolve this. May take a day or two.

mystiberry-okta:
Moved the comment to after the table for clarity--if it's not in the table, then the default 10k limit applies: https://github.com/okta/okta.github.io/blob/cfb8dc14c821d1ba5f355118c525109ecf62781a/_source/_docs/api/getting_started/design_principles.md#org-wide-rate-limits

@chrisbarry-okta if this is not clear, please ping me.

_source/_docs/api/getting_started/design_principles.md
```diff
@@ -349,6 +350,7 @@ For all endpoints not listed, the API rate limit is a combined 10,000 requests p
 			<td colspan="1" rowspan="1" style="text-align: right;">600</td>
 		</tr>
 		<tr>
+		<!-- Here the number is 1000. If these are two different limits, can we be more explicit about the difference? -->
```
mystiberry-okta: right, for /api/v1/X it is one thousand. I moved the note to below the table to make more sense. We have endpoints that don't begin with /api/v1, like /oauth2. do I need to add an example or is moving the note enough?

chrisbarry-okta: Ahh, okay, now I understand! That helps a lot. Idea: maybe state it like this:
...
If not listed above, but under /api/v1/: 1000
...
If not above (everything else under /): 10,000

mystiberry-okta:
Resolved in https://github.com/okta/okta.github.io/pull/1414 . Thanks for great suggestion!

PR #1389, opened 2017-10-18
Fix screenshot and other minor bugs

nbarbettini:
## Description:
- Correct screenshot (I messed this up earlier)
- Stick with widget@2.1.0 for now, but don't call it latest
- More direction as to where to place code snippets
- Add missing navbar link instructions

cc @leebrandt

PR #1401, opened 2017-10-20
React widget Blog Post fixes

leebrandt:
## Description:
- Changed widget install to 2.3.0
- Removed `npm install` from beginning of article
- Added instructions for fixin NPM dependency issues in widget
- Copied over @nbarbettini changes from PR #1389


_source/_posts/2017-03-30-react-okta-sign-in-widget.md
```diff
@@ -14,39 +14,37 @@ Start by cloning the simple React seed project.
 ```bash
 git clone https://github.com/leebrandt/simple-react-seed.git okta-react-widget-sample
 cd okta-react-widget-sample
-npm install
-npm start
```
mraible: Why did you remove these instructions to install and start?

_source/_posts/2017-03-30-react-okta-sign-in-widget.md
```diff
@@ -252,20 +261,43 @@ export default class LoginPage extends React.Component{
...
+
+Add a Login link to the top navbar by editing `./src/components/common/Navigation.js` and adding a new link under the existing Contact link:
+
+```html
+<li><Link to="login">Login</Link></li>
+```
+
 ## Check It Out
+
+Now install the NPM packages:
```
mraible: Minor nit - I've always seen npm spelled in lowercase, never uppercase.

_source/_posts/2017-03-30-react-okta-sign-in-widget.md
leebrandt: Widget dependency tree has `q` 1.4.1, but if you `npm install` early there is a dependency on `q` 1.5.1 that gets installed, so when you add the widget it doesn't get downgraded and the widget just spins. work around for widget dependency issues.

jmelberg-okta: @leebrandt You can get around this by importing the widget via `/dist/js/okta-signin-widget.min.js` instead of the module.

Curious what you think is the better developer experience?

leebrandt: It should work either way. I don't know that one is better than the other, but not sure why a minor version release caused the widget to not work. Sounds like either we're doing something strange with promises in q (don't see what we COULD be doing), or the q library is sending breaking changes in minor versions (more likely).

leebrandt: I think from a React standpoint, React developers are used to using the modules. Webpack actually follows the module imports in order to determine dependencies and try to make the final minified file as small as possible (AFAIK). Not sure how/if that will be affected by CDN or script import.

jmelberg-okta: For context: This is more of an npm problem since `1.11.` (I believe). `okta-auth-js` uses a specific version of `q`, which could result in multiple versions of `q` installed in `node_modules`. This issue only seems to arise when it is imported in as a module.

Our team is looking at a fix for this soon -- but this in the meantime the workaround is to import in the minified `js` instead.

PR #1415, opened 2017-10-24
Second draft of Grant API

mystiberry-okta:
First draft of Grant API doc

## Description:
- Grant API

### Partially Resolves:
 [OKTA-134810](https://oktainc.atlassian.net/browse/OKTA-134810)

PR #1426, opened 2017-10-26
second draft consent beta

mystiberry-okta:
## Description:
- All beta doc except grants api for consent dialog

### Partially Resolves:
 [OKTA-134810](https://oktainc.atlassian.net/browse/OKTA-134810)


mystiberry-okta:
@johngronberg-okta , @cwu-okta , @larsjohansen-okta , and optionally @federations-okta , everything excluding the grants api should be in this PR. If there's something I should be doc'ing that's not here or in https://github.com/okta/okta.github.io/pull/1415, let me know!

I'm a little worried that the new content might be hard to sort out since it's just tagged with "beta" but no reference to the feature, so let me know if you want me to add something like "Consent Feature" after the beta tag. 

Also not sure I captured this bullet point from the gist:

Authorize endpoint changes
...
- do not fall the request if consents of any of the scopes are granted(implicit or previous granted)

_source/_docs/api/resources/apps.md
```diff
@@ -987,22 +987,25 @@ Adds an OAuth 2.0 client application. This application is only available to the
...
+| grant_types                             | Array of OAuth 2.0 grant type strings                                            | Array of `authorization_code`, `implicit`, `password`, `refresh_token`, `client_credentials` | FALSE    | FALSE  | TRUE       |
+| initiate_login_uri                      | URL that a third party can use to initiate a login by the client                 | String                                                                                       | TRUE     | FALSE  | TRUE       |
+| application_type                        | The type of client application                                                   | `web`, `native`, `browser`, or `service`                                                     | TRUE     | FALSE  | TRUE       |
+| tos_uri {% api_lifecycle beta %}        | Client's terms of service document                                               | URL                                                                                          | TRUE     | FALSE  | FALSE      |
```
cwu-okta: To be consistent with earlier descriptions of uri (such as client_uri) this should use that similar language. This would change to

"URL string of a web page providing the client's terms of service"

Same change needs to be made to policy_uri

_source/_docs/api/resources/apps.md
```diff
@@ -987,22 +987,25 @@ Adds an OAuth 2.0 client application. This application is only available to the
...
+| policy_uri {% api_lifecycle beta %}     | Client's policy document                                                         | URL                                                                                          | TRUE     | FALSE  | FALSE      |
+| consent_method {% api_lifecycle beta %} | Valid values: `REQUIRED` , `TRUSTED`. Read-only: configure in the App Wizard UI. | String                                                                                       | TRUE     | FALSE  | TRUE       |
```
cwu-okta: Data type is Enum since valid values are REQUIRED | TRUSTED.

I wish this table had a Required column. The consent_method is not required. If it is omitted we should indicate the default value is TRUSTED (for this API)

Same that policy_uri and tos_uri are also not required.

_source/_docs/api/resources/apps.md
```diff
@@ -1020,6 +1023,17 @@ Adds an OAuth 2.0 client application. This application is only available to the
     value that includes `authorization_code` implies a `response_types` value that includes `code`, as both values are defined as part of
     the OAuth 2.0 authorization code grant.
 
+* {% api_lifecycle beta %} A consent dialog is displayed depending on the values of `prompt`, `consent_method`, and `consent`:
```
cwu-okta: ``` A consent dialog is displayed ``` -> ``` A consent dialog is displayed to the end user```

_source/_docs/api/resources/apps.md
```diff
@@ -1020,6 +1023,17 @@ Adds an OAuth 2.0 client application. This application is only available to the
...
+
+    | `prompt` Value    | `consent_method` (Apps Property) | `consent` (Scopes Property) | Result       |
```
cwu-okta: I suggest indicating (out of this table) where the prompt, consent_method, and consent come from. We should say prompt is a query param on the /authorize endpoint (with link to that documentation). Same is true for describing consent_method and consent.

_source/_docs/api/resources/apps.md
```diff
@@ -1036,8 +1050,9 @@ curl -v -X POST \
     "signOnMode": "OPENID_CONNECT",
     "credentials": {
       "oauthClient": {
-        "autoKeyRotation": true
-        "token_endpoint_auth_method": "client_secret_post",
+        "client_id":"example.client", //Is this correct? Why not an ID of the form 0oa1hm4POxgJM6CPu0g4 ? See response.
```
cwu-okta: you are the writer. that's your call how you want to show the id. we support both kinds :)

_source/_docs/api/resources/oauth-clients.md
```diff
@@ -53,30 +53,34 @@ flows defined by [the OAuth 2.0 spec](http://oauth.net/documentation) or [OpenID
     "authorization_code"
   ],
   "token_endpoint_auth_method": "client_secret_post",
-  "initiate_login_uri": "https://www.example-application.com/oauth2/login"
+  "initiate_login_uri": "https://www.example-application.com/oauth2/login",
+  "tos_uri":"https://example.com/client/tos",
```
cwu-okta: Would be great if we had the same uri within the doc so we have ***consistency*** :) Please change to example->example-application

_source/_docs/api/resources/oauth-clients.md
```diff
@@ -53,30 +53,34 @@ flows defined by [the OAuth 2.0 spec](http://oauth.net/documentation) or [OpenID
...
+| grant_types                         | Array of OAuth 2.0 grant type strings. Default value: `authorization_code`                                                 | Array of `authorization_code`, `implicit`, `password`, `refresh_token`, `client_credentials` | TRUE     | FALSE  | FALSE     |
+| token_endpoint_auth_method          | requested authentication method for the token endpoint. Default value: `client_secret_basic`                               | `none`, `client_secret_post`, `client_secret_basic`, or `client_secret_jwt`                  | TRUE     | FALSE  | FALSE     |
+| initiate_login_uri                  | URL that a third party can use to initiate a login by the client                                                           | String                                                                                       | TRUE     | FALSE  | FALSE     |
+| tos_uri {% api_lifecycle beta %}    | Client's terms of service document                                                                                         | URL                                                                                          | TRUE     | FALSE  | FALSE     |
```
cwu-okta: Same comment as above that this is not required. We should have a required column.

_source/_docs/api/resources/oauth2.md
```diff
@@ -90,6 +90,18 @@ This is a starting point for OAuth 2.0 flows such as implicit and authorization
...
     Then the client passes the *code_challenge* and *code_challenge_method* in the authorization request for code flow. When a client tries to redeem the code, it must pass the *code_verifier*. Okta recomputes the challenge and returns the requested token only if it matches the *code_challenge* in the original authorization request. When a client, whose *token_endpoint_auth_method* is ``none``, makes a code flow authorization request, *code_challenge* is required.
     Since *code_challenge_method* only supports S256, this means that the value for *code_challenge* must be: `BASE64URL-ENCODE(SHA256(ASCII(*code_verifier*)))`. According to the [PKCE spec](https://tools.ietf.org/html/rfc7636), the *code_verifier* must be at least 43 characters and no more than 128 characters.
+ 
+{% api_lifecycle beta %} The value of `prompt` interacts with other properties to determine if a consent dialog is displayed:
```
cwu-okta: See above comment about this table.

_source/_docs/api/resources/oauth2.md
```diff
@@ -215,6 +227,8 @@ The following parameters can be posted as a part of the URL-encoded form values
...


[JWT Details](#token-authentication-methods)                                                                                                                                                                                     | String |
 | client_assertion_type | Indicates a JWT is being used to authenticate the client. Per the 



[Client Authentication spec](http://openid.net/specs/openid-connect-core-1_0.html#ClientAuthentication), the valid value is `urn:ietf:params:oauth:client-assertion-type:jwt-bearer`.                                                                                  | String |
 
+{% api_lifecycle beta %} Note: if the grant type is `password`, a consent dialog isn't shown regardless of other settings. If the grant type is `client_credentials`, a `consent_required` error response is returned if scopes property `consent` is `REQUIRED`, unless the client is trusted.
```
cwu-okta: What is a trusted client? :) Should elaborate on this (ie must be consent_method=TRUSTED)

_source/_docs/api/resources/oauth2.md
```diff
@@ -2232,20 +2246,36 @@ Token limits:
...
+| id                                   | ID of the scope                                                                                         | String  | FALSE                         |
+| name                                 | Name of the scope                                                                                       | String  | TRUE                          |
+| description                          | Description of the scope                                                                                | String  | FALSE                         |
+| system                               | Whether Okta created the scope                                                                          | Boolean | FALSE                         |
+| default                              | Whether the scope is a default scope                                                                    | Boolean | FALSE                         |
+| displayName {% api_lifecycle beta %} | Name of the end user displayed in a consent dialog <!-- Read only? where does this value come from? --> | String  | FALSE                         |
```
cwu-okta: This PR does not look complete if there are questions in it.

This value is set by the API or UI. Should make sure this property is also in the Scopes api.

_source/_docs/api/resources/apps.md
johngronberg-okta: consent method is writable.

_source/_docs/api/resources/apps.md
johngronberg-okta: you can set a custom client_id. for this example it would probably be best to have app_id and client_id equal

_source/_docs/api/resources/oauth-clients.md
```diff
@@ -53,30 +53,34 @@ flows defined by [the OAuth 2.0 spec](http://oauth.net/documentation) or [OpenID
...
-  "initiate_login_uri": "https://www.example-application.com/oauth2/login"
+  "initiate_login_uri": "https://www.example-application.com/oauth2/login",
+  "tos_uri":"https://example.com/client/tos",
+  "policy_uri":"https://example.com/client/policy"
 }
 ~~~
 
 ### Client Application Properties
 
 Client applications have the following properties:
```
johngronberg-okta: we should note somewhere in here (probably under the create client section), that any client created through this API will have consent_method=required, and will thus pop a consent dialog whenever scopes that require consent are requested.

_source/_docs/api/resources/oauth2.md
johngronberg-okta: This reads a bit confusingly. 

"Note: The `password` grant type should only be used with scopes the require consent if the `consent_method` for the client is `TRUSTED`. Requests for scopes that require consent using the `password` grant type will receive a `consent_required` error response if consent_method=REQUIRED."

_source/_docs/api/resources/oauth2.md
```diff
@@ -2232,20 +2246,36 @@ Token limits:
...
+| consent {% api_lifecycle beta %}     | Read-only value set in scope UI. Valid values: `REQUIRED`, `IMPLICIT`.                                  | Enum    | FALSE                         |
```
johngronberg-okta: This value is writable

_source/_docs/api/resources/apps.md
larsjohansen-okta: While this is technically true, it's a little awkward for an admin to read; Could we word it as "A consent dialog is displayed to the end user depending on the values of the prompt and the apps and scope consent properties."? It's slightly less "correct" but I would assume it's clear enough from the table below what this means.

_source/_docs/api/resources/apps.md
```diff
@@ -1020,6 +1023,17 @@ Adds an OAuth 2.0 client application. This application is only available to the
     value that includes `authorization_code` implies a `response_types` value that includes `code`, as both values are defined as part of
     the OAuth 2.0 authorization code grant.
 
+* {% api_lifecycle beta %} A consent dialog is displayed depending on the values of `prompt`, `consent_method`, and `consent`:
+
+    | `prompt` Value    | `consent_method` (Apps Property) | `consent` (Scopes Property) | Result       |
+    |:------------------|:---------------------------------|:----------------------------|:-------------|
+    | `CONSENT`         | `TRUSTED` or `REQUIRED`          | `REQUIRED`                  | Prompted     |
+    | `CONSENT`         | `TRUSTED`                        | `IMPLICIT`                  | Not prompted |
+    | `NONE` or `LOGIN` | `TRUSTED`                        | `REQUIRED` or `IMPLICIT`    | Not prompted |
```
larsjohansen-okta: Since prompt=login is officially not supported for consent beta, can we take that out of the table?

_source/_docs/api/resources/oauth2.md
```diff
@@ -90,6 +90,18 @@ This is a starting point for OAuth 2.0 flows such as implicit and authorization
...
     Since *code_challenge_method* only supports S256, this means that the value for *code_challenge* must be: `BASE64URL-ENCODE(SHA256(ASCII(*code_verifier*)))`. According to the [PKCE spec](https://tools.ietf.org/html/rfc7636), the *code_verifier* must be at least 43 characters and no more than 128 characters.
+ 
+{% api_lifecycle beta %} The value of `prompt` interacts with other properties to determine if a consent dialog is displayed:
+     
+| `prompt` Value    | `consent_method` (Apps Property) | `consent` (Scopes Property) | Result       |
+|:------------------|:---------------------------------|:----------------------------|:-------------|
+| `CONSENT`         | `TRUSTED` or `REQUIRED`          | `REQUIRED`                  | Prompted     |
+| `CONSENT`         | `TRUSTED`                        | `IMPLICIT`                  | Not prompted |
+| `NONE` or `LOGIN` | `TRUSTED`                        | `REQUIRED` or `IMPLICIT`    | Not prompted |
```
larsjohansen-okta: Same here, can we remove login?

_source/_docs/api/resources/oauth2.md
```diff
@@ -2232,20 +2246,36 @@ Token limits:
     "name": "car:drive",
     "description": "Drive car",
     "system": false,
-    "default": false
+    "default": false,
+    "displayName": "Saml Jackson",
+    "consent": "REQUIRED"
```
larsjohansen-okta: Should we not mark this and the display name as only relevant for consent beta anywhere?

_source/_docs/api/resources/oauth2.md
larsjohansen-okta: Name of the end user displayed in a consent dialog -> Human readable name of the scope displayed in a consent dialog

_source/_docs/api/resources/oauth2.md
```diff
@@ -2232,20 +2246,36 @@ Token limits:
...
+| displayName {% api_lifecycle beta %} | Name of the end user displayed in a consent dialog <!-- Read only? where does this value come from? --> | String  | FALSE                         |
+| consent {% api_lifecycle beta %}     | Read-only value set in scope UI. Valid values: `REQUIRED`, `IMPLICIT`.                                  | Enum    | FALSE                         |
+
+{% api_lifecycle beta %} A consent dialog is displayed when the end user signs in depending on the values of `prompt`, `consent_method`, and `consent`:
+
+| `prompt` Value    | `consent_method` (Apps Property) | `consent` (Scopes Property) | Result       |
+|:------------------|:---------------------------------|:----------------------------|:-------------|
+| `CONSENT`         | `TRUSTED` or `REQUIRED`          | `REQUIRED`                  | Prompted     |
+| `CONSENT`         | `TRUSTED`                        | `IMPLICIT`                  | Not prompted |
+| `NONE` or `LOGIN` | `TRUSTED`                        | `REQUIRED` or `IMPLICIT`    | Not prompted |
```
larsjohansen-okta: remove login

_source/_docs/api/resources/apps.md
mystiberry-okta: @cwu-okta is Nullable different from Required? All three are marked Nullable as true.  I'll add default column for consent_method, great catch!

_source/_docs/api/resources/apps.md
mystiberry-okta: Done. (admins read this doc?!)

_source/_docs/api/resources/apps.md
mystiberry-okta: Done (combined with Lars request above).

_source/_docs/api/resources/apps.md
mystiberry-okta: Fixed, swapped example.client for an ID.

_source/_docs/api/resources/oauth-clients.md
mystiberry-okta: Done.

_source/_docs/api/resources/oauth-clients.md
mystiberry-okta: Is nullable = true not the same as required = false, @cwu-okta ?

_source/_docs/api/resources/oauth2.md
mystiberry-okta: Done.

_source/_docs/api/resources/oauth2.md
mystiberry-okta: Done, and added link, too.

_source/_docs/api/resources/oauth2.md
mystiberry-okta: @larsjohansen-okta, is the displayName derived from some other property? Is it set in the AS or scope dialog? Is it read only? Thanks!

_source/_docs/api/resources/apps.md
mystiberry-okta: Fixed.

_source/_docs/api/resources/apps.md
mystiberry-okta: Fixed.

_source/_docs/api/resources/oauth-clients.md
mystiberry-okta: Done.

mystiberry-okta: We missed moving the data model to the end of the topic, so I did that as well.

_source/_docs/api/resources/oauth2.md
mystiberry-okta: Done. @cwu-okta this rewrite is a little different from yours, so I kept the link you requested but used this wording.

_source/_docs/api/resources/oauth2.md
mystiberry-okta: Replaced the read only phrase with this description: Indicates whether a consent dialog is needed for the scope.

_source/_docs/api/resources/oauth2.md
mystiberry-okta: Done.

_source/_docs/api/resources/oauth2.md
mystiberry-okta: Done.

_source/_docs/api/resources/apps.md
```diff
@@ -1020,6 +1023,20 @@ Adds an OAuth 2.0 client application. This application is only available to the
     value that includes `authorization_code` implies a `response_types` value that includes `code`, as both values are defined as part of
     the OAuth 2.0 authorization code grant.
 
+* {% api_lifecycle beta %} A consent dialog is displayed depending on the values of three elements:
+    * `prompt`, a query parameter used in requests to [`/oauth2/:authorizationServerId/v1/authorize`](docs/api/resources/oauth2.html) or [`/oauth2/v1/authorize`](/docs/api/resources/oidc.html)
+    * `consent_method`, a property on apps that is set in the Okta user interface.
+    * `consent`, a property on scopes that is set in the Okta user interface.
+
+    | `prompt` Value    | `consent_method` (Apps Property) | `consent` (Scopes Property) | Result       |
```
cwu-okta: Since you have the above indicating where one finds out what consent_method and consent are, you don't need the parens here with ```(... Property)```

_source/_docs/api/resources/oauth2.md
```diff
@@ -2232,20 +2249,36 @@ Token limits:
...
+| name                                 | Name of the scope                                                                                       | String  | TRUE                          |
+| description                          | Description of the scope                                                                                | String  | FALSE                         |
+| system                               | Whether Okta created the scope                                                                          | Boolean | FALSE                         |
+| default                              | Whether the scope is a default scope                                                                    | Boolean | FALSE                         |
+| displayName {% api_lifecycle beta %} | Name of the end user displayed in a consent dialog                                                      | String  | FALSE                         |
+| consent {% api_lifecycle beta %}     | Indicates whether a consent dialog is needed for the scope. Valid values: `REQUIRED`, `IMPLICIT`.       | Enum    | FALSE                         |
```
cwu-okta: This enum is listed as not required. As a reader my next logical question is "What default value is used if I omit this from my create API request?"

_source/_docs/api/resources/apps.md
```diff
@@ -987,22 +987,25 @@ Adds an OAuth 2.0 client application. This application is only available to the
...
+| application_type                        | The type of client application                                                   | `web`, `native`, `browser`, or `service`                                                     | TRUE     | FALSE  | TRUE       |           |
+| tos_uri {% api_lifecycle beta %}        | URL string of a web page providing the client's terms of service document        | URL                                                                                          | TRUE     | FALSE  | FALSE      |           |
+| policy_uri {% api_lifecycle beta %}     | URL string of a web page providing the client's policy document                  | URL                                                                                          | TRUE     | FALSE  | FALSE      |           |
+| consent_method {% api_lifecycle beta %} | Valid values: `REQUIRED` , `TRUSTED`.                                            | String                                                                                       | TRUE     | FALSE  | TRUE       | `TRUSTED` |
```
cwu-okta: I think we should explicitly call out (in the bullet points below) that the consent_method here is different than the /clients API, and put the analogous callout in the /clients API.
@gronberg Do you have an opinion?

johngronberg-okta: Yes, we should note that the behavior is different. On /clients we should not include consent_method as a request param, but just note it as a callout at the bottom of the request table. 

Something like:
"Apps created on /apps will default to consent_method=TRUSTED, while those created on /clients will default to consent_method=REQUIRED"

_source/_docs/api/resources/oauth2.md
mystiberry-okta: @johngronberg-okta  or @larsjohansen-okta  if you omit `consent`, no consent dialog is displayed--but if OIDC or custom scopes require consent, you get an error, yes?

mystiberry-okta: @johngronberg-okta Please confirm or correct my guess.

_source/_docs/api/resources/apps.md
mystiberry-okta: @johngronberg-okta We don't have consent_method as a property on any object. That seems wrong. Removing from request params as mentioned, but shouldn't it be in the list of properties for the Application Object or Application User Object ?

_source/_docs/api/resources/apps.md
mystiberry-okta: Done.

_source/_docs/api/resources/apps.md
```diff
@@ -987,22 +987,25 @@ Adds an OAuth 2.0 client application. This application is only available to the
...
+| grant_types                             | Array of OAuth 2.0 grant type strings                                            | Array of `authorization_code`, `implicit`, `password`, `refresh_token`, `client_credentials` | FALSE    | FALSE  | TRUE       |           |
+| initiate_login_uri                      | URL that a third party can use to initiate a login by the client                 | String                                                                                       | TRUE     | FALSE  | TRUE       |           |
+| application_type                        | The type of client application                                                   | `web`, `native`, `browser`, or `service`                                                     | TRUE     | FALSE  | TRUE       |           |
+| tos_uri {% api_lifecycle beta %}        | URL string of a web page providing the client's terms of service document        | URL                                                                                          | TRUE     | FALSE  | FALSE      |           |
```
cwu-okta: Nit: sorry I missed this on the last review but tos_uir and policy_uri both now say "URL string" but above there is initiate_login_uri which is "URL that...". Can we make beginning descriptions all match?

_source/_docs/api/resources/apps.md
cwu-okta: @mystiberry-okta  consent_required is only exposed in the /apps api. It is not exposed in the /clients object. The idea is to call out that a client created via /apps API vs one created via /clients will produce different clients with respect to consent. A client created via the /apps API is considered a trusted client. A client created via the /clients API is untrusted. Even though the /clients API does not expose the consent_required attribute, one can tell if the client is trusted/untrusted by navigating to the UI and it can be seen there. The second way one can see if the client is trusted/untrusted is by issuing a GET of the /apps api to see what value we set.

We do not expose consent_method in /clients because it our spec-compliant API. The spec doesn't have a consent_required attribute (or analogous one) for the clients API so that is why we do not expose it either.

_source/_docs/api/resources/apps.md
```diff
@@ -1020,6 +1023,20 @@ Adds an OAuth 2.0 client application. This application is only available to the
     value that includes `authorization_code` implies a `response_types` value that includes `code`, as both values are defined as part of
     the OAuth 2.0 authorization code grant.
 
+* {% api_lifecycle beta %} A consent dialog is displayed depending on the values of three elements:
+    * `prompt`, a query parameter used in requests to [`/oauth2/:authorizationServerId/v1/authorize`](docs/api/resources/oauth2.html) or [`/oauth2/v1/authorize`](/docs/api/resources/oidc.html)
```
cwu-okta: Nit: would you consider specifying one /authorize uri is for API AM and the other for OIDC?

_source/_docs/api/resources/apps.md
```diff
@@ -1020,6 +1023,20 @@ Adds an OAuth 2.0 client application. This application is only available to the
...
+    * `consent_method`, a property on apps that is set in the Okta user interface.
```
cwu-okta: Link to the /apps api body which defines this property. Same for consent.

_source/_docs/api/resources/apps.md
cwu-okta: @mystiberry-okta  this doesn't look like it was done (or not pushed up)

_source/_docs/api/resources/oauth2.md
cwu-okta: @mystiberry-okta  This has to deal with how scopes are created, not how the consent dialog is displayed per se. We always remain backwards compat. So default is consent=implicit.

_source/_docs/api/resources/apps.md
mystiberry-okta: I was only asking about /apps.

_source/_docs/api/resources/apps.md
mystiberry-okta: Yup, not pushed up. You're re-reviewing the same content.

_source/_docs/api/resources/oauth2.md
mystiberry-okta: Thanks!

_source/_docs/api/resources/apps.md
mystiberry-okta: done.

_source/_docs/api/resources/oauth2.md
mystiberry-okta: Added default column with IMPLICIT.

_source/_docs/api/resources/apps.md
mystiberry-okta: Why are tos_uri and policy_uri of type URL but initiate_login_uri of type string?

mystiberry-okta: Also, why call those first two a URL string if the datatype is URL?

_source/_docs/api/resources/apps.md
mystiberry-okta: is it that, or is it API AM vs. not API AM (i.e. custom AS vs. org AS)?

_source/_docs/api/resources/apps.md
mystiberry-okta: link on consent_method and consent added.

mystiberry-okta: I don't think link on consent_method is a good idea, it's in the table that these tables are "children" of, and there's no way to link to a line in the table. So for it I put (listed in the Settings table above), and provided link to consent.

mystiberry-okta:
@cwu-okta I wasn't able to put "done" after a few of your comments. This is ready for 4th draft review.

_source/_docs/api/resources/apps.md
cwu-okta: They should all be datatype url.

_source/_docs/api/resources/apps.md
cwu-okta: It is based on the protocol, so it is OIDC and API AM.

_source/_docs/api/resources/apps.md
```diff
@@ -987,22 +987,25 @@ Adds an OAuth 2.0 client application. This application is only available to the
...
+| Parameter                                 | Description                                                                                 | DataType                                                                                     | Nullable | Unique | Validation | Default   |
+|:------------------------------------------|:--------------------------------------------------------------------------------------------|:---------------------------------------------------------------------------------------------|:---------|:-------|:-----------|:----------|
+| client_uri                                | URL string of a web page providing information about the client                             | String                                                                                       | TRUE     | FALSE  | FALSE      |           |
+| logo_uri                                  | URL string that references a logo for the client                                            | String                                                                                       | TRUE     | FALSE  | FALSE      |           |
```
cwu-okta: Since we are scrutinizing the data types now, the logo_uri is also a data type of URL

_source/_docs/api/resources/apps.md
```diff
@@ -1020,6 +1023,21 @@ Adds an OAuth 2.0 client application. This application is only available to the
     value that includes `authorization_code` implies a `response_types` value that includes `code`, as both values are defined as part of
     the OAuth 2.0 authorization code grant.
 
+* {% api_lifecycle beta %} A consent dialog is displayed depending on the values of three elements:
+    * `prompt`, a query parameter used in requests to [`/oauth2/:authorizationServerId/v1/authorize`](/docs/api/resources/oauth2.html#obtain-an-authorization-grant-from-a-user)(custom authorization server) or [`/oauth2/v1/authorize`](/docs/api/resources/oidc.html#authentication-request) (Org authorization server)
```
cwu-okta: As mentioned in my other comment, we could use API AM and OIDC terms here instead of custom/org AS.
@gronberg

_source/_docs/api/resources/apps.md
```diff
@@ -1020,6 +1023,21 @@ Adds an OAuth 2.0 client application. This application is only available to the
...
+    * `consent_method`, a property (listed in the Settings table above) that is set in the Okta user interface
```
cwu-okta: ```set in the Okta user interface``` Is not entirely accurate. This is also settable via API. We could split the diff and say settable when you configure the OAuth 2.0 client

_source/_docs/api/resources/apps.md
```diff
@@ -1020,6 +1023,21 @@ Adds an OAuth 2.0 client application. This application is only available to the
...
+    * `consent`, a property on [scopes](/docs/api/resources/oauth2.html#scopes-properties) that is set in the Okta user interface
```
cwu-okta: Again, this property is settable via UI also. We could follow the same strategy and indicate it is set when you configure the scope.

_source/_docs/api/resources/oauth2.md
```diff
@@ -90,6 +90,20 @@ This is a starting point for OAuth 2.0 flows such as implicit and authorization
...
     Then the client passes the *code_challenge* and *code_challenge_method* in the authorization request for code flow. When a client tries to redeem the code, it must pass the *code_verifier*. Okta recomputes the challenge and returns the requested token only if it matches the *code_challenge* in the original authorization request. When a client, whose *token_endpoint_auth_method* is ``none``, makes a code flow authorization request, *code_challenge* is required.
     Since *code_challenge_method* only supports S256, this means that the value for *code_challenge* must be: `BASE64URL-ENCODE(SHA256(ASCII(*code_verifier*)))`. According to the [PKCE spec](https://tools.ietf.org/html/rfc7636), the *code_verifier* must be at least 43 characters and no more than 128 characters.
+ 
+ * {% api_lifecycle beta %} A consent dialog is displayed depending on the values of three elements:
+     * `prompt`, a query parameter used in requests to [`/oauth2/:authorizationServerId/v1/authorize`](/docs/api/resources/oauth2.html##obtain-an-authorization-grant-from-a-user) or [`/oauth2/v1/authorize`](/docs/api/resources/oidc.html).
```
cwu-okta: We should make sure this table and description matches the same as the other one. Currently the two are different.

_source/_docs/api/resources/apps.md
mystiberry-okta: Since this isn't wrong, I'll leave as is for now and follow up.

_source/_docs/api/resources/apps.md
mystiberry-okta: Done.

_source/_docs/api/resources/apps.md
mystiberry-okta: Ack, copied from /oauth-clients.md, thanks for catching! Done.

mystiberry-okta: Actually, I just deleted, and added "only" to consent (only set in the UI).

mystiberry-okta: So, I'm deleting the phrase entirely, not needed.

_source/_docs/api/resources/oauth2.md
mystiberry-okta: Tables are the same now, except that there's a link to consent_method in the oauth2.md file (since consent is doc'd in apps).

_source/_docs/api/resources/oauth2.md
mystiberry-okta: Also, it's in the UI, therefore it's meant for people.

_source/_docs/api/resources/apps.md
mystiberry-okta: Every time I've done it that way, we've had customer complaints and I've had to revise. So leaving it as is (that's the difference between the two endpoints), and have filed a JIRA to follow up on this issue. It creates a lot of churn.

PR #1438, opened 2017-10-31
Disabling quick-start server selection on headless chrome

vijetmahabaleshwar-okta:
The test still runs on chrome browser using sauce labs

## Description:
- Disabled a failing test on headless chrome


jmelberg-okta:
@vijetmahabaleshwar-okta: Can you provide a little more details as to why these tests are failing here and not locally?

PR #1450, opened 2017-11-03
Fix default framework selection in quickstarts

robertjd:
## Description:

The default framework was not being selected, it would always default to Express if not specified

Adding a new `default` property in the code to make this more explicit, and decouple from active link state

### Resolves:
* [OKTA-145141](https://oktainc.atlassian.net/browse/OKTA-145141)

@jmelberg-okta to review functionality 
@vijetmahabaleshwar-okta to look at test strategy


tests/selenium/framework/page-objects/QuickStartsPage.js
```diff
@@ -30,6 +30,20 @@ class QuickStartsPage extends BasePage {
...
   }
 
+  /**
+   *
+   * In order to test default selections, we need to leave the page and then come back.
+   * Why? Because changing the URL fragment in-page will not cause the quickstart app's
+   * main() function to re-run, which is what we want to test.  We want to test when
+   * the app is bootstrapped with specific URL fragments.
+   */
+  leave() {
```
jmelberg-okta: Since this is a utility method, it would make more sense to put it in `shared/util.js`. This may come in handy in other pages.

robertjd: Perhaps we could create a generic "go to another page" utility, but "leaving" the quickstart has a pretty specific meaning IMO so I'm inclined to leave it here.  Leaving could mean other things in other contexts (probably just the URL you want to go to, but :man_shrugging: )

jmelberg-okta: Thats a fair point. I can't really see another scenario where that would be useful.

PR #1461, opened 2017-11-07
Redirecting /code/java/jwt-validation to github project page

bdemers:
Depends on: https://github.com/okta/okta-jwt-verifier-java/pull/6


bdemers:
It would be in search indexes, so we should keep it.  The goal is to send people to the updated README in the actual project repo. Though we do need to go over the root lang pages and update them with new samples, and links to various projects (in another PR)

PR #1464, opened 2017-11-08
Manual GA tracking for quickstart anchors

robertjd:
## Description:

The autotrack library is not creating pageview events when we change the document hash, so now I'm doing it manually.

Following this advice: https://developers.google.com/analytics/devguides/collection/analyticsjs/single-page-applications#tracking_virtual_pageviews

And debugging with this strategy: https://developers.google.com/analytics/devguides/collection/analyticsjs/debugging


_source/_assets/js/quickstart.js
```diff
@@ -272,6 +272,10 @@
   function applySelectionTuple(selectionTuple) {
     window.location.replace('#/' + selectionTuple.join('/'));
     loadContent.apply(null, selectionTuple);
+    if (window.ga) {
+      window.ga('set', 'page', window.location.pathname + window.location.hash);
```
jmelberg-okta: Did we try setting these values on the current <https://developer.okta.com> site to see if they were being recorded?

PR #1475, opened 2017-11-10
Fixes incorrect params when updating the swa3 template

bdemers:


mystiberry-okta:
@bdemers I have a bug for this that says url is also wrong (should be targetURL). I'll fix it, just want to double-check that my info is correct: https://oktainc.atlassian.net/browse/OKTA-147418 Thank you for fixing the other three!

bdemers:
@bretterer ^
Looks like I also missed updating the table for the SWA 3, @mystiberry-okta can you hit that when you make the other changes?

PR #1488, opened 2017-11-16
Add error examples to oauth clients

mystiberry-okta:
## Description:
- Add error examples to each endpoint
- Re-ordered two endpoints (list before get)
- Fixed error in limit param (default 20, Max 200)
- Changed nesting of two endpoints because why nest them if there are only two?

Note: There are only so many mistakes you can make with these endpoints. Do we really need one per endpoint?

### Resolves:

[OKTA-134742](https://oktainc.atlassian.net/browse/OKTA-134742)


mystiberry-okta:
Ready for review @federations-okta

_source/_docs/api/resources/oauth-clients.md
```diff
@@ -118,34 +119,57 @@ Content-Type: application/json;charset=UTF-8
...
+{:.api .api-response .api-response-example}
+
+~~~http
+HTTP/1.1 400 Bad Request
+Content-Type: application/json
+~~~
+~~~json
+{
+    "errorCode": "E0000021",
+    "errorSummary": "Bad request.  Accept and/or Content-Type headers likely do not match supported values.",
```
danielzeller-okta: This doesn't look right. The correct error message should be:
```
{
    "error": "invalid_client_metadata",
    "error_description": "client_name: An object with this field already exists in the current organization"
}
```

_source/_docs/api/resources/oauth-clients.md
```diff
@@ -155,61 +179,137 @@ curl -v -X GET \
...
+            "refresh_token",
+            "implicit"
+        ],
+        "token_endpoint_auth_method": "client_secret_basic",
+        "application_type": "web"
+    }
+]
+~~~
+
+##### Response Example: Incorrect Domain Name
```
danielzeller-okta: This kinda seems like a strange place to be documenting this error, as it's api-wide and more generic than just the incorrect domain (i.e. may be that the token is incorrect/expired/user deactivated etc.)

Do we really have people tripping up over the subdomain that often?

_source/_docs/api/resources/oauth-clients.md
mystiberry-okta: Just tested and you are right. Will fix.

_source/_docs/api/resources/oauth-clients.md
mystiberry-okta: Agreed--it's just the only error you could reasonably make. I can leave it out entirely--or if someone can suggest a better error for here, @danielzeller-okta  or anyone?

_source/_docs/api/resources/oauth-clients.md
```diff
@@ -424,8 +531,22 @@ Content-Type: application/json;charset=UTF-8
...
-### Generate new client secret
+~~~http
+HTTP/1.1 400 Bad Request
+Content-Type: application/json;charset=UTF-8
+~~~
+
+~~~json
+{
+    "error": "invalid_client_metadata",
+    "error_description": "'application_type' cannot be modified. client_name: The field cannot be left blank"
```
danielzeller-okta: just missing the client_name should only have the second part of this error description:
`"error_description": "client_name: The field cannot be left blank"`

_source/_docs/api/resources/oauth-clients.md
```diff
@@ -531,6 +670,24 @@ curl -v -X DELETE \
...
+~~~
+
+~~~json
+{
+    "errorCode": "E0000007",
+    "errorSummary": "Not found: Resource not found: 0jrubyXXm4B9zVJPboyT (PublicClientApp)",
+    "errorLink": "E0000007",
+    "errorId": "oae7LrklqzDQCmZSkcOgdImDg",
+    "errorCauses": []
+}
```
danielzeller-okta: I don't think this error is correct, I think it should be 
```
{
    "error": "invalid_client",
    "error_description": "Invalid value for 'client_id' parameter."
}
```

_source/_docs/api/resources/oauth-clients.md
danielzeller-okta: In my opinion it should be left out, but I don't feel strongly about it.

mystiberry-okta: Removed.

_source/_docs/api/resources/oauth-clients.md
mystiberry-okta: Ack, Thanks for catching this!

_source/_docs/api/resources/oauth-clients.md
mystiberry-okta: Yes, the HTTP code is wrong as well. Fixed.

_source/_docs/api/resources/oauth-clients.md
```diff
@@ -155,61 +176,120 @@ curl -v -X GET \
 HTTP/1.1 200 OK
 Content-Type: application/json;charset=UTF-8
 ~~~
+Header links for use in the next request's `after` parameter:
```
danielzeller-okta: the phrasing here sounds a little awkward, specifically calling out the `after` param like this since we only mention that this is for pagination later on.

_source/_docs/api/resources/oauth-clients.md
```diff
@@ -155,61 +176,120 @@ curl -v -X GET \
...
+            "implicit"
+        ],
+        "token_endpoint_auth_method": "client_secret_basic",
+        "application_type": "web"
+    }
+]
 ~~~
 
-### List Client Applications
+### List Client Applications Matching a Search Filter
```
danielzeller-okta: is there somewhere that we call out what kind of search filters we support?

mystiberry-okta: Yup, line 148 says `q` only searches on client_name.

_source/_docs/api/resources/oauth-clients.md
mystiberry-okta: @danielzeller-okta I added "Response body:" before the JSON for the response body, and changed the text here and in the other instance to "Header links for pagination:" Does that make it less awkward? @cwu-okta or @johngronberg-okta , can you weigh in? I'm not sure I'm fixing the problem Daniel's worried about--not sure how to if this doesn't do it.

cwu-okta: Can you remind me why we want to add the complexity of paging to this simple example? If there wasn't a good reason to add it why not make this just a simple example where there is a single page of data, where you don't have a next link.

danielzeller-okta: "Header links for pagination" sounds good to me.

PR #1498, opened 2017-11-20
Add support for Disqus comments

mraible:
## Description:

Add comment support via Disqus. 

Please review @rdegges.

This is what it looks like when you're not logged in.

screen shot 2017-11-20 at 2 47 30 pm

And this is what it looks like when you are logged in.

screen shot 2017-11-20 at 2 47 18 pm


_source/_includes/blog_post.html
```diff
@@ -1,30 +1,54 @@
...
+            {% endif %}
+            <span class="BlogPost-author">{{ author.full_name }}</span>
+        </div>
+
+    </header>
+
+    <div class="BlogPost-content">
+        {{ include.post_content }}
+    </div>
+
```
mraible: All of the changes above seem to be caused by my IDE changing from tabs to space. Happy to change it back, but I'm a fan of spaces. ;0)

mraible:
Latest screenshot, with a top-margin of 40px. 

screen shot 2017-11-21 at 11 14 27 am

Also, I added a feature where you can add `comments: false` in a posts header to turn comments off for that post.

rdegges:
Looks good to me. Who can merge? Way overdue:)

PR #1512, opened 2017-11-28
Fix issues in React guide

lboyette-okta:
## Description:
- These were issues that were causing confusion

### Related-to:
* https://github.com/okta/okta-oidc-js/issues/52
* https://github.com/okta/okta-oidc-js/issues/43

### DIFF
--- a/_source/quickstart/react/default-example.md
+++ b/_source/quickstart/react/default-example.md
@@ -57,12 +57,12 @@ You will need the values from the OIDC client that you created in the previous s
 In your application's `App.js` file, import the following objects and pass in your configuration:
 
 ```typescript
-import { Security, SecureRoute, ImplicitCallback } from '@okta/okta-react';
+import { Security, ImplicitCallback } from '@okta/okta-react';
 
 const config = {
   issuer: 'https://{yourOktaDomain}.com/oauth2/default',
-  redirectUri: window.location.origin + '/implicit/callback',
-  clientId: '{clientId}'
+  redirect_uri: window.location.origin + '/implicit/callback',
+  client_id: '{clientId}'
 }
 
 ```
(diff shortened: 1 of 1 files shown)


_source/quickstart/react/default-example.md
```diff
@@ -168,7 +167,7 @@ export default withAuth(class MessageList extends Component {
 
   async componentDidMount() {
     try {
-      const response = await fetch('http://localhost:{serverPort}/api/messages', {
+      const response = await fetch('http://localhost:{port}/api/messages', {
```
jmelberg-okta: FYI: In the intro we say:
> If you are running this locally and using the defaults from the [React Quickstart](), your port will be `3000`

This might confuse the reader as we changed the variable to be `port` instead of `serverPort`. I'd suggest adding a comment above this line, or changing the variable back to `{serverPort}`

_source/quickstart/react/default-example.md
```diff
@@ -134,7 +133,7 @@ class App extends Component {
     return (
       <Router>
         <Security issuer={config.issuer}
-                  client_id={config.clientId}
+                  client_id={config.client_id}
                   redirect_uri={config.redirect_uri} >
           <Route path='/' exact={true} component={Home}/>
```
jmelberg-okta: nit: Space on closing angle bracket

mraible: FYI: I'm using Create React App and its default `tslint.json` settings require this block to be as follows:

```html
<Router>
  <Security
    issuer={config.issuer}
    client_id={config.clientId}
    redirect_uri={config.redirectUri}
  >
    <Route path="/" exact={true} component={Home}/>
    <Route path="/implicit/callback" component={ImplicitCallback}/>
  </Security>
</Router>
```

jmelberg-okta: @mraible Is this via the `create-react-app` project scaffold?

_source/quickstart/react/default-example.md
jmelberg-okta: We use `{port}` to indicate our front-end, while `{serverPort}` is for the back-end.

_source/quickstart/react/default-example.md
lboyette-okta: @mraible this is really small, so we can make accommodations, but there are a ton of reasons why the typescript create-react-app will not work with this guide

mraible: Yeah, it was just an FYI more than anything. The good news is I got our React SDK to work with TypeScript so I'll be writing up a blog post on that soon.

PR #1519, opened 2017-11-30
Fix incorrect HTTP command and add note

mystiberry-okta:
## Description:
- request logout is a GET not a POST. also call out it's not a server-side call.

### Resolves:
 [OKTA-142771](https://oktainc.atlassian.net/browse/OKTA-142771)


_source/_docs/api/resources/oidc.md
```diff
@@ -761,9 +761,11 @@ Content-Type: application/json;charset=UTF-8
 ### Logout Request
 {:.api .api-operation}
 
-{% api_operation post /oauth2/v1/logout %}
+{% api_operation GET /oauth2/v1/logout %}
 
-The API takes an ID Token and logs the user out of Okta if the subject matches the current Okta session. A `post_logout_redirect_uri` may be specified to redirect the User after the logout has been performed. Otherwise, the user is redirected to the Okta login page.
+The API takes an ID Token and logs the user out of the Okta session if the subject matches the current Okta session. A `post_logout_redirect_uri` may be specified to redirect the User after the logout has been performed. Otherwise, the user is redirected to the Okta login page.
+
+Note that this is not a server-side call.
```
mystiberry-okta: Is this sufficient?

royalchan-okta: This is correct, but maybe it would be good to expand the last sentence and say what it *is* for. Maybe something like:
```
This is provided for RP-initiated logout for the User's browser session.
```

This feature still needs quite a bit of polish (prompting the user to see if they want to log out, etc) but there hasn't been enough demand to prioritize more work, so I'm not keen on expanding too much and advertising this strongly.

mystiberry-okta: Understood, thank you!

PR #1534, opened 2017-12-05
POC: Paginate Release notes - DO NOT MERGE

bdemers:
Build normally, then browse to the release notes, or directly to: http://localhost:4000/collection/release-notes

Still a few problems to take care of before merging, but this works enough to talk about:
- All collections are rendered with the paginate template (need to configure this to ONLY be release_notes)
- The current template prefixes `/collection/:coll` this can be changed but the previous item needs to be resolved first
- I made a first pass at cleaning up the individual release note files, but we will need to review the headings in each file so the they show up correctly in the ToC


_source/_data/docs.yml
```diff
@@ -1,5 +1,5 @@
 # -- Okta API Release Notes
-- section: platform-release-notes/platform-release-notes.html
+- section: collection/release-notes
```
jmelberg-okta: Is this creating a new path? `https://developer.okta.com/collection/release-notes/XYZ`?

_config.yml
```diff
@@ -79,9 +79,28 @@ collections:
...
   quickstarts:
     output: true
-
-paginate: 10
-paginate_path: "/blog/page/:num"
+  release_notes:
+    output: false
+    permalink: /:name
+
+autopages:
```
jmelberg-okta: Curious why we are using a different version of `jekyll-paginate` if we can repeat the logic used for our `blog` with our previous implementation. Can you provide some details why we are changing dependencies?

_source/_layouts/blog_index.html
```diff
@@ -17,7 +17,7 @@
 	<div class="Blog-pagination">
 		
 		{% if paginator.previous_page %}
-		<a href="{{ paginator.previous_page_path | prepend: site.baseurl | append: '/' | replace: '//', '/' }}">&laquo; Prev</a>
+		<a href="{{ paginator.previous_page_path | prepend: site.baseurl | replace: '//', '/' }}">&laquo; Prev</a>
```
jmelberg-okta: I'm not sure why we'd ever have `//` now, while previously we were appending an extra `/`

_source/_data/docs.yml
mystiberry-okta: Yes, but we're changing the folder & file names anyway as they are too long for good SEO.

jmelberg-okta: Gotcha. It looks like we are continuing to nest it under `docs`, so we wont need to setup any other behaviors in Cloudfront. My suggestion to avoid this is to update the URL to:
> https://developer.okta.com/docs/release-notes/:year/:week

Thus:
> https://developer.okta.com/docs/release-notes/2018/1

mystiberry-okta: Yah, we need to include "api" to disambiguate from prod release notes, so /docs/api-release-notes/2018 makes sense to me. Do we need a week directory for one file? (api for api products, not that we only ever have api info in the rn).

bdemers: If paginated we wouldn't have the year in the URL. As that is not important to end users, for example the 3rd week of January would have one item (not helpful to the user as it requires them to click deeper to find the info)

End users likely only care about the last `N` release notes. ( I don't know what that number is, 4, or 10), but that is easily adjusted with a couple key strokes.

mystiberry-okta: Let's assume we're going to publish the whole year to a single file and revise that if needed. We can't guess which RN are of interest (per the GA analysis), but it seems clear that a sizeable (10-50 views for a few dozen if I remember right) views are for surprisingly old RN.

bdemers: I'm a big `-1` for publishing the year files, and am strongly advocating in a paginated solution (newest-to-oldest)

mystiberry-okta: Yup, pending feedback from @jakubvul, I officially give up on my concerns about repeated text in the file (we'll minimize it by putting most of it in the template). If we get feedback that the repeated text is painful, we can revisit.

mystiberry-okta: @bdemers What's the reason behind not having a collection per year? Because a year is an artificial boundary? We'll have to make a separate collection eventually (or is the idea that we'd just delete stuff after 4-5 years? I'm building as one file, but wondering what the theory is behind not having them by year?

bdemers:
The collection is of a type, in this case, release notes, year is an attribute of that type. We can easily add sorting in the future, but the year portion is arbitrary, so the collection will work the same way with 2 or a million.

But we likely wouldn’t sort them by year as that doesn’t provide any end user benefit. 

-Brian

_config.yml
mystiberry-okta: I made all these changes but don't know why it's okay to delete the things we deleted here.

Gemfile.lock
```diff
@@ -220,31 +80,25 @@ GEM
...
     yell (2.0.7)
 
 PLATFORMS
   ruby
 
 DEPENDENCIES
-  github-pages (= 160)
   html-proofer (= 3.7.2)
   jekyll-assets (= 2.3.2)
+  jekyll-paginate-v2 (= 1.9.0)
```
mystiberry-okta: I made all these changes--but not sure why we are deleting all the things we deleted here.

jmelberg-okta:
Closing due to #1596

PR #1535, opened 2017-12-05
Adding a new RSS feed: `feed-edgar.xml`.

rdegges:
This generates RSS articles with tweets embedded in them! This is used by our social media monitoring tool (meet edgar) to help automate some of our articles.

With this change, any "tweets" in markdown metadata will be used to create actual tweets for us =)


rdegges:
Thanks @jmelberg-okta and @bretterer <3

PR #1537, opened 2017-12-06
Dok 142771

mystiberry-okta:
## Description:
- We need to add one more stce about /logout not being a server-side call

### Resolves doc for:
 [OKTA-142771](https://oktainc.atlassian.net/browse/OKTA-142771)


nbarbettini:
I would suggest: `You should call this endpoint by redirecting the user's browser, not as a server-side API call.`

mystiberry-okta:
Re-using this branch is causing trouble. Will resubmit to Fed in another PR. I took all three inputs from Nate, Reman, and Sohaib and combined them.

PR #1544, opened 2017-12-07
Fixed navbar gap

bretterer:
## Description:
- Fixes the gap in the navbar.  Still investigating why this just showed up.

The Before:

screen shot 2017-12-07 at 12 13 33 pm

And the after:

screen shot 2017-12-07 at 1 02 32 pm

### DIFF
--- a/_source/_assets/css/okta/components/_PageContent.scss
+++ b/_source/_assets/css/okta/components/_PageContent.scss
@@ -802,8 +802,7 @@
 
 @include media('<medium') {
 	.PageContent {
-		padding-top: 0px;
-		margin-top: -5px;
+		 padding-top: 0;
 
 		&-main {
 			@include padding(medium small);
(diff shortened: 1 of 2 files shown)


_source/_assets/css/okta/components/_PageContent.scss
```diff
@@ -802,7 +802,8 @@
 
 @include media('<medium') {
 	.PageContent {
-		 padding-top: 0;
+		padding-top: 0px;
+		margin-top: -5px;
```
jmelberg-okta: I think what we are looking for is in the `DynamicSideBar`. We shouldn't be introducing negative margins.

I think removing this [line should do the trick](https://github.com/okta/okta.github.io/blob/dd67a2cb85f0f97a692bdcf2885243ffc24e1f2f/_source/_assets/css/okta/components/_Sidebar.scss#L115)

PR #1559, opened 2017-12-15
Minor edits to Preact blog post

nbarbettini:
* IMO, we no longer need to say that the Platform choices include OpenID Connect, because it is implied in the new dev console.

* The proper terminology is [Higher-Order Components](https://reactjs.org/docs/higher-order-components.html).


LindsayB610:
H2 change lgtm! Go forth! :)

Issue #1545, opened 2017-12-08
Useful System Log Queries link is broken


nicolastriballier-okta:
https://developer.okta.com/docs/api/resources/system_log.html#examples > “Useful System Log Queries” link is broken

nbarbettini:
Thanks! Tracking in #1562

nbarbettini:
Turns out the link is fine, it's this extra query param that's causing it to break:

![url-gack](https://user-images.githubusercontent.com/7525482/34125321-c1037e78-e3ea-11e7-9b21-979fc5f989b7.png)

@robertjd can you link this issue to the open one you mentioned for fixing the ga params?

jmelberg-okta:
Closing due to duplicated issue #1562

nbarbettini:
@jmelberg-okta This is still an issue. #1562 was a PR with an attempted fix, not a dupe of this. 😄 

@robertjd had mentioned that there was another issue open for fixing this problem.

jmelberg-okta:
@natebarbettini-okta This problem is actually due to GA running on `support.okta.com`, not `developer.okta.com`. We cannot control this :bug:

nbarbettini:
How so? If I disable JavaScript on `developer.okta.com` and click the link, the page loads fine.

It looks like a script on `developer.okta.com` is adding the query param when the link is clicked:

![image](https://user-images.githubusercontent.com/7525482/34235125-5ba33f30-e5a4-11e7-9dd5-f9a4a8abf1d9.png)

Maybe I'm missing something here.

jmelberg-okta:
As discussed, this isn't an issue with `developer.okta.com`, but an issue with a script running on `support.okta.com`. Selecting certain `support.okta.com` links cause a redirect loop that eventually turn into a broken link (some require auth to proceed). Our link checking script picks up that this is a valid link, as the GA query params are not picked up.

The correct team has been notified of this bug.

nbarbettini:
Reopening since the link is still broken.

jmelberg-okta:
Resolved 1/19

nbarbettini:
Just tested and it works. Thanks @jmelberg-okta!

PR #1562, opened 2017-12-18
Remove broken link

nbarbettini:
As noted in #1545, this link is broken.


robertjd:
This link is valid, but an analytics script is appending query params to it in the browser.  Try removing those on the final link to see what I mean.  There's an open issue for this, I'll follow up tomorrow.

nbarbettini:
Oh man, you're totally right. I didn't even see that. Thanks for catching.

PR #1578, opened 2017-12-22
Add ASP.NET 4.x code flow quickstart

nbarbettini:
## Description:
No more "coming soon" for ASP.NET 4.x! 🎉

PR #1581, opened 2017-12-23
Update ASP.NET Core code flow quickstart for consistency

nbarbettini:
Updates the ASP.NET Core quickstart to be more consistent with #1578


nbarbettini:
Thanks @jmelberg-okta!

PR #1585, opened 2017-12-28
Adds Vue.js quickstart

jmelberg-okta:
## Description:
- :seedling: Adds Vue.js quickstart

> This will fail tests until [this PR](https://github.com/okta/okta-oidc-js/pull/76) is merged into `master`

### Resolves:
* [OKTA-153617](https://oktainc.atlassian.net/browse/OKTA-153617)

### DIFF
--- a/_source/quickstart-fragments/vue/default-example.md
+++ b/_source/quickstart-fragments/vue/default-example.md
@@ -1,0 +1,179 @@
+---
+layout: quickstart_partial
+exampleDescription: Vue Implicit
+---
+
+## Okta Vue.js Quickstart
+
+This guide will walk you through integrating authentication into a Vue app with Okta by performing these steps:
+
+1. Add an OpenID Connect Client in Okta
+2. Install the Okta Vue.js SDK
+3. Attach Components to Routes
+4. Use the Access Token
+
+At the end of the Vue instructions you can choose your server type to learn more about post-authentication workflows, such as verifying tokens that your Vue application can send to your server.
+
+## Prerequisites
+* If you do not already have a **Developer Edition Account**, you can create one at [https://developer.okta.com/signup/](https://developer.okta.com/signup/).
+* If you don't have a Vue app, or are new to Vue, please start with the [Vue CLI](https://github.com/vuejs/vue-cli) guide. It will walk you through the creation of a Vue app, creating [routers](https://router.vuejs.org/en/essentials/getting-started.html), and other Vue.js development essentials.
+
+## Add an OpenID Connect Client in Okta
+In Okta, applications are OpenID Connect clients that can use Okta Authorization servers to authenticate users.  Your Okta Org already has a default authorization server, so you just need to create an OIDC client that will use it.
+* Log into the Okta Developer Dashboard, click **Applications** then **Add Application**.
+* Choose **Single Page App (SPA)** as the platform, then populate your new OpenID Connect application with values suitable for your app. If you are running this locally and using the defaults from the [Vue CLI](https://github.com/vuejs/vue-cli), your `port` will be `8080`:
+
+| Setting             | Value                                          |
+| ------------------- | ---------------------------------------------- |
+| App Name            | My SPA App                                     |
+| Base URIs           | http://localhost:{port}                        |
+| Login redirect URIs | http://localhost:{port}/implicit/callback      |
+| Grant Types Allowed | Implicit                                       |
+
+After you have created the application there are two more values you will need to gather:
+
+| Setting       | Where to Find                                                                  |
+| ------------- | ------------------------------------------------------------------------------ |
+| Client ID     | In the applications list, or on the "General" tab of a specific application.   |
+| Org URL       | On the home screen of the developer dashboard, in the upper right.             |
+
+
+These values will be used in your Vue application to setup the OpenID Connect flow with Okta.
+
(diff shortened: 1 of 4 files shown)


_source/quickstart-fragments/vue/default-example.md
```diff
@@ -0,0 +1,179 @@
...
+In your application's [vue-router](https://router.vuejs.org/en/essentials/getting-started.html) configuration, import the `@okta/okta-vue` plugin and pass it your OpenID Connect client information:
+
+```typescript
+// router/index.js
+
+import Auth from '@okta/okta-vue'
+
+Vue.use(Auth, {
+  issuer: 'https://{yourOktaDomain}.com/oauth2/default',
+  client_id: '{client_id}',
```
nbarbettini: I've seen `{yourClientId}` and `{clientId}` in other quickstarts. I think we should pick one style and stick with it, what do you think? I'd vote for `{clientId}`.

_source/quickstart-fragments/vue/default-example.md
```diff
@@ -0,0 +1,179 @@
...
+      // Navigate back to home
+      this.$router.push({ path: '/' })
+    }
+  }
+}
+</script>
+```
+
+### Create the Callback Handler
+In order to handle the redirect back from Okta, you need to capture the token values from the URL. You'll use `/implicit/callback` as the callback URL, and use the default `Auth.handleCallback()` component included.
```
nbarbettini: Suggestion: "Your Vue app needs to handle the redirect the redirect back from Okta. You can do this by wiring up the `/implicit/callback` route to the `Auth.handleCallback()` component from the Okta Vue.js SDK:"

_source/quickstart-fragments/vue/default-example.md
```diff
@@ -0,0 +1,179 @@
...
+```bash
+npm install @okta/okta-vue --save
+```
+
+### Configuration
+You will need the values from the OIDC client that you created in the previous step to instantiate the middleware. You will also need to know your Okta Org URL, which you can see on the home page of the Okta Developer console.
+
+In your application's [vue-router](https://router.vuejs.org/en/essentials/getting-started.html) configuration, import the `@okta/okta-vue` plugin and pass it your OpenID Connect client information:
+
+```typescript
```
nbarbettini: Any reason to use typescript highlighting if the file is `.js`?

_source/quickstart-fragments/vue/default-example.md
```diff
@@ -0,0 +1,179 @@
...
+import axios from 'axios'
+
+export default {
+  data () {
+    return {
+      posts: []
+    }
+  },
+  async created () {
+    axios.defaults.headers.common['Authorization'] = `Bearer ${await this.$auth.getAccessToken()}`
```
nbarbettini: Is this setting a default header on the axios singleton? I think it'd be better to show either:
* Attaching the Bearer header to this GET call only
* Setting the axios defaults in a more general place (maybe `router/index.js`?)

_source/quickstart-fragments/vue/default-example.md
```diff
@@ -0,0 +1,179 @@
...
+    ...
+  ]
+})
+```
+
+## Use the Access Token
+
+When your users are authenticated, your Vue application has an access token that was issued by your Okta Authorization server. You can use this token to authenticate requests for resources on your server or API. As a hypothetical example, let's say you have an API that provides messages for a user. You could create a `MessageList` component that gets the access token and uses it to make an authenticated request to your server.
+
+Here is what the Vue component could look like for this hypothentical example using [axios](https://github.com/axios/axios):
```
nbarbettini: nit: hypothetical

_source/quickstart-fragments/vue/default-example.md
```diff
@@ -0,0 +1,179 @@
...
+  routes: [
+    { path: '/implicit/callback', component: Auth.handleCallback() },
+    ...
+  ]
+})
+```
+
+## Use the Access Token
+
+When your users are authenticated, your Vue application has an access token that was issued by your Okta Authorization server. You can use this token to authenticate requests for resources on your server or API. As a hypothetical example, let's say you have an API that provides messages for a user. You could create a `MessageList` component that gets the access token and uses it to make an authenticated request to your server.
```
nbarbettini: Suggestion: "When your users sign in, your Vue app gets an access token from Okta. You can use this token to make secure requests to to your server or API. (para break)

As a hypothetical example..."

mraible:
@jmelberg-okta Is this one ready-to-go?

jmelberg-okta:
@mraible Not quite yet as the artifact still needs to be published. I'm targeting EOD for release.

jmelberg-okta:
@nbarbettini & @mraible: I'll get this rolled out by EOD today. Still need to address some of the feedback here.

_source/quickstart-fragments/vue/default-example.md
jmelberg-okta: In an ideal case, adding the Authorization header will be done deliberately in `router/index.js`. Since we wanted to keep this short and in one file, lets keep this here for now.

jmelberg-okta:
@nbarbettini Thank you for the feedback! Most of your requested changes can be applied to all of our frontend quickstarts, so I'll update all of those at once.

PR #1591, opened 2017-12-29
:seedling: Adds markdown linter

jmelberg-okta:
## Description:
- Runs various lint checks on markdown files to keep style consistency.
- **NOTE** Doesn't include API docs (> 2000 errors)

To run the lint script:

```
$ npm run lint
```

Tests:
- Header levels should only increment by one level at a time
- Inconsistent indentation for list items at the same level
- Consider starting bulleted lists at the beginning of the line
- Unordered list indentation
- Trailing spaces
- Reversed link syntax
- Multiple consecutive blank lines
- No space after hash on atx style header
- No space inside hashes on closed atx style header
- Multiple spaces inside hashes on closed atx style header
- Multiple spaces after hash on atx style header
- Multiple spaces after blockquote symbol
- Blank line inside blockquote
- Headers must start at the beginning of the line
- Fenced code blocks should be surrounded by blank lines


mystiberry-okta:
Sadly, some of these errors are done deliberately (to work around another issue) in the api docs. Many are not. The two I know we need to address are Hx/Hx+2, and I think we have a single black line in the middle of some block quotes because we need paragraphs. 

Also, assuming nested lists are still possible, that these fixes are only for unneeded spaces...

mystiberry-okta:
Also, should we document these tests somewhere so a writer knows what's up?

jmelberg-okta:
Closing for now - probably not a good idea to tell doc writers how to "write" :)

PR #1593, opened 2018-01-02
Conductor integration

uguraktepe-okta:
## Description:
These changes:
- Generate conductor file by looking at all index.html's to setup redirections
- Adds a postpublish script to remove .html extension from all files
- Sends promotion messages for conductor that will trigger automatic deployments

### Resolves:
[OKTA-152047](https://oktainc.atlassian.net/browse/OKTA-152047)
[OKTA-147564](https://oktainc.atlassian.net/browse/OKTA-147564)

@jmelberg-okta @robertdamphousse-okta @rchild-okta @bdemers @jtodd-okta @jonhohle-okta

### DIFF
--- a/_source/conductor.yml
+++ b/_source/conductor.yml
@@ -1,0 +1,15 @@
+# This file instructs conductor to set up S3 redirections. Each redirect has a from and to property, and
+# conductor will set up a 301 redirect for each entry.
+# For the 'from:' attribute, include the leading slash, but not a leading slash.
+# For the 'to:' attribute, any valid URL path can be used, but make sure to include the leading slash.
+# Some examples:
+#   - from: /old.html
+#     to: /new.html
+#   - from: /old-path
+#     to: /new-path
+#   - from: /iossetupinstructions.html
+#     to: /ios/setup/instructions.html
+
+redirects:
+  - from: /swift
+    to: /code/ios/
--- a/scripts/helpers.sh
+++ b/scripts/helpers.sh
@@ -164,7 +164,7 @@ function header_checker() {
 function check_sample_code_orgs() {
     # Sample code URLS should be in the following format:
     # Currently: https://your-org.okta.com
-    
+
     local dir=$(pwd)
     local yourOrgUrls=$(grep -EoR "(http|https)://your-org.okta*" --include="*.md" $dir --exclude-dir={node_modules,scripts,tests,dist} | sort | uniq)
     local yourExampleUrls=$(grep -EoR "(http|https)://example.okta*" --include="*.md" $dir --exclude-dir={node_modules,scripts,tests,dist} | sort | uniq)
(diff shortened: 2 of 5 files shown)


_source/conductor.yml
```diff
@@ -0,0 +1,3 @@
+redirects:
```
jtodd-okta: Very cool. Is there a README somewhere which details what's allowed in this file? I think it'd be very useful to have that and a link here to that doc so that folks less familiar with this than yourself can self-serve.

uguraktepe-okta: What you see is pretty much it. @jmelberg-okta any preference on where to put that documentation? Normally I'd just put it in the internal wiki and link to it from here but treading carefully since this is a public repo.

scripts/setup.sh
```diff
@@ -78,6 +78,26 @@ function generate_html() {
     fi
 }
 
+function removeHTMLExtensions() {
+    interject "Removing HTML extensions"
+    # Removing all generated .html files (excludes the main 'index.html' in the dir)
+    find ./dist/ -type f ! -iname 'index.html' -name '*.html' -print0 | while read -d $'\0' f; do mv "$f" "${f%.html}"; done
+}
+
+function generate_conductor_file() {
```
jtodd-okta: Is there an earlier build step that we could tie this to?

I'm assuming that publish doesn't run with every CI loop and only happens on promote. My concern is that we don't exercise this logic until promote time and thus don't get an early warning when it's broken / coverage by CI.

bdemers: I might be missing something here, but why remove them? Why not just symlink/copy them?

_source/conductor.yml
jmelberg-okta: @uguraktepe-okta I'm not sure if adding instructions to our [repo wiki](https://github.com/okta/okta.github.io/wiki) is the best place, since adding to this file is only for edge-cases. Instead, can we add a block comment in this file explaining what this is for?

scripts/setup.sh
bdemers: @jmelberg-okta is this something that can be configured in Jekyll?

jmelberg-okta: @bdemers That is a good point. I'm thinking we should add redirect logic into the `.html` files to redirect to the page without the trailing extension.

As an alternative (or first pass), we could duplicate the file:
  - `page.html` -> Available via `https://developer.okta.com/page.html`
  - `page` -> Available via `https://developer.okta.com/page`

robertjd: @bdemers since this is going to a CDN the paths have to match exactly.  I prefer renaming rather than copying so that there is only one path per page.  I agree with @jtodd-okta , that we should be doing this as part of the build process to that breakage is caught before trying to push to the CDN

uguraktepe-okta: Following this, I will pull this part of the code out of this PR, looks like you still need to reach consensus on this one. We can still go ahead with the rest of the changes and set up CD to trex.

scripts/helpers.sh
```diff
@@ -232,3 +232,10 @@ function fold() {
     ${command}
     echo -en "travis_fold:end:${name}\\r"
 }
+
+function send_promotion_message() {
+    curl -H "Authorization: Bearer ${TESTSERVICE_SLAVE_JWT}" \
+      -H "Content-Type: application/json" \
+      -X POST -d "[{\"artifactId\":\"$1\",\"repository\":\"npm-okta\",\"artifact\":\"$2\",\"version\":\"$3\",\"promotionType\":\"ARTIFACT\"}]" \
```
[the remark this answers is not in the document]
uguraktepe-okta: Yes, at least what the current publish script uses:
`REGISTRY="${ARTIFACTORY_URL}/api/npm/npm-okta"`

PR #1607, opened 2018-01-09
Adds extensionless files + 302 redirects

jmelberg-okta:
## Description:
- 🔥  Removes `.html` extension on all pages
- 🔁 Adds `302` redirect from `/page.html` -> `/page`
- Updates `post-build-prod` script to only copy `assets` (was previously copying all files from `master`)

## Tests:
- I'm leaning toward **not** implementing `.html` extension tests. Since the default behavior is to redirect from `page.html` -> `page`, this is enforcing a style link writing onto the writer.

- Updated tests [here](https://github.com/okta/okta.github.io/pull/1607/files#diff-532c78fa98ecc2208244893fb71966f3R39)

### Resolves:
* [OKTA-154192](https://oktainc.atlassian.net/browse/OKTA-154192)


scripts/helpers.sh
```diff
@@ -247,3 +247,20 @@ function send_promotion_message() {
...
       -k "${APERTURE_BASE_URL}/v1/artifact-promotion/createPromotionEvent"
 }
+
+function removeHTMLExtensions() {
+    # Removing all generated .html files (excludes the main 'index.html' in the dir) and
+    # create 302 redirects to extensionless pages
+    find ./dist -type f ! -iname 'index.html' -name '*.html' -print0 | while read -d $'\0' f
+    do
+        
+        if [ -a `echo ${f%.html}` ] ;
```
robertjd: `-a` is deprecated: https://www.tldp.org/LDP/abs/html/fto.html

Also, why is this check necessary?

jmelberg-okta: We want to make sure **if** the conversion has already taken place, we do not attempt to do this again.

It is an unlikely scenario, but a simple check to avoid the possibility of having our `.html` and extension-less pages both be redirects.

PR #1608, opened 2018-01-09
Weekly

mystiberry-okta:
## Description:
- We need to pull changes

### Resolves:

No JIRA, we're just getting the new release notes into weekly. But I expected to see the new files.


mystiberry-okta:
So, this PR is pushing weekly into source, but we need to push source into weekly, so closing.

PR #1599, opened 2018-01-05
Docs for Email Operations APIs (Beta)

amoghvasekar-okta:
## Description:
- Docs for Email Operations APIs (Beta)
- This is still WIP.

### Resolves:
* [OKTA-153174](https://oktainc.atlassian.net/browse/OKTA-153174)

PR #1610, opened 2018-01-09
Docs for Email Operations APIs (Beta)

amoghvasekar-okta:
## Description:
- Docs for Email Operations APIs (Beta)
- Two questions 
1. There doesn't seem a good pattern to document "Errors". I've used my intuition, but let me know if you think it should be formatted differently.
2. The object model is included within the section where operations are listed. As an example, the (unrelated) pre-existing "Clients Grant" lists the model with the "User" model. However, that seems confusing and misleading since the object is not a part of the "User" model. The new Email object model I have added is kept close to the actual operations, but let me know if there is any pattern to follow here.

- Old PR with some comments in https://github.com/okta/okta.github.io/pull/1599

### Resolves:
* [OKTA-153174](https://oktainc.atlassian.net/browse/OKTA-153174)

### DIFF
--- a/_source/_docs/api/resources/users.md
+++ b/_source/_docs/api/resources/users.md
@@ -3111,6 +3111,405 @@ curl -v -X GET \
 ]
 ~~~
 
+## User Email Operations
+
+{% api_lifecycle beta %}
+
+These operations can be used to manage a user's emails.
+
+### User Email Object
+
+| Property    | Description                                 | Datatype                                                        | Unique |
+|:------------|:--------------------------------------------|:----------------------------------------------------------------|:-------|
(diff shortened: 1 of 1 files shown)


mystiberry-okta:
1. Where you put the Errors is fine. It matches what we do in newer API topics.
2. Typically we put all the models at the end but your Actions object has only one property, tightly linked to the request param so let's leave it here.

mystiberry-okta:
HOLD DON'T MERGE until after 2018.02 goes out to preview.

mystiberry-okta:
@jakubvul-okta This doc is for 2018.03 -- you should OKTA-153174 in the list for RN.

PR #1623, opened 2018-01-11
Adds gh-pages test for known repos

jmelberg-okta:
## Description:
- Adds HTTP status code checks for the following repos that serve their documentation off of GitHub pages:
    - okta-sdk-java
    - okta-sdk-php
    - okta-sdk-nodejs
    - okta-sdk-dotnet

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -5,6 +5,7 @@ description: > # this means to ignore newlines until "baseurl:"
   Secure, scalable, and highly available authentication and user management for any app.
 baseurl: ""
 url: "https://developer.okta.com"
+weekly_url: "http://developer.trexcloud.com"
 twitter_username: OktaDev
 github_username:  oktadeveloper
 disqus_shortname: oktadevblog
@@ -130,3 +131,12 @@ dev_forum:  https://devforum.okta.com/
 
 # Location of main Drupal site for navigation and shared resources
 external_domain: https://developer.okta.com
+
+github_pages_docs:
+  - okta-sdk-java
+  - okta-sdk-php
+  - okta-sdk-nodejs
+  # - okta-sdk-python
+  - okta-sdk-dotnet
+  # - okta-jwt-verifier-java
+  # - okta-spring-boot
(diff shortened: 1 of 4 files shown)


_config.yml
```diff
@@ -130,3 +131,12 @@ dev_forum:  https://devforum.okta.com/
 
 # Location of main Drupal site for navigation and shared resources
 external_domain: https://developer.okta.com
+
+github_pages_docs:
+  - okta-sdk-java
+  - okta-sdk-php
+  - okta-sdk-nodejs
+  # - okta-sdk-python
```
bdemers: Why are these ones commented out ?

_config.yml
```diff
@@ -130,3 +131,12 @@ dev_forum:  https://devforum.okta.com/
...
+  - okta-sdk-dotnet
+  # - okta-jwt-verifier-java
```
bdemers: I'm not sure about all of these projects but i know some of the java ones 404 at the root, but not at /apidocs

https://developer.okta.com/okta-jwt-verifier-java/ 
vs
https://developer.okta.com/okta-jwt-verifier-java/apidocs/
(for the java repos this will be resolved in the near future similar to how the Java SDK works:
https://developer.okta.com/okta-sdk-java/

jmelberg-okta: Interesting. I though we set the behavior of all the repos to have /latest shown at root. But good to know -- so I'll update the links

jmelberg-okta:
New plan: Since it doesn't make sense to check for this on every commit, we should move this logic into a task that gets run in Bacon every 24hrs, or setup in Ghost Inspector.

_config.yml
bdemers: There could be other types of docs so `/apidocs/` is the latest, and `/${version}/apidocs/` is also deployed

bdemers:
I'm not sure how often we are checking it now, but we could check it hourly (so we don't have an average of a 12 delay for the notification of an issue)

Issue #1085, opened 2017-07-24
Incorrect information about password length in docs


omgitstom:
https://developer.okta.com/docs/api/resources/users.html#password-object

References 40 as the max, but this is incorrect, please work with the UD team to update this with the correct value

jmelberg-okta:
@mystiberry-okta Can we ping UD for the correct length? This should be an easy fix.

mystiberry-okta:
pinged.

brianzuzga-okta:
You'll want to consult with the auth team to really get the definitive answer on this.  I want to make sure there are no legacy use cases that get this all confused.

My general understanding is now that we try to store password hashes whenever possible, the password length is essentially unlimited.  I have seen some failures where a long password gets set, but I assume we're storing it in plaintext in the ADAppUser for some reason.
https://oktainc.atlassian.net/browse/OKTA-139818
The length limit on that field is 256 characters.  Directories wasn't sure about the use case.

daytonwilliams-okta:
We are definitely storing the Okta password hashed.  I know this because you can set your Okta password to include four byte UTF8, which literally cannot be written to our database.

omgitstom:
This got lost in the weeds. We need to update the table in the link in the issue with the correct value.  Right now it is incorrect.

As a user of the API, I want to know the max length of the password field, so I can add validation logic to my code without making an API call to Okta with an Error.

@daytonwilliams-okta ^

mystiberry-okta:
Jira created and assigned to DevEx Triage.

omgitstom:
I don't think devex needs to triage this. I already have, we just need to get the correct value and update the docs from the UD team.

A POST to `/users` with:

```
{
  "profile": {
    "firstName": "Isaac",
    "lastName": "Brock",
    "email": "isaac@{{email-suffix}}",
    "login": "isaac@{{email-suffix}}"
  },
  "credentials": {
    "password" : { "value": "Stormpath1Stormpath1Stormpath1Stormpath1Stormpath1" }
  }
}```

`Stormpath1` is a 10 character string, you can see that I can create a 50 character password without an issue.

mystiberry-okta:
Triage is the auto assignment, if we don’t want to use that for doc bugs, no problem, I just didn’t know. Should I take this bug or should Jakub?

Mysti

On Nov 27, 2017, at 8:42 AM, Tom <notifications@github.com<mailto:notifications@github.com>> wrote:


[...]

PR #1625, opened 2018-01-11
Correct max length for password

jakubvul:
## Description:
- Password max length is doc'd as `40` but should be `72`.

### Resolves:
* [OKTA-144636](https://oktainc.atlassian.net/browse/OKTA-144636)


jmelberg-okta:
Resolves: #1085

PR #1629, opened 2018-01-12
:arrow_up: Updates Gem dependencies

jmelberg-okta:
## Description:
A number of our libraries have improved performance recently. This cuts down our local build time by nearly 40% on average.

- `npm start`:
    - Previous: `43.657 seconds`
    - Average: `27.893 seconds`


Gemfile
```diff
@@ -1,7 +1,8 @@
 source 'https://rubygems.org'
 
-gem 'github-pages', '160'
+gem 'github-pages'
 gem 'jekyll-assets', '2.3.2'
```
jmelberg-okta: We require `2.3.2` to ensure assets are loaded properly.

_source/_layouts/base.html
```diff
@@ -12,7 +12,7 @@
     <!-- END Footer -->
   </div>
   <!-- START Post Footer -->
-  {% include post-footer.html %}
+  {% include_cached post-footer.html %}
```
jmelberg-okta: Since we include this in almost every file, instead of looking for this file and including it **everytime**, we use a cached version.

This shouldn't be problematic because the content on all of these pages are the same.

nbarbettini:
Ran it from a fresh clone on Windows 10:

```
git clone git@github.com:okta/okta.github.io.git
git checkout jv-update-deps
npm install
bundle install
npm start

Configuration file: C:/Users/Nate Barbettini/Documents/temp/okta.github.io/_config.yml
            Source: _source
       Destination: dist
 Incremental build: enabled
      Generating...
                    done in 53.696 seconds.
  Please add the following to your Gemfile to avoid polling for changes:
    gem 'wdm', '>= 0.1.0' if Gem.win_platform?

 Auto-regeneration: enabled for '_source'
    Server address: http://127.0.0.1:4000/
  Server running... press ctrl-c to stop.
```

Worked just fine! 👍  It used to take 60-90 seconds. Thanks man!

PR #1642, opened 2018-01-18
change post to put

mystiberry-okta:
## Description:
- POST should be PUT for step two (add the whitelist to your app profile)
- I tested and confirmed POST always returns a "HTTP method not supported" error, and PUT does work.

### Resolves:

[OKTA-155553](https://oktainc.atlassian.net/browse/OKTA-155553)


nbarbettini:
Thanks for testing! :)

PR #1664, opened 2018-01-22
:fire: Remove dead file

jmelberg-okta:
## Description:
- Removes a dead file copied over when overlaying the **old** master branch with the new `dist` output.

NOTE: This is the **ONLY** way we can remove files currently. 

This PR will be **reverted** once the site has been deployed.


scripts/gh-pages-publish.sh
```diff
@@ -25,6 +25,9 @@ then
     exit 1
 fi
 
+# Remove the dead file (THIS IS A ONE-TIME-ONLY THING)
+rm -rf ${GENERATED_SITE_LOCATION}/docs/api/resources/resrouce-server-beta
```
jmelberg-okta: This is intentionally spelled this way, as we generated it [here](https://github.com/okta/okta.github.io/tree/master/docs/api/resources/resrouce-server-beta)

bdemers:
But this might not be an issue long term anyway...

PR #1670, opened 2018-01-23
Spelling

danmarma-okta:
Very minor update

## Description:
- Updated one sentence in the hashing algorithm section.

### Resolves:
*N/A


bdemers:
Seems fine to me, more word smithing then spell correcting though.

nbarbettini:
I agree w/ @jakubvul - that's a simpler way of getting the point across.

mystiberry-okta:
@natebarbettini-okta or @jakubvul  just to verify, bcrypt is the only hashing function we use for anything?

nbarbettini:
@mystiberry-okta @jakubvul My understanding is: bcrypt is the only hashing function allowed for password **import**. We should double check this with the UD team and then merge this change.

mystiberry-okta:
Confirmed w/auth team.

nbarbettini:
Thanks @mystiberry-okta! I think we should go with Jakub's suggestion and then merge it.

PR #1679, opened 2018-01-25
:fire: Removes dead pages

jmelberg-okta:
## Description:
- Removed files that have been around for > 1 year w/o updates or page views


mystiberry-okta:
Will the corresponding HTML files be removed? For example: https://okta.github.io/use_cases/sso_and_provisioning/draft.html

jmelberg-okta:
In a perfect world, yes. But that will not happen today until we adjust our publish process. The page you referenced was indexed in Google, causing traffic to navigate there. FYI in the last year, we've had 21 unique page views.

I'm adding a redirect from that page just to ensure we're covered from `404` pages in the future.

PR #1680, opened 2018-01-26
:bug: Fix sitemap

jmelberg-okta:
## Description:
- :bug: `sitemap.xml` was only rendering the `posts` collection (blogposts) since the logical comparison used for `site.pages` was **always true**.
- Updated the `sitemap` template to be more explicit, referencing a list of **collections** - and ignoring accordingly.
- 🔁 Placed all previous `sdk` docs inside of the new collection `sdk`
   - `/sdk/core/cshard_api_sdk`
   - `/sdk/core/java_api_sdk`
   - `/sdk/corepython_api_sdk`
   - ...

> Note: I attempting using the [jekyll-sitemap](https://github.com/jekyll/jekyll-sitemap) plugin and modifying the [front matter](https://jekyllrb.com/docs/frontmatter/) with no success.


_config.yml
```diff
@@ -99,21 +97,14 @@ defaults:
...
-      sitemap: false
-  -
-    scope:
-      collection: change-log
-    values:
-      sitemap: false
+
+sitemap_ignore:
+  - sdk
+  - change-log
```
bdemers: We should include this too

PR #1684, opened 2018-01-26
Updates including fixing heading levels and adding test plan

mystiberry-okta:
## Description:
- Adding a test plan for scim server and okta, and fixing headings so right-hand nav does the right thing.

### Resolves:

 [OKTA-155843](https://oktainc.atlassian.net/browse/OKTA-155843)


mystiberry-okta:
@karanharibhakti-okta You can always send me updates if needed.

PR #1686, opened 2018-01-27
:white_check_mark: Trigger Runscope tests on site publish

jmelberg-okta:
## Description:
- ✅  Adds trigger for Runscope API tests on all `sitemap.xml` links


scripts/publish.sh
```diff
@@ -123,4 +123,7 @@ if ! send_promotion_message "${DEPLOY_ENVIRONMENT}" "${ARTIFACT}" "${DEPLOY_VERS
   exit ${BUILD_FAILURE};
 fi
 
+# Trigger Runscope tests
+curl -I -X GET 'https://api.runscope.com/radar/bucket/dd1974ff-3203-4c05-9818-ac1a7637b586/trigger?base_url=https://dq3iyfxeowfmd.cloudfront.net'
```
vijetmahabaleshwar-okta: We can probably add the prod and staging base urls into travis env vars.

jmelberg-okta: I was debating doing this as well. I'll update it to point to a travis var.

johnsheehan:
You should not publish Trigger URLs as anyone can initiate tests on your account with them. Please contact help@runscope.com to cycle the Trigger URL posted above.

nbarbettini:
@johnsheehan Is that true even if the trigger ID is pulled from an environment variable, as it is in this case (`$RUNSCOPE_TRIGGER_ID`)?

johnsheehan:
Env variable is what you want to use, correct. I thought I saw an earlier commit that had the UUID hard coded, but maybe it was removed or I was seeing things.

jmelberg-okta:
@johnsheehan You're correct - we pushed it originally with the `triggerId` visible. This PR has been updated to use secure environment variables.

Thanks for reaching out!

nbarbettini:
All good @johnsheehan, it must have been fixed by the time I looked at it too. You're not going crazy. 😄

PR #1714, opened 2018-02-02
OKTA-156281: Detail AuthenticationContext Object

ericzhou-okta:
## Description:
- Clearer AuthenticationContext docs.

screen shot 2018-02-09 at 04 01 00
screen shot 2018-02-09 at 04 01 08
screen shot 2018-02-09 at 03 37 48

### Resolves:
* [OKTA-156281](https://oktainc.atlassian.net/browse/OKTA-156281)

### DIFF
--- a/_source/_docs/api/resources/system_log.md
+++ b/_source/_docs/api/resources/system_log.md
@@ -359,18 +359,47 @@ This object provides a way to store additional text about an event for debugging
 
 ### AuthenticationContext Object
 
-Describes authentication data for an event.
+All authentication relies on validating one or more credentials that proves the authenticity of the actor's identity. Credentials are sometimes provided by the actor, as is the case with passwords, and at other times provided by a third party, and validated by the authentication provider.
+
+The **AuthenticationContext** contains metadata about how the actor is authenticated. For example, an **authenticationContext** for an event in which a user authenticates with IWA will look something like:
+
+```json
+{
+    "AuthenticationProvider": "ACTIVE_DIRECTORY",
+    "AuthenticationStep": 0,
+    "CredentialProvider": null,
+    "CredentialType": "IWA",
+    "ExternalSessionId": "102N1EKyPFERROGvK9wizMAPQ",
+    "Interface": null,
+    "Issuer": null
+}
+```
+In such a case, one can recognize that the user used an IWA credential to authenticate against an Active Directory instance. All of the user's future generated events in this login session will share the same **externalSessionId**.
+ 
+Among other operations, this response object can be used to scan for suspicious login activity or perform analytics on user authentication habits (e.g. how often authentication scheme X is used versus authentication scheme Y).
 
 |------------+----------------------------------------------------------------+-----------------+----------+-----------+-----------|
 | Property   | Description                                                    | DataType        | Nullable | MinLength | MaxLength |
 | ---------- | -------------------------------------------------------------- | --------------- | -------- | --------- | --------- |
-| authenticationProvider | Type of authentication provider: `OKTA_AUTHENTICATION_PROVIDER`, `ACTIVE_DIRECTORY`, `LDAP`, `FEDERATION`, `SOCIAL`, `FACTOR_PROVIDER` | String | TRUE  | | |
-| credentialProvider | Type of credential provider: OKTA_CREDENTIAL_PROVIDER, RSA, SYMANTEC, GOOGLE, DUO, YUBIKEY | Array of String | TRUE  |           |           |
-| credentialType | Type of credential: `OTP`, `SMS`, `PASSWORD`, `ASSERTION`, `IWA`, `EMAIL`, `OAUTH2`, `JWT` | String        | TRUE     |           |           |
-| issuer     | Issuer of the credential                                       | [Issuer Object](#issuer-object) | TRUE | |         |
-| externalSessionId | External Session identifier of the request              | String          | TRUE     | 1         | 255       |
-| interface  | Authentication interface                                       | String          | TRUE     | 1         | 255       |
+| authenticationProvider | The system that proves the identity of an actor using the credentials provided to it | String | TRUE  | | |
+| credentialProvider | A credential provider is a software service that manages identities and their associated credentials. When authentication occurs via credentials provided by a credential provider, that credential provider will be recorded here. | Array of String | TRUE | | |
+| credentialType | The underlying technology/scheme used in the credential    | String          | TRUE     |           |           |
+| issuer     | The specific software entity that created and issued the credential. | [Issuer Object](#issuer-object) | TRUE | |   |
+| externalSessionId | A proxy for the actor's [session ID](https://www.owasp.org/index.php/Session_Management_Cheat_Sheet) | String | TRUE | 1 | 255 |
+| interface  | The third party user interface that the actor authenticates through, if any. | String | TRUE | 1        | 255       |
 |------------+----------------------------------------------------------------+-----------------+----------+-----------+-----------|
+
+###### Possible Values
+
+Some of the fields listed above have a finite set of possible values.
+
+|------------------------+-------------------------------------------------------------------------------------------------------|
+| Property               | Possible Values                                                                                       |
+| ---------------------- | ----------------------------------------------------------------------------------------------------- |
+| authenticationProvider | `OKTA_AUTHENTICATION_PROVIDER`, `ACTIVE_DIRECTORY`, `LDAP`, `FEDERATION`, `SOCIAL`, `FACTOR_PROVIDER` |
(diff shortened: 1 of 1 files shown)


ericzhou-okta:
Some descriptions ~shamelessly taken~ reworded from https://opencreds.org/specs/source/identity-credentials/#terminology

_source/_docs/api/resources/system_log.md
```diff
@@ -359,17 +359,20 @@ This object provides a way to store additional text about an event for debugging
...
-| issuer     | Issuer of the credential                                       | [Issuer Object](#issuer-object) | TRUE | |         |
-| externalSessionId | External Session identifier of the request              | String          | TRUE     | 1         | 255       |
-| interface  | Authentication interface                                       | String          | TRUE     | 1         | 255       |
+| authenticationProvider | The system that proves the identity of an actor using the credentials provided to it. `OKTA_AUTHENTICATION_PROVIDER`, `ACTIVE_DIRECTORY`, `LDAP`, `FEDERATION`, `SOCIAL`, `FACTOR_PROVIDER` | String | TRUE  | | |
+| authenticationStep | | Integer | FALSE  | | |
+| credentialProvider | The external software service that manages identities and their associated credentials. It will typically handle requests to store credentials issued by an issuer and to retrieve credentials when requested by a credential inspector. Examples: `OKTA_CREDENTIAL_PROVIDER`, `RSA`, `SYMANTEC`, `GOOGLE`, `DUO`, `YUBIKEY`. This field is used when the credential is given by a third party, such as a social IDP. | Array of String | TRUE  |           |           |
```
ericzhou-okta: This could definitely be reworded better. Any ideas?

_source/_docs/api/resources/system_log.md
```diff
@@ -359,17 +359,20 @@ This object provides a way to store additional text about an event for debugging
...
-| credentialProvider | Type of credential provider: OKTA_CREDENTIAL_PROVIDER, RSA, SYMANTEC, GOOGLE, DUO, YUBIKEY | Array of String | TRUE  |           |           |
-| credentialType | Type of credential: `OTP`, `SMS`, `PASSWORD`, `ASSERTION`, `IWA`, `EMAIL`, `OAUTH2`, `JWT` | String        | TRUE     |           |           |
-| issuer     | Issuer of the credential                                       | [Issuer Object](#issuer-object) | TRUE | |         |
-| externalSessionId | External Session identifier of the request              | String          | TRUE     | 1         | 255       |
-| interface  | Authentication interface                                       | String          | TRUE     | 1         | 255       |
+| authenticationProvider | The system that proves the identity of an actor using the credentials provided to it. `OKTA_AUTHENTICATION_PROVIDER`, `ACTIVE_DIRECTORY`, `LDAP`, `FEDERATION`, `SOCIAL`, `FACTOR_PROVIDER` | String | TRUE  | | |
+| authenticationStep | | Integer | FALSE  | | |
```
ericzhou-okta: Coming back to this later. It's only ever set to `0` in okta-core. Gonna do some asking around today to see if this field will ever have significance in the future.

_source/_docs/api/resources/system_log.md
bobtiernay-okta: Is `OKTA_CREDENTIAL_PROVIDER` considered "external"? For both this item and the one before it, it might be better to say "the internal or external" to spell out the possibilities.

_source/_docs/api/resources/system_log.md
```diff
@@ -359,17 +359,20 @@ This object provides a way to store additional text about an event for debugging
 
 ### AuthenticationContext Object
 
-Describes authentication data for an event.
+When an action requires the [actor](#actor-object)'s identity to be authenticated, authentication metadata will be stored in the **AuthenticationContext**.
+
+All authentication relies on validating a credential that proves the authenticity of the actor's identity. Credentials are sometimes provided by the actor, as is the case with passwords, and at other times provided by a third party, and validated by the authentication provider.
```
bobtiernay-okta: "validating one or more credentials".

You might want to swap the order of these sentences. Then you can refer to it in the second one, possibly throwing in a few examples references. You could also try to explain in what situations this information would be useful (a generally good idea for docs).

_source/_docs/api/resources/system_log.md
bobtiernay-okta: What about this:

https://github.com/okta/okta-core/blob/156c4f8d2df69c897fa2f50a396954dce45d6780/components/core/events/api/src/main/java/com/okta/monolith/core/events/dto/AuthenticationContext.java#L52

image

_source/_docs/api/resources/system_log.md
mystiberry-okta: You can simply say "The service that manages identities and their associated credentials. Use this property to handle requests to store credentials issued by `iss` (if that's true), and to retrieve credentials when requested by a {???} What's an example of a credential inspector? Also it's social IdP. Are the values the only valid values? How do they know what the valid values are?

_source/_docs/api/resources/system_log.md
ericzhou-okta: 😮 
Thanks for pointing this out. I missed this comment during my scan through Core. Although I'm still not totally clear about what a multistep authentication process is like. Is `Secondary` just referring to 2-Factor Auth, or is it something else?

bobtiernay-okta: I would `git blame` that line and follow up with the editor.

ericzhou-okta: So I talked with @mohammadrahimi-okta about this field, and from what he and I understand, this is supposed to be used for 2FA. However, even when we authenticate on the second factor, this value is still set to 0, and only ever set to 0. I can't in good faith add this field into the systemlog docs if it's giving out misleading information wrt what it was intended to do. Some unfortunate developer is going to try using this field to detect 2FA and experience sore disappointment.

ericzhou-okta: Talked to Alex Bovee about this and he confirmed usage of authenticationStep in 2FA. I'm going to put down the description for the intended usage, but clarify that it's only being set to 0 for now.

ericzhou-okta:
After several days of reading online authentication guides, talking to people, static analysis, and analyzing log data, this is the best descriptions I can come up with.

_source/_docs/api/resources/system_log.md
```diff
@@ -359,19 +359,52 @@ This object provides a way to store additional text about an event for debugging
...
+    "AuthenticationProvider": "ACTIVE_DIRECTORY",
+    "AuthenticationStep": 0,
+    "CredentialProvider": null,
+    "CredentialType": "IWA",
+    "ExternalSessionId": "102N1EKyPFERROGvK9wizMAPQ",
+    "Interface": null,
+    "Issuer": null
+}
+```
+ In such a case, one can recognize that a this user used an IWA credential to authenticate against an Active Directory instance. All of the user's future generated events in this login session will share the same **externalSessionId**.
```
bobtiernay-okta: Grammar:
> [...] that a this user [...]

_source/_docs/api/resources/system_log.md
```diff
@@ -359,19 +359,52 @@ This object provides a way to store additional text about an event for debugging
...
+|------------------------+-------------------------------------------------------------------------------------------------------|
+| Property               | Possible Values                                                                                      |
+| ---------------------- | ----------------------------------------------------------------------------------------------------- |
+| authenticationProvider | `OKTA_AUTHENTICATION_PROVIDER`, `ACTIVE_DIRECTORY`, `LDAP`, `FEDERATION`, `SOCIAL`, `FACTOR_PROVIDER` |
+| authenticationStep     | `0`, `1`                                                                                              |
+| credentialProvider     | `OKTA_CREDENTIAL_PROVIDER`, `RSA`, `SYMANTEC`, `GOOGLE`, `DUO`, `YUBIKEY`                             |
+| credentialType         | `OTP`, `SMS`, `PASSWORD`, `ASSERTION`, `IWA`, `EMAIL`, `OAUTH2`, `JWT`                                |
+|------------------------+-------------------------------------------------------------------------------------------------------|
+
+**NOTE:** as of now, **authenticationStep** is only set to `0` and serves no purpose.
```
bobtiernay-okta: > [...] and serves no purpose.

It seems odd to state this. Perhaps we should leave `authenticationStep` this out of the documentation completely.

@omgitstom thoughts?

ericzhou-okta: We certainly could take `authenticationStep` out of this documentation. I just figured that it would be easier to come back and delete one warning note rather than re-input what the `authenticationStep` does in a few weeks.

omgitstom: This is a defect that should be fixed.  Documenting around defects is an antipattern. If we can't get clarity on how the MFA team is going to use it, remove it.

Issue #1694, opened 2018-01-30
Investigate syncing the styles from docs with blog for tables and blockquotes


mraible:
The blockquotes in the docs are a lot tighter than they are in the blog. The tables are colored in blue too, whereas they're gray in the blog. 

Docs: https://developer.okta.com/docs/api/resources/users#response-parameters

<img width="760" alt="screen shot 2018-01-30 at 10 31 12 am" src="https://user-images.githubusercontent.com/17892/35581497-2c6eb5bc-05a9-11e8-9b89-a67f59f94ef1.png">

I recently removed the bottom border for tables on the blog so we could use tables to layout side-by-side images. We'll need to consider this if we add the doc's table styling to the blog.

nbarbettini:
I like this idea. The crazy padding on blog blockquotes has always annoyed me:

![image](https://user-images.githubusercontent.com/7525482/35581777-b545abaa-05a1-11e8-9d2a-fb70b419f66f.png)

👍 from me for consistent styles between the blog and docs.

mraible:
Here's what tables look like on the blog:

![image](https://user-images.githubusercontent.com/17892/35590250-7aa230e6-05c3-11e8-8a6e-112f23f393f6.png)

PR #1716, opened 2018-02-02
Change blockquote so it matches doc CSS

mraible:
Contributes to fixing #1694. Screenshot below.

image

This is what it looks like before this change:

image

### DIFF
--- a/_source/_assets/css/okta/components/_BlogPost.scss
+++ b/_source/_assets/css/okta/components/_BlogPost.scss
@@ -165,21 +165,19 @@
 			border-bottom: 0;
 		}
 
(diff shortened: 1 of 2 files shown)


jamieleerice-okta:
@mraible the style looks good, but I think we should standardize on the darker blue that was originally included in the blog style. We're moving away from the light blue in our palette. Is the text size that you're using the same size as the rest of the body copy in the post?

mraible:
@jamieleerice-okta I'll change it back to the darker blue. No, the text size is `font-size: .9em`. Here's a screenshot of what it looks like with the old blue and same font size.

screen shot 2018-02-02 at 12 42 17 pm

PR #1723, opened 2018-02-06
Fix Some SEO Stuffz

rdegges:
I fixed some SEO stuff we sorely needed to handle:

- Make meta description legit
- Make swiftype search legit
- Remove bad canonical linkage


rdegges:
@mraible will you review when it's done building on travis + merge?

PR #1724, opened 2018-02-06
Fix markdown for sub-lists in Spring Microservices blog post

mraible:
https://developer.okta.com/blog/2017/08/08/secure-spring-microservices


jmelberg-okta:
@mraible I noticed we have a number of pages with bulleted lists that are malformed due to spacing.  If we imposed a `lint` rule that would throw errors for this kind of thing, would that be a deterrent to contribute? You can see the closed PR [#1591](https://github.com/okta/okta.github.io/pull/1591)

mraible:
@jmelberg-okta I think it'd be good to catch this and warn the user about sub-lists not being created correctly, but it might be difficult to catch.

PR #1746, opened 2018-02-09
OKTA-156279: Detail Logs Transaction Object

ericzhou-okta:
## Description:
screen shot 2018-02-09 at 15 10 57

### Resolves:
* [OKTA-156279](https://oktainc.atlassian.net/browse/OKTA-156279)


omgitstom:
Looks good!  The only mention I could make is that description for Details is weak, but since the product doesn't currently use this field, we can let it slide.  When we have an idea on when the product will set this, we can make sure we describe it correctly

bobtiernay-okta:
Had the same comment as Tom, but I think what he suggests makes sense.

PR #1767, opened 2018-02-14
Update Docs for Issuer Object

guyzehavi-okta:
## Description:

Update documentation for the Issuer object.

Bacon: test
Resolves: OKTA-156282

### Resolves:
* [Issue #X](#X)

* [OKTA-124190](https://oktainc.atlassian.net/browse/OKTA-156282)

### DIFF
--- a/_source/_docs/api/resources/system_log.md
+++ b/_source/_docs/api/resources/system_log.md
@@ -427,13 +427,13 @@ Some of the fields listed above have a finite set of possible values.
 
 ### Issuer Object
 
-Describes an issuer in the authentication context.
+Describes the `issuer` of the authorization server when the authentication is performed via OAuth. This is the location where well-known resources regarding details of the authorization servers are published.
 
 |------------+----------------------------------------------------------------+-----------------+----------|
 | Property   | Description                                                    | DataType        | Nullable |
 | ---------- | -------------------------------------------------------------- | --------------- | -------- |
-| id         | An ID for the issuer                                           | String          | TRUE     |
-| type       | The type of the issuer                                         | String          | TRUE     |
+| id         | Varies depending on the type of authentication. If authentication is SAML 2.0, `id` is the issuer in the SAML assertion. For social login, `id` is the issuer of the token.                                         | String          | TRUE     |
+| type       | Information regarding `issuer` and source of the SAML assertion or token.                                         | String          | TRUE     |
 |------------+----------------------------------------------------------------+-----------------+----------|
 
 ### SecurityContext Object


guyzehavi-okta:
@ryanoblak-okta and @ericzhou-okta please review.

guyzehavi-okta:
@mystiberry-okta tagging you here per Eric's input.

_source/_docs/api/resources/system_log.md
```diff
@@ -427,13 +427,13 @@ Some of the fields listed above have a finite set of possible values.
...
 
-Describes an issuer in the authentication context.
+Describes the issuer in the OAuth context.
 
 |------------+----------------------------------------------------------------+-----------------+----------|
 | Property   | Description                                                    | DataType        | Nullable |
 | ---------- | -------------------------------------------------------------- | --------------- | -------- |
-| id         | An ID for the issuer                                           | String          | TRUE     |
-| type       | The type of the issuer                                         | String          | TRUE     |
+| id         | If the authentication was SAML 2.0, the issuer would be the issuer in the saml assertion. For social login, the issuer would be the issuer of the token.                                           | String          | TRUE     |
```
bobtiernay-okta: `saml` should be uppercased as it is an acronym.

bobtiernay-okta: > .. would be the issuer of the token

Perhaps adding something like:

> (e.g. Facebook)

Would be informative?

_source/_docs/api/resources/system_log.md
```diff
@@ -427,13 +427,13 @@ Some of the fields listed above have a finite set of possible values.
...
 ### Issuer Object
 
-Describes an issuer in the authentication context.
+Describes the issuer in the OAuth context.
 
 |------------+----------------------------------------------------------------+-----------------+----------|
 | Property   | Description                                                    | DataType        | Nullable |
 | ---------- | -------------------------------------------------------------- | --------------- | -------- |
-| id         | An ID for the issuer                                           | String          | TRUE     |
-| type       | The type of the issuer                                         | String          | TRUE     |
```
bobtiernay-okta: What happened to `type`?

_source/_docs/api/resources/system_log.md
bobtiernay-okta: Might want to consider linking to https://tools.ietf.org/html/draft-ietf-oauth-discovery-06#page-4

_source/_docs/api/resources/system_log.md
```diff
@@ -427,13 +427,13 @@ Some of the fields listed above have a finite set of possible values.
 
 ### Issuer Object
 
-Describes an issuer in the authentication context.
+Describes the issuer in the OAuth context.
```
bobtiernay-okta: Can we expound on this a bit using the content from https://tools.ietf.org/html/draft-ietf-oauth-discovery-06#page-4?

_source/_docs/api/resources/system_log.md
guyzehavi-okta: @bobtiernay-okta that is partially the issue; currently we are setting values like 'Facebook', 'Google', etc but Tom Abbot thinks we should be setting 'https://accounts.google.com/', etc. . Hence my reluctance to use examples.

_source/_docs/api/resources/system_log.md
guyzehavi-okta: @bobtiernay-okta type is still there. as you can see:
screen shot 2018-02-14 at 4 00 51 pm

_source/_docs/api/resources/system_log.md
guyzehavi-okta: Note that if we add 'The authorization server's issuer identifier, which is
      a URL that uses the "https" scheme and has no query or fragment
      components.' Then we need some way to enforce that, because currently it is beign set to values like 'Google', not 'https://accounts.google.com/'

_source/_docs/api/resources/system_log.md
guyzehavi-okta: Sure.

_source/_docs/api/resources/system_log.md
ryanoblak-okta: @guyzehavi-okta Just for my own edification— the authentication context is always an OAuth context?

_source/_docs/api/resources/system_log.md
guyzehavi-okta: Sure

_source/_docs/api/resources/system_log.md
guyzehavi-okta: I do not think so. Perhaps better phrasing it is 'Describes the issuer if the logic is in OAuth Context' ?

_source/_docs/api/resources/system_log.md
ryanoblak-okta: "the issuer would be the issuer of the token" indicates an issuer name, such as "Facebook".

If/when teams start using URIs, then we can rephrase the docs to say "the issuer is the URI that generated the token" or some such.

_source/_docs/api/resources/system_log.md
ericzhou-okta: How about something like "describes the `issuer` in the `authenticationContext` when the authentication is performed via OAuth"?

ericzhou-okta:
Hi @guyzehavi-okta. For consistency with the other sections, could you use noformat when referring to the issuer?

issuer -> `issuer`

_source/_docs/api/resources/system_log.md
ryanoblak-okta: Do what happens if the event didn't occur in an OAuth Context?

Maybe "Describe the `issuer` if the event occurred in an OAuth Context; otherwise (insert info here)."?

ericzhou-okta: I think that if the event is not in an OAuth context, the issuer is blank.

guyzehavi-okta: @ericzhou-okta  is correct

_source/_docs/api/resources/system_log.md
```diff
@@ -427,13 +427,13 @@ Some of the fields listed above have a finite set of possible values.
...
 
-Describes an issuer in the authentication context.
+Describes the `issuer` of the authorization server when the authentication is performed via OAuth. This is the location where well known resources regarding details of the authorization servers are published.
 
 |------------+----------------------------------------------------------------+-----------------+----------|
 | Property   | Description                                                    | DataType        | Nullable |
 | ---------- | -------------------------------------------------------------- | --------------- | -------- |
-| id         | An ID for the issuer                                           | String          | TRUE     |
-| type       | The type of the issuer                                         | String          | TRUE     |
+| id         | If the authentication was SAML 2.0, the `issuer` would be the issuer in the SAML assertion. For social login, the `issuer` would be the issuer of the token. See https://tools.ietf.org/html/draft-ietf-oauth-discovery-06#page-4 for full spec.                                         | String          | TRUE     |
```
guyzehavi-okta: note: not sure if do do `issuer` for 'issuer of the token' and 'issuer in the SAML assertion'. There issuer is not being used as the object being defined but as a reference to whom created the token (in 'issuer of the token'); similar with 'issuer in the SAML assertion' - it is the one who asserts not the object in question. Thoughts?

_source/_docs/api/resources/system_log.md
```diff
@@ -427,13 +427,13 @@ Some of the fields listed above have a finite set of possible values.
...
+| type       | Information regarding `issuer` and source.                                         | String          | TRUE     |
```
bobtiernay-okta: Can we do a little better than the pretty vague "information"?

_source/_docs/api/resources/system_log.md
bobtiernay-okta: I would make that a hyperlink with friendly link text if you are going to include it.

_source/_docs/api/resources/system_log.md
guyzehavi-okta: @bobtiernay-okta the issue is that not even Tom Abbot is sure what this is; to be best of what we were able to gather, it could be a categorization of the issuer. For instance:
Social.Google
Social.Facebook
Federation.Google
^^^ from conversation with Tom Abbot.
However, currently it is only set to '"Certificate Authority"' in one of the event builders ... so not entirely sure where to go.
See:
screen shot 2018-02-15 at 8 46 01 am

_source/_docs/api/resources/system_log.md
```diff
@@ -427,13 +427,13 @@ Some of the fields listed above have a finite set of possible values.
 
 ### Issuer Object
 
-Describes an issuer in the authentication context.
+Describes the `issuer` of the authorization server when the authentication is performed via OAuth. This is the location where well known resources regarding details of the authorization servers are published.
```
ryanoblak-okta: Nit: "well known" -> "well-known" for grammatical correctness (see http://www.businesswritingblog.com/business_writing/2009/07/a-wellknown-problem-well-known-.html)

_source/_docs/api/resources/system_log.md
ryanoblak-okta: I'm not sure what the IETF link adds here. The IETF spec defines `issuer` as:

> The authorization server's issuer identifier, which is a URL that uses the "https" scheme and has no query or fragment components.

As Guy mentioned, this field's value is not currently a URL in generated events (though we want to move in that direction). So linking to the OAuth spec seems to confuse things.

I'm also wondering if we can avoid the past tense conditional phrasing. Maybe something like the following?

> Varies depending on the type of authentication. If authentication is SAML 2.0, `id` is the issuer in the SAML assertion. For social login, `id` is the issuer of the token.

_source/_docs/api/resources/system_log.md
ryanoblak-okta: Should "source" be "source of the SAML assertion or token"?

_source/_docs/api/resources/system_log.md
guyzehavi-okta: @ryanoblak-okta just mad the change you suggested.

_source/_docs/api/resources/system_log.md
guyzehavi-okta: sure

_source/_docs/api/resources/system_log.md
```diff
@@ -427,13 +427,13 @@ Some of the fields listed above have a finite set of possible values.
...
 
-Describes an issuer in the authentication context.
+Describes the `issuer` of the authorization server when the authentication is performed via OAuth. This is the location where well-known resources regarding details of the authorization servers are published.
 
 |------------+----------------------------------------------------------------+-----------------+----------|
 | Property   | Description                                                    | DataType        | Nullable |
 | ---------- | -------------------------------------------------------------- | --------------- | -------- |
-| id         | An ID for the issuer                                           | String          | TRUE     |
-| type       | The type of the issuer                                         | String          | TRUE     |
+| id         | Varies depending on the type of authentication. If authentication is SAML 2.0, id is the issuer in the SAML assertion. For social login, id is the issuer of the token.                                         | String          | TRUE     |
```
ryanoblak-okta: Can highlight id? `id` rather than id.

mystiberry-okta:
@guyzehavi-okta The problem has been resolved, if you pull source and rebase, this should pass tests.

guyzehavi-okta:
thank you @mystiberry-okta !

PR #1784, opened 2018-02-20
:fire: Remove old CSS and image files

jmelberg-okta:
## Description:
- Removes a handful of old CSS files from previous UI, along with images no longer used.


jmelberg-okta:
@robertjd Want to take one last look before we 🚢?

robertjd:
Still looks good, let's merge this tomorrow morning

PR #1788, opened 2018-02-23
fix fragile links

mystiberry-okta:
## Description:
- A link in oan faq was full on broken--I decided to fix the rate limit links while I was at it, and a few links to support.okta.com that should be to help.okta.com. The old rate limit links sent customers in a bit of a round trip, I fixed it so they go straight to the content. Possibly unnecessary since most changes were to old RN, but you know never what will show up in search results, so...

### Resolves:
[OKTA-159882](https://oktainc.atlassian.net/browse/OKTA-159882)


_source/_change-log/2016-27.md
```diff
@@ -23,8 +23,7 @@ To improve usability, we've moved some of the panels in the Okta user interface
 <!-- OKTA-91831, OKTA-93759 -->
 The Okta Sign-In Widget will be updated to version 1.4.0 for Production orgs.
 
-See [Okta Sign-In Widget](/docs/guides/okta_sign-in_widget) for updated sample code
-after [Release 2016.27](https://support.okta.com/help/articles/Knowledge_Article/Current-Release-Status) is rolled out to production orgs.
+See [Okta Sign-In Widget](/docs/guides/okta_sign-in_widget) for updated sample code.
```
mystiberry-okta: We don't need the blah blah around release timing, .27 is long gone.

_source/_docs/api/resources/schemas.md
```diff
@@ -6,7 +6,7 @@ excerpt: The Schemas API defines custom user profiles for Okta users and applica
 
 # Schemas API
 
-Okta's [Universal Directory](https://support.okta.com/articles/Knowledge_Article/About-Universal-Directory) allows administrators to define custom user profiles for Okta users and applications.  Okta has adopted a subset [JSON Schema Draft 4](https://tools.ietf.org/html/draft-zyp-json-schema-04) as the schema language to describe and validate extensible user profiles. [JSON Schema](http://json-schema.org/) is a lightweight declarative format for describing the structure, constraints, and validation of JSON documents.
+Okta&#8217;s [Universal Directory](https://help.okta.com/en/prod/Content/Topics/Directory/About_Universal_Directory.htm) allows administrators to define custom user profiles for Okta users and applications.  Okta has adopted a subset [JSON Schema Draft 4](https://tools.ietf.org/html/draft-zyp-json-schema-04) as the schema language to describe and validate extensible user profiles. [JSON Schema](http://json-schema.org/) is a lightweight declarative format for describing the structure, constraints, and validation of JSON documents.
```
jmelberg-okta: Why are we replacing the `'` with `$#8217;`? I'm not seeing a difference when this page is rendered.

mystiberry-okta: So at some point it was $#8217; and then got changed to ', I'm just changing it back. We removed all the single apostrophes months ago because they were causing trouble in github rendering (and maybe also in some other way I've forgotten). If that's no longer an issue I can stop replacing them. I just remember once we had a horrible tangle of red in the topics that people were reviewing, and that was the fix.

mystiberry-okta: Leaving this for now, but following up on the issue.

PR #1789, opened 2018-02-23
Syslog2: Event ID Correlation

ericzhou-okta:
## Description:
- Creates a new section in the syslog docs that describes the relationships between the various ID fields.

### Resolves:
* [OKTA-127979](https://oktainc.atlassian.net/browse/OKTA-127979)

### DIFF
--- a/_source/_docs/api/resources/system_log.md
+++ b/_source/_docs/api/resources/system_log.md
@@ -528,6 +528,35 @@ The following sections outline the major event types captured by the system log.
 * `user.authentication.sso` doesn't capture whether the SSO attempt was successful or failed because Okta can't collect the subsequent authentication attempt status from the third-party service.
 
 
+## Event ID Correlation
+
+Throughout a user's session many requests (transactions) can occur, such as logging into Okta or opening an application. Within a given request many events may be logged. To illustrate this principle, the table below shows 18 events produced from 13 transactions over 6 different sessions, all performed by one user.
+
+| External Session ID       | Transaction ID              | Event ID                             | Event Type                                  | Display Message                   |
+|:--------------------------|:----------------------------|:-------------------------------------|:--------------------------------------------|:----------------------------------|
+| trs5JnlvlaIQTOqOj9imLy7lA | WcKPxq1f8QLfFvv3UPHhhgAACGM | f24790d0-d324-47f8-aac5-c27a31ab928d | user.session.access_admin_app               | User accessing Okta admin app     |
+|                           | WcKPxq1f8QLfFvv3UPHhhgAACGM | ed317758-8776-4240-a540-277c44dcb408 | application.lifecycle.update                | Update application                |
+|                           |                             | 421c1551-71b2-4ebe-a70d-b5f7d3698429 | application.lifecycle.update                | Update application                |
+|                           |                             | 06a50bbe-44fd-40de-83e6-2e4cc2a17d16 | application.lifecycle.update                | Update application                |
+| trsUz2TG3wKS6ar1lvWzHo71w | Wij-6q4YuniRd9yTmWHpfwAAADc | 3e240ff4-6af7-47f2-b107-a2ef661ffc01 | application.user_membership.change_username | Change users application username |
+|                           |                             | 572b05e9-b6be-4dfe-8bc3-01bb3a5a1af5 | application.user_membership.add            | Add user to application membership |
+|                           |                             | 30f29bbf-3218-429b-827a-0a93809591db | application.user_membership.remove        | Remove users application membership |
+|                           | Wij-964YuniRd9yTmWHu1AAAAEA | 5f98d062-05a9-4ede-89a0-8a2ce27efdd4 | user.session.access_admin_app               | User accessing Okta admin app     |
+|                           | Wij-95eCbHF7In2MKNavlgAAD9I | 45f71ac2-e8b2-4c19-b4cc-d2560108c889 | application.lifecycle.update                | Update application                |
+|                           |                             | 46b85d65-01c6-44d2-86d2-25704804b1c5 | application.lifecycle.update                | Update application                |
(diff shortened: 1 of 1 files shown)


jakubvul:
@ericzhou-okta is this related to a weekly release, or simply an improvement of the existing documentation?

ericzhou-okta:
@jakubvul-okta I'm not familiar with what qualifies as a weekly release. Could you explain how to decide what branch should be merged into for new content?

ericzhou-okta:
@mystiberry-okta My change is producing some awkward formatting with the ToC overlapping with my table. Do you have any advice for fixing this?

screen shot 2018-02-23 at 13 45 29

jakubvul:
@ericzhou-okta if we're updating documentation because of code that's going out as part of a release, then we merge into the `weekly` branch. Basically if you're documenting something that's in the Release Notes, that should be targeted at `weekly`.

If you're just updating documentation, and it's not a change that's in the Release Notes, then that's targeted to `source`. 

This way we can get improvements out ASAP without waiting unnecessarily for the next release.

jakubvul:
@ericzhou-okta and with regards to the overflow, add the following immediately after your table:

`{:.table .table-word-break}`

For example:

```
| *null*                    | Wm@-R2s5lEMbNIB03krtvAAACyo | 566671be-ec0b-400d-ad2e-6fc73ed12fb1 | user.session.start                          | User login to Okta                | FAILURE        |
{:.table .table-word-break}
```

ericzhou-okta:
@jakubvul gotcha, thanks! In that case this would be going into `source`.

_source/_docs/api/resources/system_log.md
```diff
@@ -528,6 +528,34 @@ The following sections outline the major event types captured by the system log.
 * `user.authentication.sso` doesn't capture whether the SSO attempt was successful or failed because Okta can't collect the subsequent authentication attempt status from the third-party service.
 
 
+## Event ID Correlation
+
+Throughout a users session many requests (transactions) can occur, such as logging into Okta or opening an application. Within a given request many events may be logged. To illustrate this principle, the table below shows 18 events produced from 13 transactions over 6 different sessions, all performed by one user.
```
mystiberry-okta: Should have an apostrophe for users: "Throughout a user's session"

ericzhou-okta: woops. Thanks for catching that!

ericzhou-okta:
@jakubvul-okta I tried the table word break option. The table is fitting now, but the `Outcome Result` column just looks bizarre with the random "S" dangling off the second row. Is there a way to stop applying the word break to a specific column?

screen shot 2018-02-23 at 14 20 06

jakubvul:
@ericzhou-okta , no, sorry. Maybe it's better to include this as an image?

jakubvul:
The only other thing that I can think of is that you could manually put in newlines with `<br>` into the cell until the table fits the way you want it to.

To be honest I'm not sure what behavior I would expect with a table of this size.

_source/_docs/api/resources/system_log.md
```diff
@@ -528,6 +528,32 @@ The following sections outline the major event types captured by the system log.
 * `user.authentication.sso` doesn't capture whether the SSO attempt was successful or failed because Okta can't collect the subsequent authentication attempt status from the third-party service.
 
 
+## Event ID Correlation
+
+Throughout a user's session many requests (transactions) can occur, such as logging into Okta or opening an application. Within a given request many events may be logged. To illustrate this principle, the table below shows 18 events produced from 13 transactions over 6 different sessions, all performed by one user.
+
+| External Session ID       | Transaction ID              | Event ID                             | Event Type                                  | Display Message                   | Outcome<span style="color:#007dc1">_</span> Result | <!---span is there as a hack to get formatting to work as intended-->
```
ericzhou-okta: Hey @jakubvul-okta , what's your guys' tolerance of crazy formatting hacks?

ericzhou-okta: screen shot 2018-02-23 at 15 46 48

ryanoblak-okta: @ericzhou-okta What are your thoughts on just dropping the `Outcome Result` column altogether? It doesn't seem to add much to the purpose of the section.

ericzhou-okta: @ryanoblak-okta I'm concerned that doing so would obfuscate why the `External Session ID` is `null` in the last row. With the `Outcome Result` column, readers can clearly see that the lack of a session ID stems from a login failure. I find this beneficial because some reader might implicitly assume an "at least one" relationship between `Transaction ID`s or `Event ID`s and `External Session ID`s. 

However, I don't know if pointing out this relationship is even that important. If you or @bobtiernay-okta think otherwise, I'm content with dropping the `Outcome Result` column.

ryanoblak-okta: Ah, I see. That's a good point @ericzhou-okta. Maybe instead of having the `Outcome Result` column we can just add a parenthetical note next to the `null` in the last row to explain why it's `null`? This would save horizontal space while making it clear why `Session ID` is `null` for that row.

ericzhou-okta: I'm cool with that.

jakubvul: @ericzhou-okta I think we discourage "crazy formatting hacks" whenever possible.

ericzhou-okta: @jakubvul understandable.

ericzhou-okta:
@ryanoblak-okta here's what it looks like after deleting the `Outcome Result` column. Does this table and the explanation at the bottom look good to you?

screen shot 2018-02-26 at 12 55 08
screen shot 2018-02-26 at 12 55 24

ryanoblak-okta:
@ericzhou-okta Looks great. Nice explanation! 👍

PR #1797, opened 2018-02-27
SAML Vulnerability Blog Post

rdegges:
Needs to go live ASAP

### DIFF
--- a/_source/_posts/2018-02-27-a-breakdown-of-the-new-saml-authentication-bypass-vulnerability.md
+++ b/_source/_posts/2018-02-27-a-breakdown-of-the-new-saml-authentication-bypass-vulnerability.md
@@ -5,7 +5,7 @@ author: rdegges
 description: "An in-depth look at the new SAML authentication bypass vulnerability: what it is, how it works, and how you can protect yourself against it."
 tags: [security, research, programming]
 tweets:
-    - "Duno found a severe new SAML XML vulnerability. Check out @rdegges' write up on the subject:"
+    - "Duo found a severe new SAML XML vulnerability. Check out @rdegges' write up on the subject:"
     - "Are you using SAML (single sign-on) in your applications? Read @rdegges' article about the new SAML authentication bypass vulnerability and make sure you're safe:"
 ---
 
--- a/_source/_posts/2018-02-27-a-breakdown-of-the-new-saml-authentication-bypass-vulnerability.md
+++ b/_source/_posts/2018-02-27-a-breakdown-of-the-new-saml-authentication-bypass-vulnerability.md
@@ -1,30 +1,30 @@
 ---
 layout: blog_post
-title: "A Breakdown of Malcolm and What You Need to Know"
+title: "A Breakdown of the New SAML Authentication Bypass Vulnerability"
 author: rdegges
-description: "An in-depth look at the new Malcolm SAML vulnerability: what it is, how it works, and how you can protect yourself against it."
+description: "An in-depth look at the new SAML authentication bypass vulnerability: what it is, how it works, and how you can protect yourself against it."
 tags: [security, research, programming]
 tweets:
-    - "The new Malcolm SAML vulnerability is here. Check out @rdegges' write up on the subject:"
-    - "Are you using SAML (single sign-on) in your applications? Read @rdegges' article about the new Malcolm SAML vulnerability and make sure you're safe:"
+    - "Duno found a severe new SAML XML vulnerability. Check out @rdegges' write up on the subject:"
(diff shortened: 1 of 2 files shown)


_source/_posts/2018-02-27-a-breakdown-of-the-new-saml-authentication-bypass-vulnerability.md
```diff
@@ -0,0 +1,178 @@
...
+If you're not familiar with SAML (short for Security Assertion Markup Language), it's an open standard that allows users to share credentials between multiple web apps so they don't need to manually log in when accessing different web services. Many vendors use SAML to handle user authentication and authorization so their users can access web applications without needing individual credentials for each.
+
+The new SAML vulnerability allows an attacker to *bypass authentication* and directly assume the role of an authenticated user as part of the SAML flow. **This is a BIG DEAL.**
+
+## How the new SAML Authentication Bypass Vulnerability Works
+
+When a user is authenticating to a website using SAML, there are always three parties involved:
+
+- A user in a web browser
+- A service provider running a website that user is trying to access (EG: Salesforce)
```
mraible: Grammarly says “e.g.,” instead of “EG:”

_source/_posts/2018-02-27-a-breakdown-of-the-new-saml-authentication-bypass-vulnerability.md
```diff
@@ -0,0 +1,178 @@
...
+
+The XML document above would have the *exact same cryptographic signature* as the XML document below:
+
+```xml
+<p>hi</p>
+```
+
+This is because XML doesn't care about whitespace. When the XML document is analyzed before a signature is created, the space is removed.
+
+What this new vulnerability takes advantage of, specifically, is the fact that most canonicalization algorithms *also* don't care about comments. This means that the following two XML documents (while vastly different in parseability as noted in the previous section) will have *identical* cryptographic signatures: thereby allowing an attacker to sneak a comment into a SAML assertion and bypass authentication without ever raising any red flags in signature checks :(
```
mraible: Period at the end of the sentence.

_source/_posts/2018-02-27-a-breakdown-of-the-new-saml-authentication-bypass-vulnerability.md
```diff
@@ -0,0 +1,178 @@
...
+<saml:NameID>not-an-admin@okta.com</saml:NameID>
+```
+
+While it is theoretically possible for all SAML vendors to only allow a specific canonicalization algorithm which DOES leave XML documents unmodified (so that they contain things like whitespace, comments, etc.), in practice this would be hard to enforce across vendors and implementations.
+
+To read more about XML canonicalization issues you might want to check out the [Wikipedia article](https://en.wikipedia.org/wiki/XML_Signature#XML_canonicalization) on the topic.
+
+## How to Protect Yourself from the New SAML Authentication Bypass Vulnerability
+
+If your company is using SAML you will want to *check with your SAML vendors* to ensure they are not susceptible to this new vulnerability. [Okta](https://developer.okta.com/), by the way, **is not vulnerable** >:)
```
mraible: Add punctuation at the end.

_source/_posts/2018-02-27-a-breakdown-of-the-new-saml-authentication-bypass-vulnerability.md
```diff
@@ -0,0 +1,178 @@
...
+
+If you're a developer building systems to integrate with SAML vendors or single sign-on providers, try to NOT integrate using SAML if possible. Instead: use a more modern identity standard such as OpenID Connect.
+
+OpenID Connect is a lot simpler than SAML, doesn't use XML or [XMLDSIG](https://www.w3.org/TR/xmldsig-core1/), and is far less susceptible to attacks like these since it doesn't rely on XML parsers. There's a long, complex history with XML and SAML -- many of the SAML vulnerabilities that have been found over the last ~15 years are centered around XML parsing (just like this one).
+
+Take, for instance, [XML Signature Wrapping](https://www.youtube.com/watch?v=QLKM4USUlZs) attacks which have been known about since ~2002 but are still finding their way into modern SAML implementations as recently as 2014.
+
+If possible, avoid SAML and use [OpenID Connect](https://developer.okta.com/blog/2017/07/25/oidc-primer-part-1) whenever you can. If you must use SAML, check your SAML library (and its underlying XML parsing/validation/signature libraries) to ensure it is NOT vulnerable to the new attack.
```
mraible: URL should be relative

PR #1798, opened 2018-02-27
Fix SAML post typos

mraible:


_source/_posts/2018-02-27-a-breakdown-of-the-new-saml-authentication-bypass-vulnerability.md
```diff
@@ -109,7 +109,7 @@ This is **bad news** for web security as a simple XML comment can cause a SAML i
 
 The second issue the new SAML vulnerability takes advantage of is the relatively "weak" protection provided by XMLDSIG's canonicalization algorithms that are used to create and validate XML signatures.
 
-Typically, when a SAML assertion is created, the assertion element itself is cryptographically signed. This helps the service provider (EG: Salesforce) trust that the user they're about to log in is valid. Unfortuantely however, there are several different types of canonicalization algorithms that are allowed to be used when XML signatures are created, and most of them are not well-suited for creating tamper-proof data.
+Typically, when a SAML assertion is created, the assertion element itself is cryptographically signed. This helps the service provider (e.g., Salesforce) trust that the user they're about to log in is valid. Unfortunately, however, there are several different types of canonicalization algorithms that are allowed to be used when XML signatures are created, and most of them are not well-suited for creating tamper-proof data.
```
rdegges: Can you remove this one? It's apparently supposed to be the other way :(

_source/_posts/2018-02-27-a-breakdown-of-the-new-saml-authentication-bypass-vulnerability.md
```diff
@@ -137,21 +137,21 @@ and
 <saml:NameID>not-an-admin@okta.com</saml:NameID>
 ```
 
-While it is theoretically possible for all SAML vendors to only allow a specific canonicalization algorithm which DOES leave XML documents unmodified (so that they contain things like whitespace, comments, etc.), in practice this would be hard to enforce across vendors and implementations.
+While it is theoretically possible for all SAML vendors only to allow a specific canonicalization algorithm which DOES leave XML documents unmodified (so that they contain things like whitespace, comments, etc.), in practice this would be hard to enforce across vendors and implementations.
```
rdegges: Can you switch this one back? I think it makes more sense the other way around.

PR #1873, opened 2018-03-16
Single-source contact Support and Contact Support with link

mystiberry-okta:
## Description:
- We put the link Support suggested with every instance of "contact Support" or "Contact Support" as a single-source param so it can easily be written consistently and changed if the preferred method of contact changes.

### Resolves:

[OKTA-160463](https://oktainc.atlassian.net/browse/OKTA-160463)


mystiberry-okta:
@jakubvul-okta optional review
@jmelberg-okta since there's a bit of code in here, please review.

_config.yml
```diff
@@ -121,3 +121,8 @@ dev_forum:  https://devforum.okta.com/
 
 # Location of main Drupal site for navigation and shared resources
 external_domain: https://developer.okta.com
+
+# Link for telling people to contact support
+
+contact_support_lc: <a href="https://support.okta.com/help/open_case" target="_blank">contact Support</a>
```
jmelberg-okta: Do you want `contact` to be included in the link? IMO it looks better as text.

screen shot 2018-03-18 at 9 00 28 pm
screen shot 2018-03-18 at 9 00 05 pm

jakubvul:
Agree that it would be better to just have "Support" as the link text.

_config.yml
mystiberry-okta: Ah, I thought it would be a clearer call to action w/the whole verb phrase. Will change.

mystiberry-okta: Agree that we should doc the variables...not sure where, will follow up.

_config.yml
```diff
@@ -121,3 +121,8 @@ dev_forum:  https://devforum.okta.com/
...
+contact_support_uc: <a href="https://support.okta.com/help/open_case" target="_blank">Contact Support</a>
```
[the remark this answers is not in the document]
mystiberry-okta: ack

PR #1878, opened 2018-03-20
Fix Docs for Client id Field

guyzehavi-okta:
## Description:

Fix the docs for the Client id field.

Bacon: test
Resolves: OKTA-162578

### Resolves:
* [Issue #X](#X)

* [OKTA-162578](https://oktainc.atlassian.net/browse/OKTA-162578)

### DIFF
--- a/_source/_docs/api/resources/system_log.md
+++ b/_source/_docs/api/resources/system_log.md
@@ -271,6 +271,7 @@ When an event is triggered by an HTTP request, the `client` object describes the
 |------------+--------------------------------------------------------------------------------------------------------------------+-----------------+----------|
 | Property   | Description                                                                                                        | DataType        | Nullable |
 | ---------- | ------------------------------------------------------------------------------------------------------------------ | --------------- | -------- |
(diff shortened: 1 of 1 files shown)


_source/_docs/api/resources/system_log.md
```diff
@@ -271,6 +271,7 @@ When an event is triggered by an HTTP request, the `client` object describes the
 |------------+--------------------------------------------------------------------------------------------------------------------+-----------------+----------|
 | Property   | Description                                                                                                        | DataType        | Nullable |
 | ---------- | ------------------------------------------------------------------------------------------------------------------ | --------------- | -------- |
+| id  | For OAuth requests this is the id of the OAuth [client](https://tools.ietf.org/html/rfc6749#section-1.1) making the request. For SSWS token requests, this is the id of the agent making the request. | [UserAgent Object](#useragent-object) | TRUE |
```
ericzhou-okta: Are these the only two usages of this field?

ericzhou-okta: Are these the only two usages of this field?

guyzehavi-okta: Correct; there is the Oauth case and the SSWS agent case (or at least those were the cases that Trevor asked for)

PR #1902, opened 2018-03-26
Update getting_a_token.md

gabrielsroka:
Devs don't have an Admin button (but Classic UI does)

### DIFF
--- a/_source/_docs/api/getting_started/getting_a_token.md
+++ b/_source/_docs/api/getting_started/getting_a_token.md
@@ -14,18 +14,15 @@ redirect_from: "/docs/getting_started/getting_a_token.html"
 	If you don't have an Okta organization, you can create a free Okta
     Developer Edition organization [at this link](https://developer.okta.com/signup/){:target="_blank"}.
 
-2.  Click on the blue **Admin** button.
-    {% img okta-admin-ui-button-admin.png alt:"Admin" %}
-
-3.  Click on **Tokens** in the API menu.
+2.  Click on **Tokens** in the API menu.
 	{% img okta-admin-api-token-dev.png alt:"API" %}
 
-4.  Click on the **Create token** button.
+3.  Click on the **Create token** button.
 	{% img okta-create-api-token-button.png alt:"Create Token" %}
 
-5.  Name your token and click **Create token**.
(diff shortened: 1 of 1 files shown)


jmelberg-okta:
Hey @gabrielsroka! For most cases this is true - however users will still need to navigate using the **Admin** button if they have bookmarked apps (Dev Console or Classic UI). Want to update this PR to leave a note that the **Admin** button may/may not be visible?

mystiberry-okta:
@nbarbettini This is a place where we should probably doc both classic UI and Dev Console? The paths are quite different (this rewrite gets close but isn't accurate anymore for classic UI). Dev Console isn't a full GA feature yet, is it? I think admins do have to create API tokens for a few tasks (not sure)...?

nbarbettini:
@gabrielsroka @mystiberry-okta How about a note under step 2? Something like

> If you don't see the API menu, you might be using the Classic UI. Click the Admin button in the top-right corner (if it's visible) and choose API from the Security menu.

mystiberry-okta:
It would be simpler and clearer to the customer to simply doc both paths.

Mysti
On Apr 7, 2018, at 3:09 PM, Nate Barbettini <notifications@github.com<mailto:notifications@github.com>> wrote:


@gabrielsroka<https://github.com/gabrielsroka> @mystiberry-okta<https://github.com/mystiberry-okta> How about a note under step 2? Something like

If you don't see the API menu, you might be using the Classic UI. Click the Admin button in the top-right corner (if it's visible) and choose API from the Security menu.

mystiberry-okta:
Instead of the belabored steps 2-4, more like:

2. Create a token:
     a. If you use Developer Console, navigate to {path}, then choose **Create Token.**
     b. If you use the administrators UI, navigate to {path}, then choose **Create Token.**
    Write down the token value, as you won't be able to see it again.
 (screenshot of value being displayed) 

This accomplishes a couple of things: 
* Reduces clutter (there's nothing confusing about this process, it needn't be belabored).
* Introduces customers to the term "administrators UI" which is used throughout docs.
* Introduces customers to the two different UIs

However we decide to do it, we should be doing it everywhere there are path differences between the two UIs until there is functional parity, when we can just say "use the Dev Console".

PR #1911, opened 2018-03-27
Update zones.md

gabrielsroka:
1. according to https://help.okta.com/en/prod/Content/Topics/Security/Security_Network.htm , it's 5,000 gateways or proxies for a legacy zone. i verified that it works. so, it's 150 for a regular (`system = false`) zone, and 5,000 for a legacy (`system = true`). this should probably we wordsmithed better (and so should this sentence)
2. the `system` attribute in this document isn't documented
3. the Update an IP Zone Request Example is way too verbose. no need to include all those parameters. just `name` and `type` are required.

## Description:

### Resolves:
* [Issue #X](#X)

### DIFF
--- a/_source/_docs/api/resources/zones.md
+++ b/_source/_docs/api/resources/zones.md
@@ -55,7 +55,7 @@ curl -X POST
       "value": "3.3.4.5/24"
     }
   ]
-}' "https://{yourOktaDomain}.com/api/v1/zones/"
+}' "https://{yourOktaDomain}.com/api/v1/zones"
 ~~~
 
 #### Successful Response Example
@@ -551,12 +551,7 @@ curl -X PUT
 -d '
 {
   "type": "IP",
-  "id": "nzovw2rFz2YoqmvwZ0g3",
   "name": "UpdatedNetZone",
-  "status": "ACTIVE",
-  "created": "2017-01-24T19:53:28.000Z",
-  "lastUpdated": "2017-01-24T19:53:28.000Z",
-  "system": false,
   "gateways": [
     {
       "type": "CIDR",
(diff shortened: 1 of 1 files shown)


mystiberry-okta:
@mpeng-okta  or @lilyhsiao-okta can one of you review for technical accuracy? Also, a question is raised about listing all the parameters--if they are valid we want to list them, but also indicate that only two are required.  Thanks!

mpeng-okta:
I think someone from the Adaptive team might be more up-to-date on/familiar with the Zones API.  @rpamidimarri-okta could you check this out?

_source/_docs/api/resources/zones.md
```diff
@@ -699,8 +674,8 @@ An IP zone defines several attributes:
 |type          | Type of zone (currently it can only be IP)                                                | String                                      |   Yes         |     N/A       |
 |id            | Unique identifier for this zone                                                           | String                                      | No (Assigned) |     N/A       |
 |name          | Unique name for this zone                                                                 | String                                      |   Yes         | 128 (chars)   |
-|gateways      | IP addresses (range or CIDR form) of this zone                                            | Array of [Address Objects](#address-object) |   No          | 150 (entries) |
-|proxies       | IP addresses (range or CIDR form) allowed to forward request from gateway addresses above | Array of [Address Objects](#address-object) |   No          | 150 (entries) |
+|gateways      | IP addresses (range or CIDR form) of this zone                                            | Array of [Address Objects](#address-object) |   No          | 150 or 5,000 (entries) |
```
rpamidimarri-okta: I dont see this in https://help.okta.com/en/prod/Content/Topics/Security/Security_Network.htm
But, you are right that for legacy ip zone, the limit is 5000 for gateways & proxies.

gabrielsroka:
https://help.okta.com/en/prod/Content/Topics/Security/Security_Network.htm says
"You can define a maximum of 5000 legacy network zones"

rpamidimarri-okta:
> https://help.okta.com/en/prod/Content/Topics/Security/Security_Network.htm says
"You can define a maximum of 5000 legacy network zones"

Just saw that - that is inaccurate, we should fix that. There can be only one legacy zone. The number of proxies/gateways in that legacy zone could be 5000.

_source/_docs/api/resources/zones.md
mystiberry-okta: @rpamidimarri-okta , how can we add the system property? I can't really merge this until we also explain system, since the gateways values depend on the value of system. Can you provide the info?

PR #1920, opened 2018-03-31
Fix docs for importing password hashes

nbarbettini:
Fixes a few issues that customers found with our documentation around importing password hashes:

* The user won't be able to log in immediately - they will need to be activated
* If you specify `activate=false` (as in the request example), the user will be `STAGED`, not `ACTIVE`
* Clarified that you can only use password hashes in an update operation if the user is `STAGED`. In other words, you cannot update (PUT or POST) a password hash object on an `ACTIVE` user.


nbarbettini:
Thanks for reviewing @mystiberry-okta 😄 Still looking for a review from @edjohns-okta to make sure this is technically accurate.

PR #1928, opened 2018-04-05
fix broken link external

mystiberry-okta:
## Description:
- An external link broke. Replaced with a slightly less useful link because I couldn't find the content previously linked to, then moved/removed by Apple.


mystiberry-okta:
now link checker is happy, I'll close this.

PR #1931, opened 2018-04-05
Replace CSS color values with variables

alexdahl-okta:
## Description:
- Sweeps current sass files and replaces hex values with named variables.
- In `_source/_assets/css/okta/pages/_quickstarts.scss`, removes a declaration that is no longer used.

There should be no visual or user-facing changes, and I've sanity-checked to QA as much as I can.

Does not touch legacy files for safety. After this change, we should no longer need to introduce one-off or hard-coded color values.

Color names are programmatically generated and are universally unique with our design system. cc @edburyenegren-okta 

### Resolves:
* [UX-1083](https://oktainc.atlassian.net/browse/UX-1083)

### DIFF
--- a/_source/_assets/css/okta/base/_variables.scss
+++ b/_source/_assets/css/okta/base/_variables.scss
@@ -28,7 +28,7 @@ $font-size:     (
                 'h3':               28px,
                 'h2':               32px,
                 'h1':               36px,
-                
+
                 'new-small':        12px,
                 'new-h2':           28px,
                 'new-h1':           44px,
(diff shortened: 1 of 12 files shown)


_source/_assets/css/okta/base/_variables.scss
```diff
@@ -81,64 +81,137 @@ $line-height:   (
 
 
 // Project/Client Colors
+
+// Abstract definitions:
+// (These are consistent with Nim. Changes here should be mirrored there - https://github.com/okta/courage/blob/master/assets/sass/nim/abstracts/_colors.scss)
```
edburyenegren-okta: 🙏

alexdahl-okta: 🙇

PR #1939, opened 2018-04-10
fix onSuccess in Login.js code snippet in okta_react_sign-in_widget

mschaeffner:
## Description:
Fixes the `onSuccess`function in `Login.js` code snippet in the the guide [Adding a custom login page to your React application](https://developer.okta.com/code/react/okta_react_sign-in_widget)

Handles the case if a user resets the password. Right now this results in following error:
```
Uncaught TypeError: Cannot read property 'token' of undefined
```

### DIFF
--- a/_source/_code/react/okta_react_sign-in_widget.md
+++ b/_source/_code/react/okta_react_sign-in_widget.md
@@ -122,7 +122,7 @@ export default withAuth(class Home extends Component {
     const button = this.state.authenticated ?
       <button onClick={this.props.auth.logout}>Logout</button> :
       <button onClick={this.props.auth.login}>Login</button>;
-    
+
     return (
       <div>
         <Link to='/'>Home</Link><br/>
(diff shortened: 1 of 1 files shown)


mschaeffner:
I cannot see where the guide has been rewritten. The concerning file `/_source/_code/react/okta_react_sign-in_widget.md` was last modified 4month ago. And the code that leads to an error is still there.

https://github.com/okta/okta.github.io/blob/fb55c03e359d469b0f0689c03a10dd456dbd4181/_source/_code/react/okta_react_sign-in_widget.md#L187-L189

jmelberg-okta:
Hey @mschaeffner! 

Thanks for the Pull Request! The AuthJS and Sign-In Widget guides are targeted for audiences who are just getting started with Okta. We made a few assumptions along the way including:
- MFA is not enabled per org/ application
- The user is **not** in an actionable [lifecycle operation](https://developer.okta.com/docs/api/resources/users#lifecycle-operations).

The changes you are proposing are valuable to developers following this guide who might have users in a different **transaction state**.

With that being said, I'm proposing we update this to assume the best case `res.status === 'SUCCESS'` - while accounting for different user states.

```javascript
onSuccess(res) {
  // The user has successfully completed the authentication flow
  // and got a session token in the response.
  if (res.status === 'SUCCESS') {
    return this.props.auth.redirect({
      sessionToken: res.session.token
    });
  } else {
    // The user can be in another authentication state that requires further action.
    // For more information about these states, see:
    // https://github.com/okta/okta-signin-widget#rendereloptions-success-error
  }
}
```

jmelberg-okta:
Resolved with #2067

PR #1943, opened 2018-04-10
Fix outline numbering for Client Credentials Flow

kjnckelly:
Fix outline numbering

## Description:
- Fix the outline numbering the Client Credentials Flow documentation page


jmelberg-okta:
Hey @kjnckelly - Good find, but it looks like we made this change in the incorrect file.

Want to update [this page](https://github.com/okta/okta.github.io/blob/source/_source/_authentication-guide/implementing-authentication/client-creds.md) instead?

jmelberg-okta:
Closing due to changes being picked up with #1944

PR #1951, opened 2018-04-11
remove repeated line from oauth-client doc

mystiberry-okta:
## Description:
- There was an extra client secret in one example. Now there's not.

### Resolves:
 [OKTA-166384](https://oktainc.atlassian.net/browse/OKTA-166384)

image


mystiberry-okta:
Note that I changed the client secret value to match other examples, because the client ID is the same.

PR #1973, opened 2018-04-19
Fix margins/padding on Quickstart Guides

mraible:
The quickstarts don't look good on smaller browsers because of margin/padding settings.

screen shot 2018-04-19 at 12 30 00 pm

This PR fixes it by removing CSS that causes this:

```css
// Disable the margin width determined by the parent CSS.
.PageContent-main#quickstart-body {
  max-width: none;
  padding: 64px 0;
}
```

The result is a prettier page!

screen shot 2018-04-19 at 12 30 57 pm


jmelberg-okta:
Closing due to inactivity.

This type of fix will be resolved in the future, but happy to merge in a temporary fix that works for all screen resolutions.

PR #1998, opened 2018-04-27
Document ID tokens from refresh requests

royalchan-okta:
## Description:
- Document ID tokens from refresh requests
- The response table for /token now matches other APIs and just lists the possible values. rt and id_token inclusion is now completely dependent on scope.

### Resolves:

* [OKTA-136495](https://oktainc.atlassian.net/browse/OKTA-136495)

@mystiberry-okta @danielzeller-okta

### DIFF
--- a/_source/_docs/api/resources/oidc.md
+++ b/_source/_docs/api/resources/oidc.md
@@ -138,13 +138,13 @@ Use the postMessage() data model to help you when working with the `okta_post_me
 
 `message`:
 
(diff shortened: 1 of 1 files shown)


mystiberry-okta:
Thanks for catching that, @royalchan-okta . Will add back.

mystiberry-okta:
Holding for merge, Mysti will merge when we publish weekly.

PR #2002, opened 2018-04-30
Adds golang jwt verifier to docs

bretterer:
## Description:
- Adds links for the okta-jwt-verifier-golang library

Do not publish until [okta/okta-jwt-verifier-golang](https://github.com/okta/okta-jwt-verifier-golang) and [okta/sample-golang](https://github.com/okta/samples-golang) is made public


jmelberg-okta:
@jakubvul & @bretterer Do we want to keep these in alphabetic order? If so, lets move up `Golang` to the top of the list.

PR #2018, opened 2018-05-02
fix credentials object

mystiberry-okta:
## Description:
- Credentials was incorrectly listed as part of credentials object. Fixed it. Tested that "credentials" isn't returned in the results it was removed from using Postman collection.

### Resolves:
 [OKTA-164555](https://oktainc.atlassian.net/browse/OKTA-164555)


mystiberry-okta:
@bdemers Can you review these changes (with the PUT/POST changes removed and also the hyphen. Life's too short to be super fussy about rules of English.)? If you think we should wait for Nate, just let me know.

PR #2022, opened 2018-05-04
Update SCIM doc with Group PATCH JSON payloads

asong-okta:
## Description:
- Add SCIM doc with group PATCH JSON payloads

[Preview](https://github.com/okta/okta.github.io/blob/okta-168844-add-scim-group-push-patch/_source/_standards/SCIM/index.md#update-group-details-patch-groupsid)

### Resolves:
* [OKTA-168844](https://oktainc.atlassian.net/browse/OKTA-168844)

### DIFF
--- a/_source/_standards/SCIM/index.md
+++ b/_source/_standards/SCIM/index.md
@@ -431,8 +431,7 @@ with support for filtering and pagination:
 For more details on the `/Users` SCIM endpoint, see [section 3.4.2](https://tools.ietf.org/html/rfc7644#section-3.4.2)
 of the [SCIM 2.0 Protocol Specification](https://tools.ietf.org/html/rfc7644).
 
-###### Example
-```http request
+```
 GET /v2/Users?startIndex=1&count=2 HTTP/1.1
 Accept: application/scim+json
 Accept-Charset: utf-8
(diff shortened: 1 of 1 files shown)


_source/_standards/SCIM/index.md
```diff
@@ -955,12 +946,95 @@ Accept-Encoding: gzip,deflate
 }
 ```
 
-##### Delete Group: DELETE /groups/{id}
+##### {% api_lifecycle beta %} Update Group Details: PATCH /groups/{id}
+
+This is an alternative way to update group details, mainly to update the group membership. It is strongly encouraged to implement PATCH.
```
vladdha-okta: I think we can be a bit more prescriptive here. For example - We plan to drop support for PUT therefore all new ISVs wanting to implement Okta group push feature **must** use PATCH.

vladdha-okta: My above comment assumes that at the time of publishing the information about "PATCH", it will be a supported verb in the platform? If not, then should we also say that the Patch would be available for testing for ISVs say in early Jun? 

@bng-okta

asong-okta: @vladdha-okta , I've updated the wording to say:

> **Warning:** Updating groups via PUT request will be deprecated on May 31st, 2018. ISV's currently implementing SCIM need to implement [patch support](#update-group-details-patch-groupsid) prior to submitting their integration for review.

Issue #2038, opened 2018-05-09
Group EL Whitelist Typo


petermiller-okta:
Should read `active_directory` not `active.directory`

PR #2039, opened 2018-05-09
EL group expression whitelist typo

petermiller-okta:
## Description:
- The app name should read `active_directory` not `active.directory`

### Resolves:
* [Issue #2038](#2038)


jakubvul:
Good to go. Will merge after 2018.19 release, just to keep everything clean.

PR #2061, opened 2018-05-17
Blog post with Spring Boot 2.0 screencasts

mraible:


_source/_posts/2018-05-17-microservices-spring-boot-2-oauth.md
```diff
@@ -0,0 +1,154 @@
...
+        <url>http://repo.spring.io/milestone</url>
+    </repository>
+</repositories>
+```
+
+You'll also need to copy the auto-configuration classes into your project, along with the `spring.factories` file that configures them. GitHub's SVN support makes this possible with the following commands.
+
+```bash
+svn export https://github.com/spring-projects/spring-security/trunk/samples/boot/oauth2login-webflux/src/main/java/org src/main/java/org
+svn export https://github.com/spring-projects/spring-security/trunk/samples/boot/oauth2login-webflux/src/main/resources/META-INF src/main/resources/META-INF
```
bdemers: This is really cool! I'm not sure the average dev has subversion installed anymore.  But it might be interesting to see if we get comments on that or not.
(even if you don't have/know svn it should still be relatively obvious what these lines do)

PR #2066, opened 2018-05-21
Update to latest Spring integration version

bdemers:


_config.yml
```diff
@@ -109,7 +109,7 @@ sitemap_ignore:
 # Okta Library versions
 versions:
   jwt_validator_java: 0.2.0
-  spring_boot_starter: 0.2.0
+  spring_boot_starter: 0.5.0
```
jmelberg-okta: I'm assuming the API hasn't changed from `0.3.0` and `0.5.0` for how it is used in [spring-implicit](https://github.com/okta/okta.github.io/blob/41940648dbd67431dfbcb1691b32ebab47576f9c/_source/quickstart-fragments/java/spring-implicit.md) and [spring-auth-code](https://github.com/okta/okta.github.io/blob/3ccc2a6c105f087b431ab0ace88883562233dbde/_source/quickstart-fragments/java/spring-auth-code.md), correct?

PR #2081, opened 2018-05-29
Fixing doc bug

jakubvul:
## Description:

Default value of `sendEmail` parameter is `false`, not `true`. 

### Resolves:

* [OKTA-153304](https://oktainc.atlassian.net/browse/OKTA-153304)


_source/_docs/api/resources/users.md
```diff
@@ -1938,10 +1938,10 @@ Users that don't have a password must complete the flow by completing [Reset Pas
 ##### Request Parameters
 {:.api .api-request .api-request-params}
 
-Parameter | Description                                     | Param Type | DataType | Required | Default
---------- | ----------------------------------------------- | ---------- | -------- | -------- | -------
-id        | `id` of user                                    | URL        | String   | TRUE     |
-sendEmail | Sends an activation email to the user if `true` | Query      | Boolean  | FALSE    | TRUE
+Parameter | Description                                                                | Param Type | DataType | Required |
```
nghadge-okta: All other places we have this column vs here we are removing it completely and adding it as a description. Is that okay?
@brianzuzga-okta What is the general practice here?

PR #2090, opened 2018-06-01
Changes to remove the references to ON_NETWORK and OFF_NETWORK

rpamidimarri-okta:
Related to OKTA-173191
Bacon: test

## Description:
- Remove the support for ON_NETWORK and OFF_NETWORK in policy API.

NOTE: npm start fails locally and could not build the local version of the site. Will investigate that before merging this in.

### Resolves:
* [OKTA-173191](https://oktainc.atlassian.net/browse/OKTA-173191)

@jakubvul @eklein-okta @dannavarro-okta @richardblaylock-okta please review

### DIFF
--- a/_source/_docs/api/resources/policy.md
+++ b/_source/_docs/api/resources/policy.md
@@ -830,11 +830,10 @@ Specific zone ids to include or exclude are enumerated in the respective arrays.
 
 Parameter | Description | Data Type | Required |
 | --- | --- | --- | ---
-connection | Network selection mode | `ANYWHERE`, `ZONE`, `ON_NETWORK`, or `OFF_NETWORK` | No |
+connection | Network selection mode | `ANYWHERE`, `ZONE` | No |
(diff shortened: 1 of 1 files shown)


dannavarro-okta:
Is there a concept of in zone vs out of zone. Based on this diff I don't see that explained, but the removals look good

rpamidimarri-okta:
We use "include" and "exclude" in other places in our API - in users, groups, etc. So, this concept is not unique to network condition in policies. In all other places, we just point out that "include" means the array of entities that need to be included. 
I dont know if we need to add any other information here about include/exclude for zones. 
@jakubvul thoughts?

rpamidimarri-okta:
@eklein-okta - Was discussing with @jakubvul about this question ^ looks like I need to get a sign-off from you on this PR since you work with Adaptive? Can you please confirm?

eklein-okta:
All cool with me...

PR #2093, opened 2018-06-01
Update schema.md to describe login patterns

donwoods-okta:
We are adding support for non-email format usernames by allowing the `pattern` attribute to be set on the `login` property.  The login property is also described at `users#default-profile-properties`, so I intend to make the same changes there after incorporating feedback to this PR.  (There will also be changes to `users#okta-login`.)

## Description:
- Changed description of what parts of base schema can be modified.
- Changed description of how login is validated. (I would make this a link to the next bit but I'm not sure how to represent in-page links in Kramdown.)
- Added a few paragraphs describing the new login pattern validation, including careful description of acceptable settings.
- I don't know if I was supposed to open a separate Jira from the one cited below, which is where @maureenfoulds-okta is updating the UI docs.

### Resolves:

* [OKTA-164984](https://oktainc.atlassian.net/browse/OKTA-164984)


_source/_docs/api/resources/schemas.md
```diff
@@ -1233,13 +1233,13 @@ Custom property names for the [profile object](users#profile-object) must be uni
...
 | ----------------- | ---------------------------------------------------------------------------------------------------------------------------------- | -------- | -------- | ------ | -------- | --------- | --------- | ----------------------------------------------------------------------------------------------------------------- |
-| login             | unique identifier for the user (`username`)                                                                                        | String   | FALSE    | TRUE   | FALSE    | 5         | 100       | [RFC 6531 Section 3.3](http://tools.ietf.org/html/rfc6531#section-3.3)                                            |
+| login             | unique identifier for the user (`username`)                                                                                        | String   | FALSE    | TRUE   | FALSE    | 5         | 100       | see below                                            |
```
jakubvul: So the usual pattern here is to create a section with a heading for that content, and then link to an anchor for that heading. Is there a heading that would make sense for the new content you've added below?

donwoods-okta: I could call the section "Login Pattern Validation" and the table could say "pattern" under Validation, with that word linking to the new section.  What is the markup for the internal link, and for the new section?  (I see some examples but don't know, for instance, if the number of # signs determines the level of heading or something.)

donwoods-okta: Once the new section exists, I can presumably link to that from the similar table in the Users page, instead of adding a second copy of the section there.

jakubvul:
Note to self - this is planned for the 2018.23 release.

donwoods-okta:
CHANGE OF PLAN.  Apparently the UI isn't going to make the code freeze, so we're going to hold off on this until the next major release.  (That will also give us more time to pound on the code looking for corner cases.)  Please do NOT merge these changes with 2018.23.

PR #2127, opened 2018-06-14
updated language to remove implication that JWT extends OAuth

dogeared:
## Description:
- updated language to remove implication that JWT extends OAuth per conversation with @caseysoftware


jakubvul:
Wrong branch. You should've caught that @nbarbettini :P

nbarbettini:
Yes I should have. :(

PR #2128, opened 2018-06-15
Remove bad characters from HTML and code examples

nbarbettini:
Fixes some bad/unprintable characters.


jakubvul:
@jmelberg-okta should we make that a separate PR?

jmelberg-okta:
@jakubvul - For changes like these, we want to make sure there are tests that validate we've:

1. Caught all prior occurrences of the unprintable characters
2. Catch future unprintable characters

Any PR with a change similar to this should include tests that validate it was done properly.

jakubvul:
On that note: https://oktainc.atlassian.net/browse/OKTA-158104

nbarbettini:
I agree that this should be linted. The fact that our scripts are written in bash means that they can't be run or developed cross-platform unfortunately.

I'll advocate for migrating to an xplat language like node in the future.

PR #2129, opened 2018-06-15
Add library versioning information

nbarbettini:
Adds a page describing how our libraries and SDKs are versioned.


_source/_code/library-versions/index.md
```diff
@@ -0,0 +1,112 @@
...
+
+## Version Lifecycle
+
+Semantic versioning means that within a **major** version, you can safely apply minor and patch updates without your code breaking. The status of each major version series (for example, `1.x` or `2.x`) follows this lifecycle:
+
+### Beta
+
+Beta versions are currently in development and are not ready for production use. While a library version is in beta, the code interface is not final and breaking changes could occur without warning.
+
+New libraries (or new major versions of existing libraries) typically stay in beta for a period of time while we iterate on the design and get feedback from the community. When the library is ready for production, it will be marked as stable.
```
jmelberg-okta: > marked as stable

We have different instances where we mark libs as "alpha/beta" today (see [oidc-middleware](https://github.com/okta/okta-oidc-js/tree/master/packages/oidc-middleware#warning-construction-alpha-preview-construction-warning)).

What do you suggest is the best way to communicate with devs that a library is in beta or stable? An emoji seems to work pretty well, but we could also explore [badges](https://github.com/Naereen/badges).

_source/_code/library-versions/index.md
```diff
@@ -0,0 +1,112 @@
...
+
+### Stable
+
+Stable version series are supported by Okta for use in production code. Semantic versioning means that a stable major version may get new features (minor updates) or bug fixes (patch updates), but no breaking changes.
+
+### Retiring
+
+When a new major version of a library is released, the old major version series will be marked as **retiring**. Retiring libraries are supported by Okta for at least nine months. After this period of time, the version series will be officially retired (below).
```
jmelberg-okta: Curious why nine months was chosen. Is this the same time other companies use for their deprecation approach?

_source/_code/library-versions/index.md
```diff
@@ -0,0 +1,112 @@
...
+
+For example, if version 2.0.0 of `Okta.FoobarSdk` is published in January, the 1.x series of `Okta.FoobarSdk` (any version starting 1) will be officially supported until _at least_ September. The retirement dates will be posted in the library's readme or documentation.
+
+### Retired
+
+Retired version series are no longer supported by Okta, and are no longer recommended for production use. Okta will not publish new features or bug fixes for retired versions, except for critical security fixes (at our discretion).
+
+For example, after `Okta.FoobarSdk` version 1.x has a retiring period (of at least nine months), it is no longer supported. Customers and developers are encouraged to upgrade to 2.0.0 or later.
```
jmelberg-okta: nit: `Okta.FoobarSdk` -> `okta-sdk-foobar` or `okta-foobar-sdk` to be consistent with our repo names.

_source/_code/library-versions/index.md
alexdahl-okta: +1, would be great to have a consistent label style.

I would guess emoji are easiest to use across documents and services. But I'm open minded.

Where there are parallels with our [API versions](https://developer.okta.com/docs/api/getting_started/releases-at-okta), would be nice to be consistent. There we currently use `⚠️ Beta`. Deprecated has parallels with retirement, but because the process is different perhaps we don't need share iconography for all statuses.

_source/_code/library-versions/index.md
```diff
@@ -0,0 +1,112 @@
...
+
+When in doubt, check GitHub! The library's readme will clearly state any versions that are retiring or have been retired.
+
+## Current Versions
+
+This table lists the current (and historical) major versions of our open-source libraries and tools. The most up-to-date information is always available from the GitHub readme of a particular project.
+
+| Library                                                         | Version                                                    | Status                |
+| --------------------------------------------------------------- | ---------------------------------------------------------- | --------------------- |
+| [Sign-In Widget][signin-widget]                                 | 1.13.0<br>[**2.9.0+**][signin-widget-latest]               | Retired<br>**Stable** |
```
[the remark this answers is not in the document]
nbarbettini: Easier to read in the raw source or rendered view? I tried it your way originally and found it harder to read because the shading made it less clear which row belonged to the Sign-In Widget (for example). Maybe that's just me.

_source/_code/library-versions/index.md
nbarbettini: I am planning on creating a snippet of Markdown that will be placed in each library readme for maximum consistency. Something like:

---

### Current version status

This library uses semantic versioning and follows Okta's [open-source library version policy](http://this.example.com).

✔️ The current stable major version series is: 2.9.0+

| Version | Status                    |
| ------- | ------------------------- |
| 0.1.0  | :x: Retired               |
| 1.13.0 | :warning: Retiring on 2019-xx-xx |
| 2.9.0   | :heavy_check_mark: Stable |
| 3.0.0+  | :warning: Beta            |

_source/_code/library-versions/index.md
nbarbettini: This varies a lot, from "we published a new version just now" to ~1 year (Twilio) to many years (Microsoft). I picked a number that was a little more aggressive, but still felt reasonable.

_source/_code/library-versions/index.md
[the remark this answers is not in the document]
jmelberg-okta: The markdown view does shade the line a bit heavier than the D.O.C view. Here is an example from our [React-Native Quickstart](https://developer.okta.com/quickstart/#/react-native/nodejs/express):

screen shot 2018-06-18 at 10 56 01 am

I don't have a preference, but the current approach does require injecting HTML into our tables, an extra step for the page author.

Issue #2136, opened 2018-06-19
Linting should use a cross-platform language


nbarbettini:
The `npm run post-build-lint` command uses bash to execute this shell script: https://github.com/okta/okta.github.io/blob/source/scripts/post-build-lint.sh

Relying on bash means that it's impossible for non-*nix contributors to run or develop this script. We should investigate converting it to a Node script instead.

See: #2128 (blocked)

rdegges:
Getting rid of this for now.

PR #2159, opened 2018-06-29
Moves markdown lint tests into Node script

jmelberg-okta:
## Description:
- Moves basic markdown lint checks into a separate Node script

This PR is broken into two commits:
1. [4ff09f0](https://github.com/okta/okta.github.io/pull/2159/commits/4ff09f08f68354d70ee25ad48c702ba42b544e96) - Move lint checks to a Node script
2. [d392c81](https://github.com/okta/okta.github.io/pull/2159/commits/d392c81405991541e7492120231fe5f7baed0905) - Removed bad strings/characters

### Resolves:
* Partially resolves issue #2136, moving all `*.md` lint checks into a separate command


jmelberg-okta:
@rdegges & @aaronpk - FYI - We're modifying a few `/blog` and `/books` pages here to remove bad characters. For example:

- `“` and `”`are replaced with `"`
- `‘` and `’` are replaces with `'`
- `…` is replaced with `...`

Is this ok? If needed, we can omit them from our script

aaronpk:
why are curly quotes "bad"?

jakubvul:
@aaronpk because they're non-ASCII and Github gets weird when rendering files with non-ASCII characters. Specifically, Github breaks formatting under the “change view” when smart quotes are used. When Jekyll converts the markdown to HTML, all straight quotes are converted to smart quotes.

aaronpk:
I don't mind using straight quotes in the book pages. But if GitHub is breaking on non-ASCII characters I think that's GitHub's fault and they should fix it. There's a lot of useful characters outside the ASCII set...

PR #2182, opened 2018-07-11
{yourOktaDomain} warnings and script improvements

alexdahl-okta:
## Description:

Several changes to help users avoid Okta domain copy-and-paste errors. There's a bunch going on here; discussion very welcome.

- Adds warnings about `-admin` in URLs to ~~several places~~:
  - In the quick start "values you will need to gather" tables.
  - ~~In quick start client and server sections, before the first appearance of `{yourOktaDomain}` in a code block.~~
  - ~~In the Sign-In Widget Guide, after the first appearance of `{yourOktaDomain}` in a code block.~~
  - The warning text itself is dynamic: `{yourOktaDomain}` in the notes will be replaced with the user's actual Okta domain value when possible.
- If signed in, shows the domain value in the "values you will need to gather" tables.
- Fixes an issue where switching quick start content (such as a choosing a different language) would leave `{yourOktaDomain}` unreplaced in the new content.
- Updates the domain replacement script to retry when the page returns to the foreground. This should allow replacement in a case where the user signs in on a different browser tab, then returns to an already loaded tab.

@nbarbettini, @jakubvul: ~~Is the warning content and placement good? Anything that should be better?~~

@robertjd, @jmelberg-okta: ~~I've simulated as much as I can locally, but because of iframe same-origin security I have not actually verified that the javascript works end-to-end. Are we comfortable deploying it to production? Any way we should stage it?~~ I've since gotten `my-okta` running locally and seen replacement and retries succeed! So pretty confident in JS now.

table-signed-in

### Resolves:
* [UX-1481](https://oktainc.atlassian.net/browse/UX-1481)


_source/_includes/domain-admin-warning.html
```diff
@@ -0,0 +1,9 @@
+<div class="domain-admin-warning">
+  <blockquote>
+    <p>
+      Note: <code>https://{yourOktaDomain}</code> is different from your admin URL. <strong>Don&rsquo;t</strong> include <code>-admin</code> in the value.
+      <br>
+      When copying your Okta domain from the developer console, you can find the correct value in upper right of the dashboard.
```
nbarbettini: Is upper right **corner** accurate?

_source/_includes/domain-admin-warning.html
```diff
@@ -0,0 +1,9 @@
+<div class="domain-admin-warning">
+  <blockquote>
+    <p>
+      Note: <code>https://{yourOktaDomain}</code> is different from your admin URL. <strong>Don&rsquo;t</strong> include <code>-admin</code> in the value.
```
nbarbettini: Since this isn't a span, I'm assuming this won't be replaced. Does that UX make sense? (The examples on the page may be replaced, but this won't be)
That might be fine, just wanted to bring it up as a discussion point w/ the other reviewers.

_source/quickstart-fragments/android/default-example.md
```diff
@@ -32,7 +32,7 @@ After you have created the application there are two more values you will need t
 | Setting       | Where to Find                                                                  |
 | ------------- | ------------------------------------------------------------------------------ |
 | Client ID     | In the applications list, or on the "General" tab of a specific application.    |
-| Org URL       | On the home screen of the developer dashboard, in the upper right.             |
+| Org URL       | On the home screen of the developer dashboard, in the upper right.  Note that your Okta domain does **not** include `-admin`. <br>Our docs replace a placeholder with your actual value when possible: `https://{yourOktaDomain}`.             |
```
nbarbettini: What do you think about dropping this note and instead adding a span (that will show the real value) **if** the login script succeeds?

nbarbettini: My thinking is: the domain-admin-warning include already covers the "don't include -admin" part. This table would be most useful at-a-glance if it includes the real value when it's available.

_source/_includes/domain-admin-warning.html
jmelberg-okta: Yeah - referring to `https://{yourOktaDomain}` is strange if the user **never** sees this due to our replacement macro.

Second, does this need to be solved via documentation or SDK implementation? AuthJS and the Widget both account for `*-admin` issuers, and it probably makes more sense to throw errors there.

_source/_includes/domain-admin-warning.html
alexdahl-okta: Probably better, will do!

_source/_includes/domain-admin-warning.html
alexdahl-okta: So, all my additions of `{yourOktaDomain}` are getting wrapped in `<span class="okta-preview-domain">` on page render, and will subsequently be replaced.

I didn't expect that and I don't know what code we have that does it, but it's happening. Should we be worried?

Agree that we should be consistent with replacement, and we shouldn't talk about `{yourOktaDomain}` if we've replaced that string and the user doesn't know what we're referring to.

Also agree that SDKs should throw errors for this, but in the absence of 100% coverage there I think docs should educate. (We're seeing a decent number of attempts to reach endpoints through `-admin`.)

_source/quickstart-fragments/android/default-example.md
alexdahl-okta: Could do! Any thoughts on how to label it?

"For example"? I'm struggling with positioning it because it could either be a placeholder or your real value.

_source/_includes/domain-admin-warning.html
alexdahl-okta: As an illustration, here my Okta domain is `localhost:8080` (we're pretending) and both the warning and code block have been replaced by `myOkta.js`:

quick-start

Are we suspicious that the `span` wrapping is happening "for free"?

alexdahl-okta:
@jmelberg-okta Thanks for your help! Finally got iframe messaging working, and after some previous changes can actually see the script work!

sucess-with-retry-notes

_source/quickstart-fragments/android/default-example.md
nbarbettini: What about:

Not logged in
> On the home screen of the developer dashboard, in the upper right.

Logged in
> `https://example.okta.com`
> Located in the upper right of the developer dashboard

_source/_assets/js/myOkta.js
```diff
@@ -1,15 +1,48 @@
 (function($) {
   $(function() {
     var iframe = $('<iframe id="myOktaIFrame" src="https://login.okta.com" style="display:none"></iframe>');
-
     $('body').append(iframe);
 
-    $('#myOktaIFrame').load(function() {
-      iframe.get(0).contentWindow.postMessage({messageType: 'get_accounts_json'}, 'https://login.okta.com');
+    $('#myOktaIFrame').on( 'load', function() {
```
jmelberg-okta: nit: Extra space before `'load'`

_source/_assets/js/myOkta.js
```diff
@@ -1,15 +1,48 @@
...
+      $('.is-signed-in').show();
+    }, false);
   });
 
+  /**
+   * When the page visibility changes, check if we should retry replacement.
+   */
+  function checkSuccess() {
+    if (typeof document.hidden !== "undefined") {
+      if (!document.hidden) {
```
jmelberg-okta: I'm a little confused as to what is going on here, so please correct me if I'm wrong.

As this reads, we're wanting to check for an updated `https://{yourOktaDomain}` with every page state change (page is in another tab or minimized).

Are there any other use cases I'm not thinking of?

If you're simply looking to optimize for the quickstart pages, we could just call the iframe init function when a user selects a new framework instead.

_source/_assets/js/myOkta.js
```diff
@@ -1,15 +1,48 @@
...
+        if ( $('.okta-preview-domain').first().text() == 'https://{yourOktaDomain}' ) {
```
jmelberg-okta: If someone **does not** have an existing session using `login.okta.com`, we'd continuously attempt to call `window.getMyOktaAccounts()` on page state changes. This *might* be ok.


Something to consider for the future:

We know that `https://{yourOktaDomain}` will be populated if a user has previously signed into an account. Would it make sense to modify this text, or provide an incentive to the reader to sign in? As an example, we could animate the **Login** button.

_source/_assets/js/myOkta.js
```diff
@@ -1,15 +1,48 @@
...
+        }
+      }
+    }
+  }
+
+  /**
+   * Send the iframe our request.
+   * Globally available so that other contexts can refresh when they want to.
+   */
+  window.getMyOktaAccounts = function() {
```
jmelberg-okta: Does this **need** to be attached to the global `window` object?

Instead, treat this as a function so it can be called later:

```javascript
const getMyOktaAccounts = function() { ... }

// quickstart.js
getMyOktaAccounts();
```

alexdahl-okta: It seems like it does. With that ☝️ approach, the call in `quickstart.js` errors with `Uncaught ReferenceError: getMyOktaAccounts is not defined`.

This file's whole contents are inside a function, so I think a variable declaration here is scoped to just that function. But I'm open to alternatives!

_source/_assets/js/myOkta.js
alexdahl-okta: Yep, your reading is my intent.

Fundamentally, if there's a chance that the user has signed in after the initial page load, we want to check and give them their domain if available. The visibility change seemed like the most elegant event.

When the browser calls our visibilityChange handler,
If the browser supports `document.hidden`,
If the document is not hidden (inferring it just became or is about to become visible),
If we haven't replaced the default text yet,
call `window.getMyOktaAccounts()`.

This means we are doing some logic when, say, the page becomes hidden. (Because the `visibilityChange` event fires on any type of visibility change.) But here (line 23), we'll stop if we were just backgrounded. We only care to continue our retry when we come back to the foreground.

Quick starts are my priority, but even there I want to handle the case where:
1. User loads page without being signed in
2. User selects a different framework
3. User goes to another tab to sign in
4. User comes back to quick start

_source/_assets/js/myOkta.js
alexdahl-okta: Yeah, if someone does not have a session, we're being naively hopeful and retrying an infinite number of times. But only in response to specific user events, and in a pretty lightweight way.

(If there is no `{yourOktaDomain}` placeholder on a page, `checkSuccess()` will be called the first time it comes back to the foreground, but then our search for the placeholder's default content will come back empty and we'll remove our event listener. Likewise, if the user is signed in we'll do a check once but then remove the listener. And I _think_ retries don't even involve any network activity. So there's some overhead, but not a lot.)

Good future thought! Yeah, there's probably nice things we can do on both sides of the coin – more stuff when signed out, more stuff when signed in. 😎

alexdahl-okta:
Thanks @jmelberg-okta !

Sorry this PR has gotten long and messy.. I'm thinking of squashing before merging, but want to preserve history during review.

nbarbettini:
@alexdahl-okta I rebased and cleaned up after rebasing. Hopefully I didn't mess up any actual code or content additions. 🤞

alexdahl-okta:
Thanks @nbarbettini! Always a chance I've missed something, but looks good to me!

nbarbettini:
@robertjd Can you do a review of this?

nbarbettini:
Thank you @robertjd and @alexdahl-okta 🙂

PR #2187, opened 2018-07-12
Adding new Flask post to go live immediately.

rdegges:


rdegges:
If you happen to see this PR, please approve and merge when you do! It's supposed to go live asap =)

Thank you!

mraible:
Approved and merged. Will proofread later since I’m on my phone now.

mraible:
This block of text needs fixing:

> This template works is the same way, except it also outputs some variables. For instance, the `{{ g.user.id }}` value will output that ID value into the HTML template directly. These variables will eventually be available once we hook up the OpenID Connect library.

Should be “works in” and you might need raw/endraw tags.

The only other thing I noticed is you’re using YOUR_OKTA_ORG instead of {yourOktaDomain}. Not s big deal, but the later gives you the benefit of value substitution if you’re logged in.

PR #2199, opened 2018-07-19
Add blog post on React CRUD with Spring Boot

mraible:


_source/_posts/2018-07-19-simple-crud-react-and-spring-boot.md
```diff
@@ -0,0 +1,1541 @@
...
+}
+
+export default withCookies(Home);
+```
+
+There are some things you should be aware of in this component:
+
+1. `withCookies()` wraps the `Home` component at the bottom to give it access to cookies. Then you can use `const {cookies} = props` in the constructor, and fetch a cookie with `cookies.get('XSRF-TOKEN')`. 
+2. When using `fetch()`, you need to include `{credentials: 'include'}` to transfer cookies. You will get a 403 Forbidden if you do not include this option.
+3. The CSRF cookie from Spring Security has a different name than the header you need to send back. The cookie name is `XSRF-TOKEN`, while the header name is `X-XSRF-TOKEN`.
```
bdemers: 💯 This took me a while to figure out when writing some ITs a while back

PR #2194, opened 2018-07-12
Delete ISV dev docs

jakubvul:
## Description:
- Currently these do not meet the quality standards required to be live on our site.

Deleted documentation will be preserved on this branch: https://github.com/okta/okta.github.io/tree/isv-dev-docs

PR #2210, opened 2018-07-25
Restore ISV dev docs

nbarbettini:
Restore deleted ISV documentation. Working on a plan to move these to a new structure once we are ready to deploy updated OIN documentation.


jmelberg-okta:
I thought this was a decision you and Jakub made with #2194.

Any reason we're reverting that change?

nbarbettini:
Yes, we did. I had further discussion with other teams and we need to restore this (temporarily) until we replace it with new content.

PR #2240, opened 2018-08-07
Update users.md

gabrielsroka:
per @nbarbettini 

maybe add an example using `/me` (i usu. look at examples more than text)

### DIFF
--- a/_source/_docs/api/resources/users.md
+++ b/_source/_docs/api/resources/users.md
@@ -1726,7 +1726,10 @@ curl -v -X POST \
 ### Get Assigned App Links
 {:.api .api-operation}
 
-{% api_operation get /api/v1/users/${userId}/appLinks %} {% api_cors %}
+{% api_operation get /api/v1/users/${userId}/appLinks %} 
+
+{% api_cors %}
+This endpoint supports CORS, which means you can call it either from your backend server code, or from the browser. If you call it from the browser, use `me` as the user ID to retrieve the current user's assigned app links. If there is no active Okta session in the browser, the request will fail.
 
 Fetches appLinks for all direct or indirect (via group membership) assigned applications
 


_source/_docs/api/resources/users.md
```diff
@@ -1726,7 +1726,10 @@ curl -v -X POST \
 ### Get Assigned App Links
 {:.api .api-operation}
 
-{% api_operation get /api/v1/users/${userId}/appLinks %} {% api_cors %}
+{% api_operation get /api/v1/users/${userId}/appLinks %} 
+
+{% api_cors %}
+This endpoint supports CORS, which means you can call it either from your backend server code, or from the browser. If you call it from the browser, use `me` as the user ID to retrieve the current user's assigned app links. If there is no active Okta session in the browser, the request will fail.
 
 Fetches appLinks for all direct or indirect (via group membership) assigned applications
```
nbarbettini: Let's try switching the order of these paragraphs. I think it'll flow even better to have the main description of the endpoint first.

nbarbettini:
> maybe add an example using `/me`

What would an updated request example look like? The response example would look the same right?

gabrielsroka:
> example for `/me`
like https://developer.okta.com/docs/api/resources/users#request-example-7
and this explanation
https://developer.okta.com/docs/api/resources/users#request-parameters-1

nbarbettini:
I don't think the request/response are different enough to show a separate example. Do you think the explainer text you've added makes it clear enough that you can do `/users/me/appLinks`?

gabrielsroka:
Ok

bretterer:
Thank you for this PR to help improve our documentation. As you may know, we have moved our developer documentation to a new repo.

Since this PR is targeting our developer docs, we are requesting that you migrate this PR over to the new repo https://github.com/okta/okta-developer-docs

If you have any questions about this, or need assistance with a complex PR, please let us know in #eng-devex

PR #2278, opened 2018-08-28
DWW OKTA-186278 doc Linked Object property name restrictions

daytonwilliams-okta:
## Description:
doc Linked Object property name restrictions

### Resolves:
* [OKTA-186278](#186278)

* [OKTA-186278](https://oktainc.atlassian.net/browse/OKTA-186278)

### DIFF
--- a/_source/_docs/api/resources/linked-objects.md
+++ b/_source/_docs/api/resources/linked-objects.md
@@ -43,7 +43,7 @@ Each org can create up to 200 definitions, and assign them to an unlimited numbe
 
 {% api_operation post /api/v1/meta/schemas/user/default/linkedObjects %}
 
-Adds a linked object definition to the user profile schema
+Adds a linked object definition to the user profile schema. The `name` field found in both the `primary` and `associated` objects may not start with a digit, and can only contain the following characters: `a-z`, `A-Z`, `0-9`, and `_`.
 
 ##### Request Parameters
 {:.api .api-request .api-request-params}
--- a/package-lock.json
+++ b/package-lock.json
@@ -34,8 +34,8 @@
       "integrity": "sha1-1t4Q1a9hMtW9aSQn1G/FOFOQlMc=",
       "dev": true,
       "requires": {
-        "extend": "~3.0.0",
-        "semver": "~5.0.1"
+        "extend": "3.0.1",
+        "semver": "5.0.3"
       }
     },
     "ajv": {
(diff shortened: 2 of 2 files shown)


daytonwilliams-okta:
Merging to `source` because it is fine if this doc update precedes the associated code change by a short while.

1) The change was merged last Friday, so it is already on a train

2) Because of the peculiar characteristics of this fix, the doc update is actually mostly independent of the code change.

Feature is EA, which helps.

Bug:  Hey, I did X and now this feature doesn't work
Analysis: Doing X should never have been allowed because it corrupts the database.

Code fix: Add validation to prevent people from doing X.
Doc change: Thou shalt not do X.

So there's no reason not to tell people not to do it right away.

_source/_docs/api/resources/linked-objects.md
```diff
@@ -43,7 +43,7 @@ Each org can create up to 200 definitions, and assign them to an unlimited numbe
 
 {% api_operation post /api/v1/meta/schemas/user/default/linkedObjects %}
 
-Adds a linked object definition to the user profile schema
+Adds a linked object definition to the user profile schema.  The `primary` and `associated` `name` fields may not start with a digit or contain anything other than `a-z`, `A-Z`, `0-9`, and `_` characters.
```
donwoods-okta: "digit or" ---> "digit, nor"

I also think the last part might read more smoothly if rearranged:
nor contain any characters other than `a-z`, `A-Z`, `0-9`, and `_`.

daytonwilliams-okta: I actually think it's more readable this way.  I will defer to the expertise of the tech writers who are supposed to review this.

jakubvul: Disagree about "nor", which is pretty much only used with "neither".

But I would rephrase it as such: 

> The `name` field found in both the `primary` and `associated` objects can only contain the following characters: `a-z`, `A-Z`, `0-9`, and `_`.

daytonwilliams-okta: @jakubvul That change would write out the other half of the restriction ("may not start with a digit").  How about:

> The `name` field found in both the `primary` and`associated` objects may not start with a digit, and can only contain the following characters: `a-z`, `A-Z`, `0-9`, and `_`.

jakubvul: Sorry, missed that part!

Yes I like your revision.

donwoods-okta: "Nor" is not limited to use with "neither".  Nor is "or" limited to use with "either".  (As an example.)  However, I'm fine with Dayton's rewording (recapturing the briefly lost leading-digit restriction).

PR #2287, opened 2018-08-29
Fix typo and clarify limit/after

nbarbettini:
* Fix a repeated typo (I think this was my fault)
* Clarify that `limit` has a maximum value of 200
* The default `limit` is also 200, unless you specify a query, in which case it drops to 10
* Clarify that `after` should be treated as an opaque cursor always!


_source/_docs/api/resources/users.md
```diff
@@ -922,15 +922,11 @@ The first three parameters correspond to different types of lists:
...
-| after     | Specifies the pagination cursor for the next page of users (default is 200)                                                                  | Query      | String   | FALSE    |
+| limit     | Specifies the number of results returned (maximum 200)                                                                                       | Query      | Number   | FALSE    |
+| after     | Specifies the pagination cursor for the next page of users                                                                                   | Query      | String   | FALSE    |
 
-  * If you don't specify a value for `limit` and don't specify a query, only 200 results are returned for most orgs.
-  * If you don't specify any value for `limit` and do specify a query, a maximum of 10 results are returned.
-  * The maximum value for `limit` is 200 for most orgs.
-  *  Don't write code that depends on the default or maximum value, as it may change.
+  * If you don't specify a value for `limit`, the maximum (200) is used as a default. However, if you specify a query, the default `limit` is 10.
```
brianzuzga-okta: This is a head scratcher.  I don't see anywhere where the max limit is 10 in a query.  I tried all three syntaxes: q, filter, and search.  The limit looks to be 200 in all cases.

mohitchugh-okta: @michaelhuang-okta Is there a limit in the framework like this? I doubt it, but wondering where this came from in the first place.

mohitchugh-okta: Talked to Michael as well, the framework is not enforcing any limits. So I'd trust Brian's test.

brianzuzga-okta: It isn't framework.  It is enforced in the mediation layer.  I agree with Jakub's wording.

PR #2298, opened 2018-09-06
resolves OKTA-182100

kapilduraphe-okta:
## Description:
- Add a note block for the Scim Patch api call.

### Resolves:
* [OKTA-182100](https://oktainc.atlassian.net/browse/OKTA-182100)


kapilduraphe-okta:
Not sure about the package-lock.json. Lmk if that's an unnecessary change.

jakubvul:
So you can just drop the changes to `package.json`, we don't want to commit that.

Also, if you make the updated paragraph just one line, you'll only have to use the `>` at the beginning of the paragraph. Forcing the new lines and adding `>` for each one is sub-optimal.

kapilduraphe-okta:
@jakubvul Done, thanks.

PR #2308, opened 2018-09-12
Fix doc for user collection list when using q param.

brianzuzga-okta:
## Description:
- Fix doc for user collection list when using q param.  The default limit is 10 in this case.

### Resolves:
* [OKTA-188492](https://oktainc.atlassian.net/browse/OKTA-188492)


brianzuzga-okta:
Ok.  I was a doofus.  I asserted that the default limit was 200 for 'q' requests.  It really was 10, like the doc previously said.  Thanks for following up @nbarbettini !

nbarbettini:
Thank you @brianzuzga-okta and @jakubvul  🙂

PR #2321, opened 2018-09-18
Add support for AsciiDoc

mraible:
## Description:

This is a WIP to add AsciiDoc support to our blog using [jekyll-asciidoc](https://github.com/asciidoctor/jekyll-asciidoc). To see it in action, check out this branch and run it.

```
git checkout asciidoc
npm run -- --future
```

This is something I'd like to do since AsciiDoc has support for callouts in code and is a better language than Markdown IMO. Also, I have a lengthy blog post that a friend wrote in AsciiDoc and this might be easier than rewriting it in Markdown.

Issues:

* The blog front page doesn't get processed properly.

blog-front-page

* The code block doesn't render with colored code syntax. This block uses "java".

blog-asciidoc

* Here's what the original source looks like when it's processed using Asciidoctor.

blog-asciidoctor

At the very least, we should try to fix the look of the callouts.


mraible:
Good morning @mojavelinux! Do you have any idea how to fix the issues noted in this PR?

I tried adding the following to our `_config.yml`, but it doesn't help with code highlighting.

```yaml
asciidoctor:
  attributes:
    idprefix: _
    source-highlighter: pygments
    pygments-css: style
```

Gemfile
```diff
@@ -6,4 +6,8 @@ gem 'jekyll-redirect-from'
 gem 'jekyll-include-cache'
 gem 'html-proofer', '3.7.2'
 gem 'tzinfo-data', '1.2017.2'
-gem 'wdm', '~> 0.1.0' if Gem.win_platform?
\ No newline at end of file
+gem 'wdm', '~> 0.1.0' if Gem.win_platform?
```
jmelberg-okta: Please add a comment here to note that this dependency is **only** for the blog.

Once the blog content is migrated out, this can be removed.

mojavelinux:
My preference for syntax highlighting is highlight.js. It has a lot of theme choices, supports tons of languages, runs pretty fast, and is easy to integrate. It does require adding highlight.js and the CSS to the page template.

You can see it in action here:

https://docs.couchbase.com/java-sdk/2.6/start-using-sdk.html

mraible:
@mojavelinux I tried specifying "rouge":

```yaml
asciidoctor:
  attributes:
    source-highlighter: rouge
```

And I added "asciidoctor-rouge" to `Gemfile`:

```
group :jekyll_plugins do
  gem 'jekyll-asciidoc'
end
gem 'asciidoctor-rouge'
```

However, now I get no colors. 

screen shot 2018-09-18 at 1 28 07 pm

Do I need to add a CSS file for rouge like I did for Coderay? Our blog already uses rouge for Jekyll, so I expected it to work. [This post](http://127.0.0.1:4000/blog/2018/09/12/secure-java-ee-rest-api) has some Java code in it.

I saw the instructions recommended adding the following to integrate with Font Awesome (which we already have installed too).

```css
/* Integrating with Font Awesome */
span.icon>.fa {
    cursor: default;
}
.admonitionblock td.icon {
    text-align: center;
    width: 80px;
}
.admonitionblock td.icon [class^="fa icon-"] {
    font-size: 2.5em;
    text-shadow: 1px 1px 2px rgba(0,0,0,.5);
    cursor: default;
}
.admonitionblock td.icon .icon-note:before {
    content: "\f05a";
    color: #19407c;
}
.admonitionblock td.icon .icon-tip:before {
    content: "\f0eb";
    text-shadow: 1px 1px 2px rgba(155,155,0,.8);
    color: #111;
}
.admonitionblock td.icon .icon-warning:before {
    content: "\f071";
    color: #bf6900;
}
.admonitionblock td.icon .icon-caution:before {
    content: "\f06d";
    color: #bf3400;
}
.admonitionblock td.icon .icon-important:before {
    content: "\f06a";
    color: #bf0000;
}
```

I tried adding your suggested CSS for callouts, but that doesn't seem to work. 

As far as using highlight.js, I'm just trying to integrate AsciiDoc into our blog for a couple blog posts. We plan to move to Hugo soon (just for the blog) and will consider it then.

mojavelinux:
> Do I need to add a CSS file for rouge like I did for Coderay?

Most definitely. Nothing gets colored without CSS. All the integrations / extensions do is add the CSS classes for the tokens.

mojavelinux:
> I tried adding your suggested CSS for callouts, but that doesn't seem to work.

CSS always requires a certain degree of effort to integrate. I would examine the rules and see what is being applied and what isn't, then adjust. It could be that it's being overridden.

mojavelinux:
On the Okta dev blog, the rules are looking for a pre container with the class "highlight". It's likely that asciidoctor-rouge is using a different class name, so none of the styles are getting applied.

mraible:
> Nothing gets colored without CSS. 

Where do I find the rouge CSS files? asciidoctor-rouge points to [lib/rouge/themes](https://github.com/jneen/rouge/tree/master/lib/rouge/themes), but those are all Ruby files.

mojavelinux:
Oh, I forgot to mention that you have to require asciidoctor-rouge. It's not enough just to add the gem.

I think it's like this:

```
asciidoctor:
  requires:
  - asciidoctor-rouge
```

mojavelinux:
Actually, asciidoctor-rouge does use "highlight" as the CSS class for the container, so the styles should work so long as asciidoctor-rouge is actually being loaded.

mojavelinux:
Hmm, my mistake. To require asciidoctor-rouge, you'd follow the same instructions as for Asciidoctor Diagram.

https://github.com/asciidoctor/jekyll-asciidoc#enabling-asciidoctor-diagram

Which approach you use depends on your setup.

mojavelinux:
Okay, in your case it's like this:

```
group :jekyll_plugins do
  gem 'jekyll-asciidoc'
  gem 'asciidoctor-rouge'
end
```

mojavelinux:
> > Nothing gets colored without CSS.

> Where do I find the rouge CSS files?

As you mentioned, you already have CSS for rouge on the blog. So you're in luck. I think once you activate asciidoctor-rouge, everything should just work.

mraible:
Booyah! It works!!

screen shot 2018-09-18 at 1 59 55 pm

Now to try and fix those pesky callouts. ;)

mraible:
I was able to fix the callouts by adding `icons: font`. Now it looks _bee-yoo-tiful!_ 🎉

Thanks for all your help @mojavelinux. I owe you a pint or two! 🍻

fancy-callouts

mojavelinux:
Hooray!!

Glad to be of assistance. You are rocking the Jekyll AsciiDoc now!

Gemfile
mraible: Comment added.

mraible:
@mojavelinux I found one more issue. In our Markdown templates, we use an `img` macro to generate a link to our processed images. We also add a CSS class to the image using the second `{...}` in the code below.

```
{% img blog/electron-react-appauth-js/developer-tools.png alt:"Electron's Developer Tools" width:"800" %}{: .center-image }
```

If I put this in an `.adoc` file (after integrating jekyll-asciidoc), it just renders the escaped HTML, and the CSS class is not added. 

```
<img src="/assets/blog/electron-react-appauth-js/developer-tools-
8419cf2ad4dd16e3c56867827289696f287ce6abc419df4594e41f56aa3a119a.png" 
alt="Electron’s Developer Tools" width="800">{: .center-image }
```

I tried the following too:

```
image::../_assets/electron-react-appauth-js/developer-tools.png[alt=Electron's Developer Tools,width=1000]
```

This is unable to find the image, but I'm sure if I modified the `imagesdir` setting, I could fix it. However, it doesn't reference the processed image with the hash. Is it possible to make jekyll-asciidoc unescape the HTML?

mraible:
Using a passthrough block works, and I found I can add the class to center as an attribute.

```
++++
{% img blog/electron-react-appauth-js/developer-tools.png alt:"Electron's Developer Tools" width:"800" class:"center-image" %}
++++
```

I thought I could put `[.center-image]` before the block, but that doesn't seem to work.

The other way I figured out is to use `asset_path` to calculate the absolute URL of the image.

```
image::{% asset_path 'blog/electron-react-appauth-js/developer-tools.png' %}[alt=Electron's Developer Tools,width=800,align=center]
```

Thanks @mojavelinux!

mojavelinux:
Using the liquid tag in the target of the image block seems to me like the most sound approach to maintain good integration w/ the rest of the site.

Here's a slightly cleaner way to write it:

```
image::{% asset_path 'blog/electron-react-appauth-js/developer-tools.png' %}[Electron's Developer Tools,800,role=center-image]
```

or

```
[.center-image]
image::{% asset_path 'blog/electron-react-appauth-js/developer-tools.png' %}[Electron's Developer Tools,800]
```

PR #2325, opened 2018-09-18
Document limit of 20 groups for user creation

donwoods-okta:
## Description:
- Document limit of 20 groups for user creation.
- Also took the opportunity to fix a minor typo elsewhere.

### Resolves:
* [Issue #184773](#184773)

* [OKTA-184773](https://oktainc.atlassian.net/browse/OKTA-184773)


donwoods-okta:
Note to @jakubvul-okta -- this is adding missing details about the existing implementation, so it is not tied to any coming release and can go live at your convenience.

PR #2328, opened 2018-09-19
OKTA-188842: Adds event type catalog data and templated page.

bobtiernay-okta:
## Description:

Adds event type catalog based on metadata produced by the platform. See https://bit.ly/2xsgk47 for details.

- Adds `event-types.json` as Jekyll [data file](https://jekyllrb.com/docs/datafiles/)
- Adds `/docs/api/resources/event-types` page generated from template and data file
- Adds live search to filter down results
- Removes TOC from right-hand column
- Expands layout width to accommodate large templated table
- Adds custom stylesheet 
- Adds custom javascript file 
- Adds support for removing ToC based on page variable

#### Page
image

#### Search
image

#### Listing
image

#### Mobile
image

### Resolves:

* [OKTA-188842](https://oktainc.atlassian.net/browse/OKTA-188842)


bobtiernay-okta:
@jakubvul Good point. What are your thoughts here? Perhaps link from https://developer.okta.com/docs/api/resources/system_log#event-types and remove the existing callout?

image


I can change the wording to something like:

> The following sections outline the major event types captured by the system log. For a complete listing see the [Event Types catalog](link).

bobtiernay-okta:
Also, should we add a side nav link?

bobtiernay-okta:
Added link from System Log per discussion above.

PR #2335, opened 2018-09-21
Describe limit on search-based user queries

donwoods-okta:
When paginating through the users matching a search-based query, at most 50000 results can be viewed.  The current behavior is a bit rough around the edges (see OKTA-189268) so I describe it as it currently works, where an error is returned if the user tries to follow the "next" link after 50000 results.  We may need to touch up this description after that issue is addressed.

I deliberately avoid mentioning the behavior when fewer than 50000 results have been seen but the "next" link tries to go past 50000.  That cannot arise with the default "limit" of 200 per page (or any other size that evenly divides 50000), and the fix for OKTA-189268 should make it so that pagination always hits 50000 results exactly before yielding an error when requesting the "next" page.

## Description:

### Resolves:
* [Issue #X](#X)

### DIFF
--- a/_source/_docs/api/resources/users.md
+++ b/_source/_docs/api/resources/users.md
@@ -1263,11 +1263,11 @@ curl -v -X GET \
 
 > Listing users with search is an {% api_lifecycle ea %} feature and should not be used as a part of any critical flows, like authentication.
 
-Searches for user by the properties specified in the search parameter (case insensitive)
+Searches for users based on the properties specified in the search parameter (case insensitive)
 
 This operation:
 
-* Supports pagination.
+* Supports pagination (to a maximum of 50000 results; see note below).
 * Requires [URL encoding](http://en.wikipedia.org/wiki/Percent-encoding).
 For example, `search=profile.department eq "Engineering"` is encoded as `search=profile.department%20eq%20%22Engineering%22`.
 Examples use cURL-style escaping instead of URL encoding to make them easier to read.
@@ -1286,6 +1286,8 @@ Use an ID lookup for records that you update to ensure your results contain the 
     | `profile.department eq "Engineering"`         | Users that have a `department` of `Engineering` |
     | `profile.occupation eq "Leader"`              | Users that have an `occupation` of `Leader`     |
     | `profile.lastName sw "Sm" `                   | Users whose `lastName` starts with "Sm"         |
+
+> When paginating through a search-based query (see [Pagination](/docs/api/getting_started/design_principles#pagination)), the pages are limited to a total of 50000 results.  If the `limit` parameters of the sequence of pages sum to 50000, attempting to follow the `next` link from the last page will yield an “Invalid search criteria” error.
 
 ##### Search Examples
 


_source/_docs/api/resources/users.md
```diff
@@ -1287,6 +1287,8 @@ Use an ID lookup for records that you update to ensure your results contain the
     | `profile.occupation eq "Leader"`              | Users that have an `occupation` of `Leader`     |
     | `profile.lastName sw "Sm" `                   | Users whose `lastName` starts with "Sm"         |
 
+> When paginating through a search-based query (see [Pagination](/docs/api/getting_started/design_principles#pagination)), the pages are limited to a total of 50000 results.  If the `limit` parameters of the sequence of pages sum to 50000, attempting to follow the `next` link from the last page will yield an "Invalid search criteria" error.
```
nghadge-okta: This looks right. But when we say "If the `limit` parameters of the sequence of pages sum to 50000" I think we are opening the door for devs to ask the question about what happens if limit parameter is not evenly divides 50000.
Should we say something like "Once first 50000 results are fetched, attempting to follow the `next` link from the last page will yield an "Invalid search criteria" error"." 
Not sure if this is any better than what you are suggesting above.

donwoods-okta: As I mention in the commit message, I deliberately avoided answering that question because the current behavior is bugged.  I didn't want to describe the buggy behavior, and I also didn't want to describe the intended behavior given it doesn't currently work that way.

If we make the fix I suggest in OKTA-189268, then the limit parameters will always sum to exactly 50000 just before the link that generates an error, because the last successful page will have reduced limit if necessary to make it come out to 50000.

nghadge-okta: Yea. We should fix OKTA-189268 soon and update the doc.

nbarbettini:
Documenting the limit of 50,000 results makes sense.

I don't think we should document the error condition. That's a bug we should fix, not behavior we want to enshrine in our API docs.

donwoods-okta:
@nbarbettini The current proposal is that there will continue to be an error in that situation, though the precise error will change.  I could remove the part where I say what error message to expect, but I want to leave in the part that says it attempting to follow the "next" link after 50000 results will return an error.  Is that okay?

_source/_docs/api/resources/users.md
donwoods-okta: @nbarbettini That phrasing is still mostly accurate.  Both it and mine gloss over the existing bug that can arise if the limit parameter does not divide into 50000 and therefore the last full page of results has a NEXT link that tries to go _past_ 50000.  My wording avoids bringing up that case, whereas yours makes it sound as though the last page will return results up to the 50000 limit and the next link after that will yield an error.  In reality, the bug means the page that should have been truncated to stop at 50000 instead comes back empty, with no NEXT link.  But I think we can ignore this different for now, since the default limit (and most likely alternatives) will divide 50000 evenly and thus the last page will hit 50000 exactly.

The main difference in the wording seems to be that you avoided talking about summing the limit parameters over the course of the search, leaving it to the caller to figure out how to keep track of the total results seen so far.  That should be fine.

Meanwhile, though, I took one of your earlier comments to mean that I should not include the specific error string here, so I was very surprised to see it in your version as well.  What's the guidance there?

nbarbettini: 👍 

Fair point, we could say:

> ... will yield an error.

pranav-dubey:
Any specific reasons for limiting this to 50k?

@nbarbettini - Search feature is a superset to List User with Filter. Most customers have accounts in the access of 100k and might want to filter out the records using search criteria. If the search is returning more records than 50k, they might not be fetched by the implementor and would provide an incorrect impression for the customer.

donwoods-okta:
@pranav-dubey The usual reasons would be (1) performance and/or (2) how much we need to pay a  third party to allow us to do larger searches.  I forget how much each of those played into this particular decision.

PR #2354, opened 2018-09-26
Update apps.md

gabrielsroka:
`attributeStatements` seems to be nullable for Custom SAML


jakubvul:
Confirmed with @danielzeller-okta

PR #2360, opened 2018-09-27
Fixing the hashed password object api doc

arunelengovan-okta:
## Description:
Link to the current doc - https://developer.okta.com/docs/api/resources/users#hashed-password-object

**First change required in the PR**
_Existing documentation says_ 

value | String | The actual base64-encoded hashed password | TRUE | N/A | N/A
-- | -- | -- | -- | -- | --

**Second change required in the PR**
_Existing documentation says_

salt | String | Specifies the password salt used to generate the hash | TRUE | 22 (only for BCRYPTalgorithm) | 22 (only for BCRYPTalgorithm)
-- | -- | -- | -- | -- | --

Issue with the above doc is we don't clearly state which algorithm we are referring to. For SHA-256 scenario, it's going to be the base64 encoded password hash that was generated whereas for the BCRYPT scenario it's going to be radix64-encoded password hash.

### Resolves:

* [OKTA-190285](https://oktainc.atlassian.net/browse/OKTA-190285)

### Reviewers:
@tthompson-okta @jakubvul @motikagan-okta @jerrylee-okta 

### Secondary reviewers
@mpeng-okta @security-okta


jerrylee-okta:
The wording was a bit confusing to me. I'm wondering if should be slightly altered as below. I'll leave others to comment on the technical side of the documentation.

Under the VALUE row:
- For `SHA-256`: This is the actual base64-encoded hashed password.
- For `BCRYPT`: This is the actual radix64-encoded hashed password.

Under the SALT row:
- For `SHA-256`: Specifies the base64-encoded password salt used to generate the hash.
- For `BCRYPT`: Specifies the radix64-encoded password salt used to generate the hash

jakubvul:
I do prefer @jerrylee-okta 's phrasing, actually.

arunelengovan-okta:
modified as suggested by @jerrylee-okta

PR #2383, opened 2018-10-09
[OKTA-159451] AS API Doc: Updating keyRotate parameter

motikagan-okta:
## Description:
- Authorization Server API: Updating description of the keyRotate endpoint's 'use' parameter. The update makes the text clearer. It also changes the 'Required' field to 'True' for this parameter, because it turns out that supplying this parameter is mandatory.

### Resolves:

* https://oktainc.atlassian.net/browse/OKTA-159451


motikagan-okta:
Update were made after discussing with @federations-okta

PR #2415, opened 2018-10-23
OIDC doc minor fixes [OKTA-159463] and [OKTA-159461]

motikagan-okta:
## Description:
- OIDC doc minor fixes:
   - Corrections to postMessage() Data Model parameter descriptions
   - Correction to description 'authorize' endpoint 'display' parameter

### Resolves:

* [OKTA-159463](https://oktainc.atlassian.net/browse/OKTA-159463)
* [OKTA-159461](https://oktainc.atlassian.net/browse/OKTA-159461)


_source/_docs/api/resources/oidc.md
```diff
@@ -72,7 +72,7 @@ of the callback response.
...
 | code_challenge_method | Method used to derive the code challenge for [PKCE](/authentication-guide/implementing-authentication/auth-code-pkce). Valid value: `S256`                                                                                                                                                                                                                                                                                                                                               | Query      | String   | FALSE    |
-| display               | How to display the authentication and [consent](/docs/api/resources/apps#add-oauth-20-client-application) UI. Valid value: `page`                                                                                                                                                                                                                                                                                                                      | Query      | String   | FALSE    |
```
jakubvul: Is there a reason we're no longer calling out the only valid value for this property?

motikagan-okta: @jakubvul, that edit was based on a Jira item (https://oktainc.atlassian.net/browse/OKTA-159461):

> Note from the Fed team: We have not implemented the display parameter, just pass whatever value to the social idp. We should update the description to: "The `display` parameter to be passed to the Social Identity Provider when performing Social Login."

jakubvul: Ah ok.

_source/_docs/api/resources/oidc.md
```diff
@@ -145,11 +145,11 @@ Use the postMessage() data model to help you when working with the `okta_post_me
...
-| error_description | Additional description of the error (if relevant).                                                                | String   |
-| id_token          | An [ID token](#id-token). This is returned if the `response_type` includes `id_token`.                            | String   |
-| state             | If the request contained a `state` parameter, then the same unmodified value is returned back in the response.    | String   |
+| access_token      | An [access token](#access-token). This is returned if the `response_type` included `token`.                       | String   |
+| error             | The error code, if something went wrong.                                               | String   |
+| error_description | Additional error information (if any).                                                                | String   |
+| id_token          | An [ID token](#id-token). This is returned if the `response_type` included `id_token`.                            | String   |
+| state             | If the request contained a `state` parameter, then the same unmodified value is returned in the response.    | String   |
```
danielzeller-okta: It's a little odd to be saying "If the request contained a `state` parameter", when just above we are saying that `state` is required.
I'm guessing that we phrased it this way initially because there _wont_ be a `state` in okta-initiated sign on (clicking on a chiclet). I don't know if we want to dive into the weeds like that in this set of docs though- up to you.

motikagan-okta: @danielzeller-okta, thanks. Since there's better information on what `state` is in the Parameter Details table, just above this section, I can replace the mention here with a reference to that:
| state             | See information on the `state` parameter in [Parameter Details](#parameter-details).    | String   |

_source/_docs/api/resources/oidc.md
yuliu-okta: I think the original description is correct, that Okta only supports `page` display type. The OIDC spec also defines other display types like `popup`, `touch` and `wap`, we do have JIRA to support more. It is better that we do not lock the expectation that Okta just blindly pass this parameter to social idp, we definitely will check this parameter in the future.
Maybe we can just add `When the client uses authorize endpoint to trigger the social auth, Okta will pass it to idp.`

_source/_docs/api/resources/oidc.md
motikagan-okta: After merging, decided to edit the description of the state parameter again in a new pull request: https://github.com/okta/okta.github.io/pull/2448

_source/_docs/api/resources/oidc.md
motikagan-okta: @yuliu-okta, I missed your comment and already merged this pull request. Could we discuss this in the Jira item for that change?-- https://oktainc.atlassian.net/browse/OKTA-159461

PR #2418, opened 2018-10-23
removed incantations from next to last para

leebrandt:


leebrandt:
Misunderstood. Keith already created a PR https://github.com/okta/okta.github.io/pull/2417

PR #2419, opened 2018-10-23
OKTA-190608: Doc: Update expected q param behavior in Find Users API documentation

nghadge-okta:
## Description:
- Our documentation for Find Users API (https://developer.okta.com/docs/api/resources/users#find-users) needs to be updated to note the following q parameter behavior:

The suggested change is,

Does not return users in the Deprovisioned [status](link to status section). To return all users, use a [filter query] instead.

Resolves: OKTA-190608
Bacon: test

### Resolves:

* [OKTA-190608](https://oktainc.atlassian.net/browse/OKTA-190608)


_source/_docs/api/resources/users.md
```diff
@@ -1098,6 +1098,7 @@ curl -v -X GET \
 ]
 ~~~
 
+> Note: Does not return users in the `Deprovisioned` status. To return all users, use a [filter query] instead
```
nghadge-okta: I wasn't sure what `[filter query]` means here? Do I need to point to some section?
Also, we already have filter query for user lookup with email so not adding it.

nghadge-okta:
**Before change**
image

**After change**

image

donwoods-okta:
By analogy with the first sentence under "List All Users" on this same page, I suggest rewording the note as:
Note: This omits users that have a status of `DEPROVISIONED`.
(In addition to rearranging the sentence, note the all-caps status value.)

Also, there is a section right below your change titled "List Users with a Filter"; perhaps that's where your "filter query" should link?

donwoods-okta:
Oh, and I'm not sure whether you used the right links for @jakubvul-okta and @nbarbettini .  Jakub has both a -okta and non-okta name on github.  I didn't see a -okta version for Nate.

_source/_docs/api/resources/users.md
jakubvul: I assume we'd want this to link to this anchor: `#list-users-with-a-filter` which will take people here: https://developer.okta.com/docs/api/resources/users#list-users-with-a-filter

jakubvul:
@donwoods-okta It's ok, I get notifications from both :)

nghadge-okta:
@anishashetty-okta take a look at this.

andreyshatniy-okta:
The second sentence needs a period at the end. Looks great otherwise!

PR #2420, opened 2018-10-24
Merge new language pages and content

nbarbettini:
https://github.com/okta/okta.github.io/pull/2370/ has already been reviewed/approved, but had a broken test. I fixed it here. Merging this can supersede #2370.


nbarbettini:
FYI flickerbox-tom

_source/_layouts/software.html
```diff
@@ -2,75 +2,96 @@
 <html class="no-js" lang="en" dir="ltr">
```
jmelberg-okta: We have nearly identical files with our new `languages.html` and `software.html`. We should remove `software` and use `language` instead.

Currently, if you visit a `software` page, we aren't going to show you the `.NET` nav.

_source/_layouts/language.html
```diff
@@ -0,0 +1,125 @@
+<!doctype html>
+<html class="no-js" lang="en" dir="ltr">
+{% include head.html %}
+<body>
+	<div class="Page Page--docs-page">
```
jmelberg-okta: We need to keep with our previous convention by using `2` spaces instead of a `tab`.

_source/_layouts/language.html
```diff
@@ -0,0 +1,125 @@
...
+						</li>
+						{% endfor %}
+					</ul>
+				</div>
+			</aside>
+			<!-- END Fixed sidebar -->
+			<!-- START Page Content -->
+			<div class="PageContent-main language-content">
+				<nav aria-label="breadcrumb">
+		          <ol class="breadcrumb">
```
jmelberg-okta: nit: Formatting

_source/_assets/css/okta/components/_PageContent.scss
```diff
@@ -1,5 +1,11 @@
+.page {
```
jmelberg-okta: Looks like this is used for the `software` page - lets remove this from the global `PageContent` file and nest it within the file where this override needs to be applied.

_source/_assets/css/okta/components/_DocsPage.scss
```diff
@@ -78,9 +78,22 @@
     .Sidebar {
       background: get-color('ebony-clay');
 
+    	&.no-promo {
+			margin-top: 20px;
```
jmelberg-okta: We should remove the `Sidebar` border here, otherwise Chrome will pick it up:

screen shot 2018-10-24 at 3 22 30 pm

nbarbettini: Isn't this an existing defect on the site?

Are we introducing a new defect here?

jmelberg-okta: This is a visual bug from an existing site change. IMO, since we're already modifying the CSS for the new `language` template, we should include that fix here.

PR #2431, opened 2018-10-26
Update Okta-SCIM-11-SPEC-Test.json

andreiiatiniuc-okta:
## Description:

These two tests in SCIM 1.1 Spec Test Suite (https://developer.okta.com/standards/SCIM/SCIMFiles/Okta-SCIM-11-SPEC-Test.json) do not set the Content-Type Header which causes the customer's SCIM application to fail.

1. 'Required Test: Create Okta user with realistic values'
2. 'Required Test: Expect failure when recreating the user with same values'

SCIM 1.1 specs (http://www.simplecloud.info/specs/draft-scim-api-01.html#create-resource) shows examples of POST request where Content-Type header is set.
### Resolves:

* [OKTA-192001](https://oktainc.atlassian.net/browse/OKTA-192001)

### Reviewers
@ivanvoloshyn-okta 
@balathiruvalluvan-okta


jakubvul:
@nbarbettini should this file be hosted on the Integrate site now?

PR #2432, opened 2018-10-26
[VuePress] Change-log pages

bretterer:
This PR updates the Changelog to be dynamic based on the files in the `change-log` folder.  This is an expirement to see if we like it better, or different than running the `yarn changelog` command

Doing this makes it so live reload works when working with the changelog files.

**Adds**
- Lodash as a dependency to deal with sorting of changelog

**Change**
- Makes it so the changelog is more dynamic and does not require a script. 
- Moves everything from devDep's to Dep's in package.json
- Updates Sidebar to the correct html format


docs/.vuepress/theme/layouts/ChangeLog.vue
```diff
@@ -0,0 +1,51 @@
...
+</template>
+
+<script>
+import _ from "lodash"
+export default {
+  name: 'ChangeLog',
+  computed: {
+    changelogPages: function () {
+      let pages = this.$site.pages
+        .filter(pages => pages.path.startsWith("/change-log/2"))
```
bdemers: I'm guessing this is to filter out any non-date based change logs?

docs/.vuepress/theme/layouts/ChangeLog.vue
```diff
@@ -0,0 +1,51 @@
...
+
+      return _.orderBy(pages, 'path').reverse()
+    }
+  },
+  data: function () {
+    return {
+      site: {
+        contact_support_lc: '<a href="https://support.okta.com/help/open_case?_ga=2.35087137.18686933.1540307948-1734068328.1494253825" target="_blank">Support</a>'
```
bdemers: That is a great URL :trollface:

docs/.vuepress/theme/layouts/ChangeLog.vue
bretterer: yes, that and to not target the index.md file

docs/.vuepress/theme/layouts/ChangeLog.vue
bretterer: This is going to be removed

PR #2445, opened 2018-10-30
Make Edgar Feed valid RSS

mraible:


mraible:
@rdegges Any idea why these changes aren't reflected in https://developer.okta.com/feed-edgar.xml? Is there some cache we need to expire or something? It works when I run `npm run build-prod` locally.

rdegges:
Must be the cache. Maybe the build system from techops is broken?

PR #2504, opened 2018-11-26
[VuePress] Yarn Workspaces

bretterer:
## Description:
- Moves to a workspace based setup.


package.json
```diff
@@ -1,25 +1,10 @@
...
-    "dev": "vuepress dev docs",
-    "build": "vuepress build docs",
-    "setup-link": "cd docs/.vuepress/vuepress-theme-default && yarn link && cd ../../.. && yarn link ‘@okta/vuepress-theme-default"
-  },
-  "repository": {
-    "type": "git",
-    "url": "git+https://github.com/okta/okta.github.io.git"
+  "private": true,
+  "workspaces": {
+    "packages": ["packages/*"]
```
jmelberg-okta: nit: Since we don't need to nest multiple keys under `workspaces`, we can change this to:

```json
"workspaces": [
  "packages/*"
]
```

packages/docs/package.json
```diff
@@ -0,0 +1,18 @@
...
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/okta/okta.github.io.git"
+  },
+  "author": "{OKTA} <developer@okta.com>",
+  "license": "Apache-2.0",
+  "bugs": {
+    "url": "https://github.com/okta/okta.github.io/issues"
+  }
+}
```
jmelberg-okta: Since we're using `vuepress` as one of our commands, lets make sure we have `vuepress` declared as a devDependency

packages/vuepress-theme-default/components/Page.vue
```diff
@@ -15,7 +15,7 @@
 </template>
 
 <script>
-import { resolveHeaders } from '../util'
+import { resolveHeaders } from '../util/index'
```
jmelberg-okta: By default, JS imports look for an `index.js` file when it locates a directory. Does `'../util/` not work?

packages/docs/package.json
bretterer: I'm guessing I should also depend on `@okta/vuepress-theme-default`

```
"dependencies": {
    "@okta/vuepress-theme-default": "^0.0.1-alpha.1",
    "vuepress": "^1.0.0-alpha.11"
  }
```

package.json
```diff
@@ -1,25 +1,12 @@
...
-    "dev": "vuepress dev docs",
-    "build": "vuepress build docs",
-    "setup-link": "cd docs/.vuepress/vuepress-theme-default && yarn link && cd ../../.. && yarn link ‘@okta/vuepress-theme-default"
-  },
-  "repository": {
-    "type": "git",
-    "url": "git+https://github.com/okta/okta.github.io.git"
+  "private": true,
+  "workspaces": {
+    "packages": [
```
jmelberg-okta: This packages object isn't needed, we can just use `workspaces`:

```json
{
  "private": true,
  "workspaces": [
    "packages/*"
  ],
  ...
}
```

packages/docs/package.json
jmelberg-okta: Yep!

PR #2526, opened 2018-12-04
Change tweet parameter to simply t for Edgar feed

mraible:


rdegges:
@mraible what's the purpose of this one? Did someone change an API or something?

mraible:
Nope, I just changed the parameter name. The only reason for it is so URLs are unique.

PR #2538, opened 2018-12-07
Increase Node heap size during build and publish

robertjd:


scripts/publish.sh
```diff
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+# Give node a larger heap to prevent tarball PUT failure to registry
```
jmelberg-okta: Can you include the JIRA with more details in this comment? _Hopefully_ we won't have to do this again, but it'll be a good reference.

bdemers: (IMHO dont' include Okta JIRA's in this repo)

PR #2539, opened 2018-12-08
Added an example for the /token endpoint

haguilar-okta:
## Description:
I noticed that there was no example for the /token endpoint. Even though is not super hard to figure out how to form the request, I think an example is super useful and can avoid frustration.

### Resolves:

Do I really need a jira for this? :)


bellatumaneng-okta:
FYI @federations-okta

PR #2555, opened 2018-12-13
Add Spring Boot JPA + PostgreSQL blog post

mraible:


mraible:
Added README to https://github.com/oktadeveloper/okta-spring-boot-jpa-example/blob/master/README.md

PR #2558, opened 2018-12-14
idps: fix curl commands

gabrielsroka:
## Description:

some of these `curl` commands look wrong. please verify my corrections.

for all the other curl commands that are `GET`s, they don't need `-d '{}'` (that's a body only for `POST`s). there are about 16 of them on this page. i ~only~ "fixed" ~1 or 2~ most (if not all) of them :)

also, the "Microsoft Identity Provider" curl example is missing a URL and `'` etc


gabrielsroka:
for all the other curl that are GETs, they don't need `-d {}` (that's a body only for POSTs). there are about 16 of them on this page. i only "fixed" 1 or 2 :)

PR #2559, opened 2018-12-14
add `q` parameter

gabrielsroka:
## Description:
should we document the `q` parameter? it's listed here:
https://github.com/okta/okta-sdk-java/blob/master/src/swagger/api.yaml#L58-L69

(feel free to wordsmith my copy)


nbarbettini:
This needs to be reviewed by the team that owns this API.

kasturikumthekar-okta:
@gabrielsroka @jakubvul-okta `q` parameter is optional, when supplied it is used as a searchTerm and lists only those apps whose `name` / `displayName` start with the string in the term.

jakubvul:
@gabrielsroka can you update the description based on @kasturikumthekar-okta 's info? Specifically the bit about it only searching `name` and `displayName`? Also, it says `Sting` instead of `String`.

gabrielsroka:
@jakubvul done. 

i made it look like 
https://developer.okta.com/docs/api/resources/groups#list-groups
which looked better than
https://developer.okta.com/docs/api/resources/users#list-users

PR #2588, opened 2018-12-19
escape the dots for the regexen

gabrielsroka:
## Description:
i'm guessing that for these regexen, instead of:
`/example.okta/`
you wanted:
`/example\.okta/`

also, you don't need to escape `'/'`


scripts/markdown-lint.js
```diff
@@ -23,14 +23,14 @@ const regexValidator = [
   // (Optional) - Pass in a list of files that can be ignored from the check.
   // For global ignores, update the getFiles() method.
   {
-    regex: 'https?:\/\/(your-org|example|rain|your-subdomain|your-domain|{org}).okta*',
+    regex: 'https?://(your-org|example|rain|your-subdomain|your-domain|{org})\\.okta*',
```
jmelberg-okta: The trailing `\\` aren't required for this regex, so they can safely be removed.

scripts/markdown-lint.js
```diff
@@ -23,14 +23,14 @@ const regexValidator = [
...
     omitFiles: [
       '/_docs/api/postman/apps.json',
       '/_docs/api/postman/example.oktapreview.com.environment'
     ]
   },
   {
-    regex: 'https?:\/\/{yourOktaDomain}.com*',
+    regex: 'https?://{yourOktaDomain}\\.com*',
```
jmelberg-okta: Same as above

gabrielsroka:
> ESLint

not a bad idea. will it know that `str = '.com'` which will later be used in `RegExp(str)` needs to be escaped? if so, wow! or will it just catch many other errors?

nbarbettini:
@gabrielsroka Can you incorporate Jordan's feedback? Then we can merge this.

scripts/markdown-lint.js
gabrielsroka: @jmelberg-okta 
the `\\.` is in a string which becomes this regex : `/\./`. if you take them out, the regex is too loose: `/./` will match any character -- we want a literal dot.

scripts/markdown-lint.js
gabrielsroka: @jmelberg-okta same as above

bretterer:
Thank you for this PR to help improve our documentation. As you may know, we have moved our developer documentation to a new repo.

Since this PR is targeting our developer docs, we are requesting that you migrate this PR over to the new repo https://github.com/okta/okta-developer-docs

If you have any questions about this, or need assistance with a complex PR, please let us know in #eng-devex

PR #2596, opened 2018-12-20
[VuePress] Deployment Scripts Updated

bretterer:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,4 @@
-language: ruby
-rvm:
-- 2.2.4
+language: node
 addons:
   chrome: stable
   sauce_connect: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,6 @@
-language: node
+language: node_js
+node_js:
+  - "10"
 addons:
   chrome: stable
   sauce_connect: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,7 @@ addons:
   chrome: stable
   sauce_connect: true
 before_install:
-- gem update bundler
-- bundle install
-- nvm install $(cat .nvmrc)
+- yarn install
 script: "./scripts/travis.sh"
 dist: trusty
 env:
--- a/package.json
+++ b/package.json
@@ -6,6 +6,7 @@
   ],
   "scripts": {
     "dev": "yarn workspace docs dev",
-    "build": "yarn workspace docs build"
+    "build": "yarn workspace docs build",
+    "test": "echo \"Tests being migrated in different branch.\" && exit 0"
   }
 }
--- a/scripts/helpers.sh
+++ b/scripts/helpers.sh
@@ -46,53 +46,21 @@ function interject() {
     echo "----- ${1} -----"
 }
 
-function check_for_npm_dependencies() {
-    interject 'Checking NPM dependencies'
-    command -v npm > /dev/null 2>&1 || { echo "This script requires 'npm', which is not installed"; exit 1; }
+function check_for_yarn_dependencies() {
+    interject 'Checking Yarn dependencies'
+    command -v yarn > /dev/null 2>&1 || { echo "This script requires 'yarn', which is not installed"; exit 1; }
     npm install
-    interject 'Done checking NPM dependencies'
+    interject 'Done checking Yarn dependencies'
 }
-
-function check_for_jekyll_dependencies() {
-    interject 'Checking Jekyll dependencies'
-    command -v rvm > /dev/null 2>&1 || { echo "This script requires 'rvm', which is not installed"; exit 1; }
-
-    # The file `.ruby-version` has the current ruby version and is used by rbenv
-    # https://rvm.io/workflow/projects#project-file-ruby-version
-    # For example, this file might contain this line: "ruby-2.0.0-p643"
-    # This version of ruby is needed for the gems that Jekyll requires.
-    ruby_version=`cat .ruby-version`
-    if ! rvm list strings | grep $ruby_version > /dev/null; then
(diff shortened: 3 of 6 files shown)


scripts/setup.sh
```diff
@@ -1,9 +1,8 @@
 #!/bin/bash
 
-[[ -s "${HOME}/.rvm/scripts/rvm" ]] && source "${HOME}/.rvm/scripts/rvm"
 
-# Where the generated Jekyll site will be placed
-GENERATED_SITE_LOCATION="dist"
+# Where the generated VuePress site will be placed
+GENERATED_SITE_LOCATION="/packages/docs/.vuepress/dist"
```
bdemers: is this a relative path? or absolute?

bretterer: oh... you are right... i need to go up a level first.   yarn build puts it in this path from the root of the project

scripts/helpers.sh
```diff
@@ -46,53 +46,21 @@ function interject() {
     echo "----- ${1} -----"
 }
 
-function check_for_npm_dependencies() {
-    interject 'Checking NPM dependencies'
-    command -v npm > /dev/null 2>&1 || { echo "This script requires 'npm', which is not installed"; exit 1; }
+function check_for_yarn_dependencies() {
```
jmelberg-okta: We can remove this now that `yarn` is included in our containers by default.

scripts/publish.sh
```diff
@@ -84,7 +53,7 @@ if ! ci-update-package --branch ${TARGET_BRANCH}; then
   exit ${FAILED_SETUP}
 fi
 
-if ! npm publish --registry ${REGISTRY}; then
+if ! yarn publish --registry ${REGISTRY}; then
```
jmelberg-okta: We should double check that we're using `yarn >= 1.3.0`, as the `--registry` flag was not introduced until then. In other repos, we just use `npm publish --registry` to be safe, but I think we're OK here.

scripts/setup.sh
```diff
@@ -21,6 +20,6 @@ setup_service node v10.7.0
 export PATH="${PATH}:$(yarn global bin)"
 
 # Install required dependencies
-npm install -g @okta/ci-update-package
-npm install -g @okta/ci-pkginfo
+yarn install -g @okta/ci-update-package
```
jmelberg-okta: This command isn't a direct mapping with `npm`, as `install <package>` throws an error. You'll need to use `yarn global add <package>`

package.json
```diff
@@ -6,6 +6,7 @@
   ],
   "scripts": {
     "dev": "yarn workspace docs dev",
-    "build": "yarn workspace docs build"
+    "build": "yarn workspace docs build",
```
jmelberg-okta: Does the `@okta/vuepress-theme-default` package need to be built at the same time? Yarn `1.10` has a nice `yarn workspaces run build` command that will run `yarn build` in each package automatically.

.travis.yml
```diff
@@ -5,9 +5,7 @@ addons:
   chrome: stable
   sauce_connect: true
 before_install:
-- gem update bundler
-- bundle install
-- nvm install $(cat .nvmrc)
+- yarn install
```
jmelberg-okta: Something to consider for the future: Travis and Bacon use different yarn versions. For `okta-oidc-js` we needed to update the yarn version within Travis to get the entire feature set. You can see how we did that [here](https://github.com/okta/okta-oidc-js/blob/master/.travis.yml#L21-L22).

scripts/setup.sh
```diff
@@ -1,9 +1,8 @@
 #!/bin/bash
 
-[[ -s "${HOME}/.rvm/scripts/rvm" ]] && source "${HOME}/.rvm/scripts/rvm"
 
-# Where the generated Jekyll site will be placed
-GENERATED_SITE_LOCATION="dist"
+# Where the generated VuePress site will be placed
+GENERATED_SITE_LOCATION="../packages/docs/.vuepress/dist"
```
jmelberg-okta: I could be wrong, but because this script gets invoked from the root of the repo, this _might_ need to be "packages/docs/.vuepress/dist". Can you verify this?

scripts/helpers.sh
bretterer: Good to know... was not aware

scripts/publish.sh
bretterer: what do you mean by `to be safe`   afaik, we need to use --registry in order to target the internal okta registry instead of the standard yarn registry

scripts/setup.sh
bretterer: Good call... thanks!

package.json
bretterer: should only need to run the build on docs.  I believe the command that is actually run in the workspace `vuepress build` will handle everything for us

.travis.yml
bretterer: Ya, I think there may be a bigger issue with `yarn install` based on convos with @vijetmahabaleshwar-okta ... will learn more once we start the testing of publishing to our test bucket

scripts/setup.sh
bretterer: I will once we start running through bacon.  you may be right, but I am not 100% sure.

scripts/publish.sh
jmelberg-okta: Yep. Main concern was whether or not `--registry` would throw an error when using yarn.

I just double checked and Travis uses `1.3.2` in their builds, so this command will work as expected.

scripts/helpers.sh
bdemers: I'm a little cautious of this, we need to make sure we are able to pin our build to a node and yarn version. We have had problems the past couple times other projects have bumped the default version

PR #2617, opened 2018-12-28
updated content with the two ways the user can get to Tokens and fixe…

susanharper-okta:
## Description:
Updated content with the two ways the user can get to Tokens and fixed a few typos

### Resolves:

* [OKTA-195052](https://oktainc.atlassian.net/browse/OKTA-195052)

Question: Do we need to keep the screen shot? Not sure how much value it provides and currently needs to be updated to match changes to the UI. Thanks.


nbarbettini:
> Question: Do we need to keep the screen shot?

We can remove it, I agree that it doesn't add a ton of value.

PR #2631, opened 2019-01-08
Webfinger Doc

petermiller-okta:
## Description:
- Adding Webfinger API documentation
- Is this PR related to a Monolith release? This is being targeted for 2019.2.0 (JV)

### Resolves:

* [OKTA-198641](https://oktainc.atlassian.net/browse/OKTA-198641)


_source/_docs/api/resources/webfinger.md
```diff
@@ -0,0 +1,91 @@
+---
+layout: docs_page
+title: Webfinger
+---
+
+# WebFinger API
+
+The `webfinger` API allows you to query Okta regarding which IdP an individual user should be routed to, based on the configured rules in the IdP Discovery Policy.
```
motikagan-okta: It doesn't actually do any check of the username, right? It's only a check of the pattern of the domain? If I understood that right, I think it's important to say.

_source/_docs/api/resources/webfinger.md
```diff
@@ -0,0 +1,91 @@
...
+
+## Finding a User's IdP 
+
+{% api_operation get /.well-known/webfinger %}
```
motikagan-okta: Is supplying a query parameter mandatory? If so, maybe when giving the syntax, the query parameter should be included, i.e.: get /.well-known/webfinger?q=${resource}

_source/_docs/api/resources/webfinger.md
```diff
@@ -0,0 +1,91 @@
...
+
+| Parameter      | Description                                                             | Param Type | DataType | Required |
+| :------------- | :---------------------------------------------------------------------- | :--------- | :------- | :------- |
+| resource       | User's login value prefixed with `okta:acct:`                           | URL        | String   | TRUE     |
+| relArray       | Allows you to prefer certain IdPs                                       | URL        | Array    | FALSE    |
+| requestContext | Either an Okta state token or an Okta application's embed link          | URL        | String   | FALSE    |
```
motikagan-okta: I don't understand what `relArray` and `requestContext` do.

_source/_docs/api/resources/webfinger.md
```diff
@@ -0,0 +1,91 @@
...
+
+>Note: Valid values for relArray are `http://openid.net/specs/connect/1.0/issuer` and `okta:idp`. The first value being an Okta org and the second being any configurable IdP.
+
+##### Request Example
+{:.api .api-request .api-request-example}
+
+In this example, there is already a rule configured that has a user identifier condition, which says that users with the domain `deathstar.com` should be routed to a configured SAML IdP.
```
motikagan-okta: Maybe better to move this explanation to the Response Example section, because it explains why the response is the way it is.

_source/_docs/api/resources/webfinger.md
```diff
@@ -0,0 +1,91 @@
+---
```
motikagan-okta: Need to find a place for this content to live, probably not its own page. This is an interface that's not part of any our existing APIs. It allows for querying how an org's configured IdP routing policy would route a given username. It's a public unprotected interface, that can be queried by anyone without supplying any SSWS token.

_source/_docs/api/resources/webfinger.md
motikagan-okta: Is it correct to say that the purpose of this interface is to allow for querying how an org's configured IdP routing policy would route a given username? And it's a public unprotected interface, that can be queried by anyone without supplying any SSWS token?

petermiller-okta: I can clean this up a bit since it is possible that we actually check the list of users in the org if a shortname is provided.

johngronberg-okta: Also if we're looking up a user attribute we check against a list of users.

_source/_docs/api/resources/webfinger.md
johngronberg-okta: @wdawson-okta would be the best to comment on this now.

wdawson-okta: We said that we wouldn't document the requestContext portion of this API. However, the rest should be good to go.

_source/_docs/api/resources/webfinger.md
wdawson-okta: Definitely the second part. For the former, I think we should describe it in terms of a user identifier (username). I want to make it clear that this is only for one user at a time and will not expose an org's routing logic. So something like:
> The purpose of this interface is to allow a client to determine the identifier for a given username (or identifier) based on the org's Identity Provider Routing Rules (IdP Discovery Policy)

_source/_docs/api/resources/webfinger.md
wdawson-okta: Do we already have example domains that we use in other documentation? Would like to make sure we're consistent with that.

_source/_docs/api/resources/webfinger.md
wdawson-okta: I believe we agreed not to doc this part. @tthompson-okta to confirm.

_source/_docs/api/resources/webfinger.md
```diff
@@ -0,0 +1,91 @@
...
+In this example, there is already a rule configured that has an application condition which says if the user is targeting a specific app instance, route them to a configured Facebook IdP.
+
+~~~sh
+curl -v -X GET \
+-H "Accept: application/json" \
+-H "Content-Type: application/json" \
+"https://{yourOktaDomain}/.well-known/webfinger?q=okta:acct:joe.stormtrooper%deathstar.com&requestContext=%2Fhome%2Foidc_client%2F0oas72uoTu9yhcHQr0g3%2FalnivcK7lCqtQ1jOE0g3"
+~~~
+
+>Note: This request looks similar to the previous one, but it includes a `requestContext` which is an app's embed link.
```
wdawson-okta: Same as a previous comment, I thought we didn't want to doc `requestContext`

_source/_docs/api/resources/webfinger.md
```diff
@@ -0,0 +1,91 @@
...
+
+{% api_operation get /.well-known/webfinger %}
+
+| Parameter      | Description                                                             | Param Type | DataType | Required |
+| :------------- | :---------------------------------------------------------------------- | :--------- | :------- | :------- |
+| resource       | User's login value prefixed with `okta:acct:`                           | URL        | String   | TRUE     |
+| relArray       | Allows you to prefer certain IdPs                                       | URL        | Array    | FALSE    |
+| requestContext | Either an Okta state token or an Okta application's embed link          | URL        | String   | FALSE    |
+
+>Note: Valid values for relArray are `http://openid.net/specs/connect/1.0/issuer` and `okta:idp`. The first value being an Okta org and the second being any configurable IdP.
```
wdawson-okta: I'm not sure what the `relArray` attribute does. Limit to a particular set?

_source/_docs/api/resources/webfinger.md
petermiller-okta: The table says the first one is required. Are we supposed to doc possible error messages here?

_source/_docs/api/resources/webfinger.md
petermiller-okta: I'll change the table description to `Allows you to limit the result to certain IdPs `

_source/_docs/api/resources/webfinger.md
petermiller-okta: Will remove after confirmation

_source/_docs/api/resources/webfinger.md
petermiller-okta: Will remove after confirmation

_source/_docs/api/resources/webfinger.md
motikagan-okta: @nbarbettini, if it does need to be documented, should we discuss where it should go?

_source/_docs/api/resources/webfinger.md
motikagan-okta: @wdawson-okta, I'm not sure I understand. Should that be "allow a client to determine the **IdP** for a given username"?

_source/_docs/api/resources/webfinger.md
motikagan-okta: @petermiller-okta, no, I was just curious what happens. I don't think we need to doc error messages here.

_source/_docs/api/resources/webfinger.md
motikagan-okta: The IdP API doc page uses "example.com" as the domain for examples.

_source/_reference/webfinger/index.md
```diff
@@ -0,0 +1,101 @@
...
+{:.api .api-request .api-request-example}
+
+~~~sh
+curl -v -X GET \
+-H "Accept: application/json" \
+-H "Content-Type: application/json" \
+"https://{yourOktaDomain}/.well-known/webfinger?q=okta:acct:joe.stormtrooper%example.com&rel=http%3A%2F%2Fopenid.net%2Fspecs%2Fconnect%2F1.0%2Fissuer"
+~~~
+
+>Note: This request looks similar to the previous one, but it includes a `rel` parameter which limits the results to a particular set of IdPs.
```
susanharper-okta: which to that

Note: This request looks similar to the previous one, but it includes a `rel` parameter that limits the results to a particular set of IdPs.

_source/_reference/webfinger/index.md
```diff
@@ -0,0 +1,101 @@
...
+
+#### Response Example
+{:.api .api-response .api-response-example}
+
+In this example, there is already a rule configured that has a user identifier condition which says that users with the domain `example.com` should be routed to a configured SAML IdP. However, we supplied a `rel` parameter of `http://openid.net/specs/connect/1.0/issuer` which limited the result to Okta:
```
susanharper-okta: which to that

In this example, there is already a rule configured that has a user identifier condition which says that users with the domain `example.com` should be routed to a configured SAML IdP. However, we supplied a `rel` parameter of `http://openid.net/specs/connect/1.0/issuer` that limited the result to Okta:

_source/_reference/webfinger/index.md
```diff
@@ -0,0 +1,101 @@
+---
+layout: docs_page
+title: WebFinger
+---
+
+# WebFinger
+
+The purpose of the WebFinger interface is to allow a client application to determine the Identity Provider a given username (or identifier) should be routed to, based on your organization's Identity Provider Routing Rules (IdP Discovery Policy).
```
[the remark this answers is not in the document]
motikagan-okta: As discussed, will use: "The purpose of the WebFinger interface is to allow a client application to determine the Identity Provider that a given username (or identifier) should be routed to based on your organization's Identity Provider Routing Rules (IdP Discovery Policy)."

PR #2677, opened 2019-01-30
Add First Angular PWA post

mraible:


_source/_posts/2019-01-30-first-angular-pwa.md
```diff
@@ -0,0 +1,770 @@
...
+
+This is where progressive web applications come into play. PWA leverage a number of current browser technologies in order to provide a smooth user experience even in situations with little or no network connection. They use service workers which act a little like a proxy to intercept network requests and cache the responses. They allow the complete application to be installed in the client's browser. This means that the user can experience the web application.
```
leebrandt: This means that the user can experience the web application.

Kind of a weird sentence. Not sure what it means.

mraible: I changed it to:

```
This is where progressive web applications come into play. PWAs leverage a number of current browser technologies in order to provide a smooth user experience even in situations with little or no network connection. They use service workers which act a little like a proxy to intercept network requests and cache the responses. They allow the complete application to be installed in the client's browser. This means that the user can use the web application when they're offline.
```

PR #2683, opened 2019-02-04
Heather's first intro blog post

Quorralyne:


mraible:
@Quorralyne It looks like tests are still failing because of some invalid characters. You should probably rename the file to use today's date too.

Quorralyne:
> @Quorralyne It looks like tests are still failing because of some invalid characters. You should probably rename the file to use today's date too.

Agreed. Updated!

PR #2690, opened 2019-02-06
OKTA-159498 Empty Table Columns in Factors API Docs

gautamborah-okta:
## Description:
- The Factors API page has many empty columns: https://developer.okta.com/docs/api/resources/factors#factor-model

Many of these columns are read only or system generated or ENUMs, hence the max, min and default length for the attributes have been removed. If there is a length restriction it is explained in the descriptions.
 
### Resolves:

* [OKTA-159498](https://oktainc.atlassian.net/browse/OKTA-159498)


_source/_docs/api/resources/factors.md
```diff
@@ -3271,11 +3271,11 @@ Specifies the profile for a `web` factor
...
-| ------------- | ----------------------------- | --------------------------------------------------------------- | -------- | ------- | -------- | --------- | --------- | ---------- |
-| email   | email address of the user | String  | FALSE    | TRUE    | FALSE    |           | 100        |            |
-|---------------+-------------------------------+-----------------------------------------------------------------+----------+---------+----------+-----------+-----------+------------|
+|---------------+-------------------------------+-----------------------------------------------------------------+----------+---------+----------|
+| Property      | Description                   | DataType                                                        | Nullable | Unique  | Readonly |
+| ------------- | ----------------------------- | --------------------------------------------------------------- | -------- | ------- | -------- |
+| email   | email address of the user, max length 100 | String  | FALSE    | TRUE    | FALSE    |
```
antonioruberto-okta: Can we make the phrasing consistent with previous changes in this PR.

Using `max n char long` in previous changes vs `max length n` here

_source/_docs/api/resources/factors.md
```diff
@@ -3166,13 +3166,13 @@ Profiles are specific to the [factor type](#factor-type).
...
-| question      | unique key for question   | String    | FALSE    | TRUE   | TRUE     |           |           |            |
-| questionText  | display text for question | String    | FALSE    | FALSE  | TRUE     |           |           |            |
-| answer        | answer to question        | String    | TRUE     | FALSE  | FALSE    |           |           |            |
-|---------------+---------------------------+-----------+---------+---------+----------+-----------+-----------+------------|
+|---------------+---------------------------+-----------+---------+---------+----------|
+| Property      | Description               | DataType  | Nullable | Unique | Readonly |
+| ------------- | ------------------------- | --------- | -------- | -------| -------- |
+| question      | unique key for question   | String    | FALSE    | TRUE   | TRUE     |
+| questionText  | display text for question | String    | FALSE    | FALSE  | TRUE     |
+| answer        | answer to question, minimum 4 char long | String    | TRUE     | FALSE  | FALSE    |
```
antonioruberto-okta: Maybe we should use short form `min` instead of `minimum ` here?

We are using short form `max` instead of `maximum` in rest of changes.

Feel like it should be consistent one way or the other.

PR #2744, opened 2019-02-26
Create template to render tweets for Bulk Buffer

mraible:
This will allow us to upload "from the archive" tweets to [Bulk Buffer](https://www.bulkbuffer.com/).

Issues:

* It currently selects the first tweet. This should be changed each time we run this by modifying the `tweetIndex` variable.
* I couldn't figure out how to remove the first line break, or indent the tweet rendering and avoid leading spaces.

For the 2nd issue, adding `-%}` to the end of the following if statement removes the first line break. However, it also causes all tweets to be on the same line.

```
{%- if post.tweets and tags contains "angular" or tags contains "java" or tag contains "spring boot" or tags contains "dotnet" or tags contains "react" or tags contains "aspnet" %}
```

The same thing happens (no line breaks) if I change `{{ post.tweets[tweetIndex] ...` to be `{{- post.tweets ...`.

I'll remove the 1st line break and accept the outdent for now. The important thing is it works!

Below is an example of the first five tweets it renders. It currently renders 100 tweets.

```
Okta's Sign-In Widget makes it easy to add authentication to an Angular 6 app. See how in this tutorial from @mraible: https://developer.okta.com/blog/2017/03/27/angular-okta-sign-in-widget { from the archives } #angular #signinwidget #okta #typescript #angularcli
Build a simple Angular 6 app and add authentication with @oktadev! https://developer.okta.com/blog/2017/04/17/angular-authentication-with-oidc { from the archives } #angular #authentication #oidc #openidconnect #okta #typescript #angularcli
Want to see how to develop a Spring Boot API and and Angular 5 UI that talks to it? We have a tutorial for that!  https://developer.okta.com/blog/2017/04/26/bootiful-development-with-spring-boot-and-angular { from the archives } #springboot #start.spring.io #java #angular #typescript #angularcli
Did you know that PWAs are the way of the future? Native apps that require 60MB to download? Peshaw! Just use good ol' web technologies, a service worker, and caching to develop an awesome mobile experience →  https://developer.okta.com/blog/2017/05/09/progressive-web-applications-with-angular-and-spring-boot { from the archives } #springboot #java #pwa #progressivewebapp #angular #typescript #angularcli
Want to add user authentication to a Spring Boot + Angular app? This tutorial will hook you up! https://developer.okta.com/blog/2017/06/13/add-authentication-angular-pwa { from the archives } #pwa #progressivewebapp #angular #typescript #oidc #springboot #java
```


aaronpk:
I think something funny is going on with HTML escaping here. Here's an escaped character from one of the tweets:

> `Let's build something fun! &lt;3`

Might need to explicitly not escape, or maybe unescape in the template?

mraible:
@aaronpk That's probably from `{{ post.tweets[tweetIndex] | xml_escape }}`. Maybe we don't need the XML escaping?

aaronpk:
Yep, removing that fixes it!

PR #2762, opened 2019-03-05
Add blog post about using Schematics with React

mraible:


_source/_posts/2019-03-05-react-schematics.md
```diff
@@ -0,0 +1,378 @@
...
+  it('works', () => {
+    const runner = new SchematicTestRunner('schematics', collectionPath);
+    const tree = runner.runSchematic('rsi', {}, Tree.empty());
+
+    expect(tree.files).toEqual([]);
+  });
+});
+```
+
+One neat thing about Schematics is they don't perform any direct actions on your filesystem. Instead, you specify actions against a `Tree`. The `Tree` is a data structure with a set of files that already exist and a staging area (of files that will contain new/updated code).
```
leebrandt: Not sure I understand what this means. I re-read it a few times...

_source/_posts/2019-03-05-react-schematics.md
```diff
@@ -0,0 +1,378 @@
...
+
+## Add Dependencies to Your React App with Schematics
+
+The [reactstrap docs](https://reactstrap.github.io/) provide installation instructions. 
+These are the steps you will automate with the `rsi` schematic.
+
+1. `npm i bootstrap reactstrap`
+2. Import Bootstrap CSS:
+
+        import 'bootstrap/dist/css/bootstrap.min.css';
```
leebrandt: Where do these `import` statements go?

leebrandt: After reading farther in the post, this makes sense. But to avoid confusion, I might just drop the `import 'whatever'` statements and leave it at: Import Bootstrap CSS and Import ReactStrap...

Just a thought.

PR #2776, opened 2019-03-07
Removed beta tag for Okta MFA Policy

AjayThomas-okta:
## Description:
- Okta MFA enroll policy is in GA. 
- Doc incorrectly says it is BETA.
- https://developer.okta.com/docs/api/resources/policy#OktaMFAPolicy

### Resolves:

* [OKTA-213366](https://oktainc.atlassian.net/browse/OKTA-213366)


AjayThomas-okta:
False alarm. API is not GA

PR #2802, opened 2019-03-15
Update .NET quickstarts

laura-rodriguez:
## Description:
- Remove ClientId from Web API examples.
- Update MessageController code.
- Add a note in the implicit content to check CORS example in our GitHub.
- Change Domain to OktaDomain in the appsettings section.

### Resolves:

* [OKTA-212972](https://oktainc.atlassian.net/browse/OKTA-212972)

### DIFF
--- a/_source/quickstart-fragments/dotnet/aspnet4-implicit.md
+++ b/_source/quickstart-fragments/dotnet/aspnet4-implicit.md
@@ -41,7 +41,6 @@ public void Configuration(IAppBuilder app)
     app.UseOktaWebApi(new OktaWebApiOptions()
     {
         OktaDomain = ConfigurationManager.AppSettings["okta:OktaDomain"],
-        ClientId = ConfigurationManager.AppSettings["okta:ClientId"]
     });
 }
 ```
@@ -78,11 +77,14 @@ public class MessagesController : ApiController
             .SingleOrDefault(c => c.Type == System.IdentityModel.Claims.ClaimTypes.NameIdentifier)
             ?.Value;
 
-        return new dynamic[]
+        return Json(new
         {
-            new { Date = DateTime.Now, Text = "I am a Robot." },
-            new { Date = DateTime.Now, Text = "Hello, world!" },
-        };
+            messages = new dynamic[]
+            {
+                new { date = DateTime.Now, text = "I am a Robot." },
+                new { date = DateTime.Now, text = "Hello, world!" },
+            },
+        });
     }
 }
 ```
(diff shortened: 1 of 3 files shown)


_source/quickstart-fragments/dotnet/aspnet4-implicit.md
```diff
@@ -78,11 +77,14 @@ public class MessagesController : ApiController
             .SingleOrDefault(c => c.Type == System.IdentityModel.Claims.ClaimTypes.NameIdentifier)
             ?.Value;
 
-        return new dynamic[]
+        return Json(new
```
OkGoDoIt: If you're going to return Json(...), then you also need to change the Get() function's return type to IHttpActionResult in line 72, as seen here: https://github.com/okta/samples-aspnet/blob/master/resource-server/okta-aspnet-webapi-example/Controllers/MessagesController.cs#L18

_source/quickstart-fragments/dotnet/aspnetcore-implicit.md
```diff
@@ -84,11 +82,14 @@ public class MessagesController : Controller
             .SingleOrDefault(c => c.Type == ClaimTypes.NameIdentifier)
             ?.Value;
 
-        return new dynamic[]
+        return Json(new
```
OkGoDoIt: Same as above:

If you're going to return Json(...), then you also need to change the Get() function's return type to IHttpActionResult in line 72, as seen here: https://github.com/okta/samples-aspnet/blob/master/resource-server/okta-aspnet-webapi-example/Controllers/MessagesController.cs#L18

_source/quickstart-fragments/dotnet/aspnet4-implicit.md
laura-rodriguez: good call!

_source/quickstart-fragments/dotnet/aspnetcore-implicit.md
```diff
@@ -76,19 +74,22 @@ public class MessagesController : Controller
 {
     [HttpGet]
     [Route("~/api/messages")]
-    public IEnumerable<dynamic> Get()
+    public JsonResult Get()
```
OkGoDoIt: Generally you return an `ActionResult` but this works too.

laura-rodriguez: yeah, I just wanted to use the same type we used in our samples.

PR #2805, opened 2019-03-15
Import Inline Hook: Corrections

motikagan-okta:
## Description:
- Adding coverage of the com.okta.user.update command.
- Adding note that the UI cannot be used with AD and LDAP.

### Resolves:

* [OKTA-214512](https://oktainc.atlassian.net/browse/OKTA-214512)


jackychen-okta:
LGTM 📎

PR #2831, opened 2019-04-02
Add blog post on Better, Faster, Light Java with JHipster 6

mraible:
TODO: Create and add social image.


mraible:
@LindsayB610 URL will be `https://developer.okta.com/blog/2019/04/04/java-11-java-12-jhipster-oidc` and I've decided to give it a title of "Better, Faster, Lighter Java with Java 12 and JHipster 6". Are you OK with these parameters?

PR #2856, opened 2019-04-30
Pin Selenium version to 3.14

mraible:


package.json
```diff
@@ -12,10 +12,11 @@
     "build-prod": "npm run generate build -- --verbose",
     "start": "npm run generate serve -- --incremental --future",
     "pretest": "npm run generate serve -- --no-watch & wait-on http://localhost:4000 && npm run test:update-web-driver",
-    "test:update-web-driver": "./node_modules/protractor/bin/webdriver-manager update --gecko false",
-    "test": "echo 'selenium tests have been removed'",
+    "test-only": "protractor tests/selenium/conf.js",
+    "test:update-web-driver": "./node_modules/protractor/bin/webdriver-manager update --versions.chrome 74.0.3729.6 --gecko false",
```
vijetmahabaleshwar-okta: @mraible - Change this to `./node_modules/protractor/bin/webdriver-manager update --versions.standalone=3.141.59 --versions.chrome 2.45 --gecko false`
That should fix the failures.

vijetmahabaleshwar-okta: Actually try with this first - `./node_modules/protractor/bin/webdriver-manager update --versions.standalone=3.141.59 --gecko false`
We don't really need the `--versions.chrome` argument unless we want to use an older version of chrome.

PR #2884, opened 2019-05-17
Remove docs and files not used

mraible:
I'm not sure if there's more we can delete, but this seems like a good start.

Seems to drop "regeneration" time from 46 seconds to 26.

Thoughts @nbarbettini, @jakubvul, @robertjd?


jakubvul:
I don't see why not.

robertjd:
Heads up @mraible @jakubvul @rdegges - we had some pages that were referring to some of these static assets, for example:

https://github.com/okta/okta-auth-js/pull/213/files

We're fixing, but you should also scan your content for similar issues

PR #2935, opened 2019-06-12
adds a page that lists all posts and their tags

aaronpk:
This should make it easier to find our old posts, since you'll be able to search on this page for titles and tags of posts.

Screenshot 2019-06-12 08 11 07


rdegges:
Hmm...

So, this adds `/blog/all-posts` -- the only reason is to make it easy to ctrl+f something? I guess that sorta kinda makes sense, but feels like a hacky solution. Feel free to merge if you want, as I don't see anything wrong with it. And we can always remove it if we want.

aaronpk:
Yeah, there was a discussion in slack a while ago about how hard it was to find posts on our blog. Gonna hold off on this for now since they removed the forum results from the search at the moment. If they add that back and the search results are crap again, then we can merge this in.

PR #3022, opened 2019-08-23
updates to sign in with apple post

aaronpk:


fjaeger:
Is it correct that the email address is still only present in the id_token on first successful authentication?

